Issue #14, opened 2023-01-11
Consider dropping renv from package directory


paleolimbot:
Is there a reason you're using renv here? The use of renv here makes it annoying for me (personally) to open up this repo and work on it...I switch projects many times a day and have a rather personalized .Rprofile (which renv overrides), which this setup works poorly with. RStudio's renv integration disables some keyboard shortcuts that are useful for package development (like command-shift-L for `devtools::load_all()`) as well. I'm not aware of any other packages that use renv for dependency management (all the ones I know about use DESCRIPTION Imports/Suggests version constraints instead).

sorhawell:
I will try to list what makes `renv`currently central to `rpolars`. Any counter suggestions should be able fill in the gap. I'm happy to learn about the alternatives out there. `renv`is used analogously to rust `cargo` in rpolars and e.g. `venv` in py-polars or it could have been `poetry`.

### pinning/locking of package versions across branches and in time
Currently `pkgdown` and `rextendr`, although not required at run time do influence the final build. Other packages like data.table is only there as reference for testing. The ability to lock a package version, is useful both to pinpoint a bug and to guarantee consistency across developers and the build the system. Pinning is required to roll back if a dependency introduces a breaking change. Weekly/monthly for `rust-polars` and `rextendr` and `extendr_api` (actually just latest currently), I bump the versions and handle the failed unit tests, due to breaking changes in dependencies. It can be very hard for multiple developers to reproduce an error, if they configure dependencies manually. It can be hard to even reproduce the build server.
 
### caching
To day I was multitasking on two machines with 3 branches using different R package stacks. The `renv::restore()` allowed me to switch stack in seconds as described in renv.lock of the branch. The cache is global across any project on the machine btw. Also renv caching is used by github runners to reuse the compiled R packages across builds. The runners can access the cache of previous runners. This saves 30min build time for the ubuntu build.

### central in build server 
I was using very early the `r-lib/pak` which is a very impressive package manager and third party installer (+cache). Very much like conda. However `pak` only supports ubuntu. I'm not a `pakrat` fan, as it takes to much control. `renv` is supported on all OS and third party dependencies are currently handled directly in the github-action yaml for each OS. Today I had your PR and another PR with new R package stacks. It is super useful I can update the renv.lock and/or the Cargo.lock and check if rpolars still builds and pass all tests on all platforms before approving a PR.

footnotes: 
[...]

PR #19, opened 2023-01-12
try simply deactivate renv

sorhawell:


sorhawell:
Deactivate renv as follow-up on #14

sorhawell:
renv.lock is still actively used by build system to resolve package stack, but the good news is the renv action works fine without .Rprofile so not need in repo

PR #21, opened 2023-01-12
nanoarrow followup: docs + adjust test

sorhawell:
@paleolimbot 

Hi I added some documentation for your interface.

I did not know this way of placing the .DataFrame conversion methods in rpolars, albeit only making the methods callable from nanoarrow. That is a smart way to ensure nanoarrow/arrow are installed.


R/pkg-nanoarrow.R
```diff
@@ -1,23 +1,67 @@
...
+#' Conversion kindly provided by "paleolimbot / Dewey Dunnington" Author of `nanoarrow`.
+#' Currently these conversions are the fastest way to convert from polars to R.
+#'
+#' These conversion function are not exported/public are only accessible via `nanoarrow`
+#' e.g. \cr\cr
+#' `df = pl$DataFrame(iris)`\cr
+#' `nanoarrow::nanoarrow_allocate_array_stream(df)`\cr\cr
+#' or\cr\cr
+#' `library(nanoarrow)`\cr
+#' `nanoarrow_allocate_array_stream(df)`\cr
```
paleolimbot, line 22: That's a bit disingenuous...they're exported, just as S3 methods?

R/pkg-nanoarrow.R
```diff
@@ -1,23 +1,67 @@
...
-as_nanoarrow_array_stream.DataFrame <- function(x, ..., schema = NULL) {
+#' rpolars to nanoarrow and arrow
+#' @description Conversion via native apache arrow array stream (fast), THIS REQUIRES ´nanoarrow´
+#' @name nanoarrow
+#' @param x a polars DataFrame
+#' @param ... not used right now
+#' @param schema must stay at default value NULL
+#' @keywords nanoarrow_interface
+#' @return a nanoarrow array stream
+#' @usage as_nanoarrow_array_stream(x, ..., schema = NULL)
```
paleolimbot, line 10: Manually providing usage is a little weird (I've never seen it in a roxygen file). The examples provide this information, I think.

R/pkg-nanoarrow.R
```diff
@@ -1,23 +1,67 @@
...
-
   stream
 }
 
-infer_nanoarrow_schema.DataFrame <- function(x, ...) {
+#' @rdname nanoarrow
+#' @return a nanoarrow array schema
+#' @usage infer_nanoarrow_schema(x, ...)
+#' @examples
+#' nanoarrow_array_schema = infer_nanoarrow_schema(df)
```
paleolimbot, line 45: Usually I see multiple methods documented together as one block where `#' @return` is ` bulleted markdown list and `#' @examples` covers all the methods that are documented.

R/pkg-nanoarrow.R
```diff
@@ -1,23 +1,67 @@
...
+#' or\cr\cr
+#' `library(nanoarrow)`\cr
+#' `nanoarrow_allocate_array_stream(df)`\cr
+#'
+#' @aliases array_stream arrow nanoarrow record_batch_reader arrow_table
+#' read more at \link{https://github.com/apache/arrow-nanoarrow/r}
+#'
+#' @examples
+#' library(nanoarrow)
+#' library(rpolars)
```
paleolimbot, line 29: You don't need to `library(rpolars)` in examples (R does this for you).

paleolimbot (approved):
A few comments on idiomatic roxygen usage that you can ignore if you feel strongly about...thanks for following this up!

R/pkg-nanoarrow.R
sorhawell: I will remove usage

Usage rendered like below when using rextendr::document() in an package-env without nanoarrow and arrow.
However if building the package fully it .DataFrame-suffix is removed in usage and it looks great

```r
?as_nanoarrow_array_stream.DataFrame
ℹ Rendering development documentation for "as_nanoarrow_array_stream.DataFrame"
```
image

R/pkg-nanoarrow.R
sorhawell: I tried to describe explain the following phenomenon 
```r
#when nanoarrow and arrow is not installed
library(rpolars)
df = pl$DataFrame(iris)

# try to use the generic method
nanoarrow_array_stream = as_nanoarrow_array_stream(df)

Error in as_nanoarrow_array_stream(df) : 
  could not find function "as_nanoarrow_array_stream"
  
# the defiend function is not exported either
as_nanoarrow_array_stream.DataFrame(df)

Error in as_nanoarrow_array_stream.DataFrame(df) : 
  could not find function "as_nanoarrow_array_stream.DataFrame"
  
# if user insists, it won't work either
library(rpolars)
df = pl$DataFrame(iris)
rpolars:::as_nanoarrow_array_stream.DataFrame(df)

Error in loadNamespace(x) : there is no package called ‘nanoarrow’
  
```

Then the user installs nanoarrow and the generic method `as_nanoarrow_array_stream()` can access  rpolars:::as_nanoarrow_array_stream.DataFrame

```r
library(nanoarrow)
as_nanoarrow_array_stream(df)
<nanoarrow_array_stream struct<Sepal.Length: double, Sepal.Width: double, Petal.Length: double, Petal.Width: double, Species: dictionary(uint32)<large_string>>>
 $ get_schema:function ()  
 $ get_next  :function (schema = x$get_schema(), validate = TRUE)  
 $ release   :function ()  
```


How about?
"The generic method `as_nanoarrow_array_stream()` is only available from nanoarrow"


What do you think?

paleolimbot: The behaviour you've described is normal S3 method implementation - with our without the dynamic `s3_register()` stuff I did to avoid the hard dependency. Indeed, it would be rather rude for rpolars to provide that method since it would overwrite the original generic in the user's namespace. I think straight up removing that text would add the most clarity to your documentation.

R/pkg-nanoarrow.R
paleolimbot: You're maintaining this by yourself so feel free to do your thing...I'm just pointing out that I've never seen anybody use that syntax (which limits the ability of others to help you maintain this code).

sorhawell: https://roxygen2.r-lib.org/articles/reuse.html?q=rdname#multiple-functions-in-the-same-topic

https://github.com/tidyverse/dplyr/blob/434caa2db5c1f48b9f0becc9295e547c1072260b/R/data-bands.R

rdname used up to 155,000 times on github
https://github.com/search?l=R&p=1&q=rdname&type=Code

sorhawell: Thx for your honest opinion. Sry for laughing smiley face, I regret that.

PR #23, opened 2023-01-13
Add R CMD check workflow

eitsupi:
Based on discussion https://github.com/pola-rs/r-polars/issues/14#issuecomment-1379472243

PR #24, opened 2023-01-14
Refactoring GitHub Actions workflows

eitsupi:
Follow up #23.

- Separate release workflow from the pkgdown workflow.
- Use dtolnay/rust-toolchain instead of actions-rs/toolchain


eitsupi:
@sorhawell Sorry, I did not push to #23 after testing with fork.
Ideally, the release workflow would be included in the check workflow, but so far it has not passed testing and is a separate workflow.

sorhawell:
@eitsupi I will start on Monday fixing the many check errors in code base. Most errors seems easy fixable. So we immediately merge this PR and sometime next week when workflow passes we potentially phase out pkgdown?

eitsupi:
The intent of my post above was that the release would only be made if it passed the checks in the check workflow.
But on second thought, that may not make much sense, since the release is triggered by a tag push...

While it doesn't look very nice to have a lot of lines duplicated between yaml files, it probably works fine in this PR state.

sorhawell:
@eitsupi  
does check r-lib/actions/check-r-package/ support cargo compilation caching? It appears for me, that  check needs to recompile in a clean directory.

eitsupi:
I'm sorry, but I don't understand.
Looking at the source code, it seems to be executing `R CMD check` without moving the directory?
https://github.com/r-lib/actions/blob/105dd107f2b0f98156397bd17a54e351bc564a84/check-r-package/action.yaml#L39-L40
https://r-lib.github.io/rcmdcheck/reference/rcmdcheck.html

sorhawell:
Sorry, nevermind. I was looking at devtool::check() and mistakenly assumed the same procedure was used in r-lib check action as I did not find 'caching' mentioned in docs.

sorhawell:
maybe I'm just missing something, but it appears to me most `R CMD check` procedures, be it from `devtools` `rcmdcheck` or `r-lib/actions` create a temp check folder and rebuild package from there during the check section ` - checking whether package ‘rpolars’ can be installed ... `. Therefore missing out on the Cargo cache making build +15-20 min longer (which is alright).
However if this is true, it is possible to add a symlink to cache via makevars, which should speed up the package build to ~13 seconds. Here the CARGO_RELEASE is hardcoded, but it could be set via an environment variable.

`./src/Makevars`
```sh
LIBDIR = ./rust/target/release
STATLIB = $(LIBDIR)/librpolars.a
PKG_LIBS = -L$(LIBDIR) -lrpolars
CARGO_MANIFEST = /Users/sorenwelling/Documents/projs/r-polars/src/rust/Cargo.toml
CARGO_RELEASE = /Users/sorenwelling/Documents/projs/r-polars/src/rust/target/release/

all: C_clean

$(SHLIB): $(STATLIB)

$(STATLIB):
	cargo build --lib --release --manifest-path=$(CARGO_MANIFEST)
	if [ -s "./rust/target/release" ] ; \
	then \
	echo "dir was there" ; \
	else \
	echo "dir was not there" ; \
	mkdir -p ./rust/target ; \
	ln -s $(CARGO_RELEASE) ./rust/target/release ; \
	fi;
	echo "done"


C_clean:
	rm -Rf $(SHLIB) $(STATLIB) $(OBJECTS)

clean:
	rm -Rf $(SHLIB) $(STATLIB) $(OBJECTS) rust/target
```

sorhawell:
Hi
Fixing check errors and caching took a little longer time expected. 
I hope we can push this PR another week :)
best

eitsupi:
Sorry I have not seen this lately. (I was busy preparing to submit [my package with extendr](https://github.com/eitsupi/prqlr) to CRAN)
Happy to help with this again when I get some time.

sorhawell:
many thanks.

It seems ready to merge now right?

eitsupi:
I think it will probably be fine. 👍

PR #26, opened 2023-01-15
Fix check errors

sorhawell:


sorhawell:
turn off error on warnings, the warnings are raised as issue #28

Issue #28, opened 2023-01-16
R CMD check scanning cargo cache and prevents caching.


sorhawell:
### preface:
Since @eitsupi nice PR to use R CMD check, I have corrected all check errors/notes/warnigs in [this branch](https://github.com/pola-rs/r-polars/tree/fix_check_errors)  with two exceptions:

### "R CMD check scans cargo cache problem"
 - Note: Library is big +80Mb
 - and worse
```sh
 Warning: Found the following file(s) containing GNU extensions:
  src/rust/target/release/build/jemalloc-sys-cc682930e7558160/out/build/Makefile
  src/rust/target/release/build/jemalloc-sys-cc682930e7558160/out/build/Makefile.in
  src/rust/target/release/build/libz-ng-sys-28692994b65cd627/out/build/Makefile
```
Besides this warning there are other warnings(windows) and/or notes about missing line-endings in temp builds.
Overall it is probably not feasible to make 112 crate maintainers promise not to trigger R CMD check.

### "R CMD caching problem"
R CMD check will create a temp and reinstall rpolars. Since the cargo cache is not moved along it seems. Some 20 minutes will be used to recompile rpolars from scratch.

### One candidate solution:
Build package with e.g. devtools::build(binary=TRUE) in a previous step which can take advantage of the Cargo cache.
Add to ./src/Makevars to check if env-variable RPOLARS_BINARY_PATH is defined. Then symlink that file e.g. `rpolars.a` and skip cargo build and jump to linking. This solution both solve the "R CMD check scans cargo cache problem" and "R CMD caching problem".

I love to learn there exists more simple and elegant solutions :)

sorhawell:
releted solutions:

[delete cargo cache / temp folder](https://github.com/cran/hellorust/blob/7a155c5a3f6ef3dcd84710a0a9bbc373c3bddb57/src/Makevars#L18-L19)

[string2path on cran by @yutannihilation ](https://github.com/yutannihilation/string2path/blob/45dbb6072bf02d0442f1574c6cfbdcaa39e67a1b/configure#L27-L35)

sorhawell:
[maybe move this solution to rextendr](https://github.com/extendr/rextendr/issues/232)

sorhawell:
switching to Swatinem/rust-cache@v2 from generic cache helped a lot

sorhawell:
fixed with #30 and #32

PR #30, opened 2023-01-18
Fix cache and check scan

sorhawell:
aims to solve #28


sorhawell:
still missing to set new action env_var RPOLARS_RUST_SOURCE in workflow

sorhawell:
wrong zstd in path error  https://github.com/actions/cache/issues/1073

src/Makevars.win
```diff
@@ -14,6 +14,11 @@ rpolars_ext_binary = "$(RPOLARS_RUST_SOURCE)/target/$(TARGET)/release/librpolars
 all: C_clean
 
 $(SHLIB): $(STATLIB)
+# CRAN policy forbids using $HOME so we build in the current working dir.
+# This makes things really slow because we have to reinstall all dependencies
+# for each build. Comment this out for non-CRAN releases.
+CARGOTMP=$(PWD)/.cargo
```
eitsupi: `$(PWD)` may not work here.
`$(CURDIR)` should be use in Makefiles.
https://stackoverflow.com/questions/52437728/bash-what-is-the-difference-between-pwd-and-curdir

sorhawell: Thank you for seeing this,  I will revisit this next week

PR #41, opened 2023-02-09
rust_result tests + fixes

sorhawell:


tests/testthat/test-rust_result.R
```diff
@@ -0,0 +1,107 @@
...
+    are_ok_result,
+    are_err_result
+  )
+
+  not_results = list(
+    are_invalid_result
+  )
+
+  #test is_result
+  for(i in are_results) expect_true( is_result(i), info = paste("testcase was ",str_string(i)))
```
eitsupi, line 44: It may be worth considering using patrick.
https://github.com/google/patrick

sorhawell: I will try it out next time :)

sorhawell: I'm not sure this rust_result should be taken too far. It will probably push non rust R users away from reading the source code. What do you think?

eitsupi: I do not think this test is hard to read.

sorhawell: how about this ?
https://github.com/pola-rs/r-polars/blob/d8ad7d9a99b169be5581f4856f3bef1d4bcc93fd/R/expr__string.R#L57

it allows me to append context to err messages without a package. But maybe it is too strange to read for non rustaceans

eitsupi: Oh, I don't know what the `pcase` function is doing......
Perhaps something like switch or case when?

sorhawell: yes exactly it is a rip-off from data.table::fcase, p for polars

https://github.com/pola-rs/r-polars/blob/fd3dcb5f92e316611597c5fb080ce44d30dba1f1/R/utils.R#L114

PR #48, opened 2023-02-20
Extendr new error handling

sorhawell:


R/namespace.R
```diff
@@ -4,3 +4,17 @@ NULL
...
+#' @export
+#' @details
+#' py-polars use python None to encode polars Null, where r-polars uses NA.
+#' When copy pasting py-polars code, forgetting to change None to NA
+#' should not be a show-stopper.
+#'
+#' None should not be used casually.
+#' @examples
+#' None
+None = NA
```
eitsupi, line 20: Should this be exported?

```suggestion
#' None should not be used casually.
#' @examples
#' None
#' @noRd
None = NA
```

sorhawell: When copy pasting and running py-polars snippets, I often forget to also change python-None to R-NA. Both encodes polars-Null. So I thought how about aliasing None with NA in R.  
So it should be exported in theory, but I think it is probably too silly, so I will likely drop the alias again.

eitsupi: I see, I think it makes sense to remove it eventually.

sorhawell: it will disappear will #47  :)

PR #49, opened 2023-02-24
Touch libgcc_eh.a

yutannihilation:
We found touching it is enough to deceive the compiler (ref: https://github.com/extendr/rextendr/issues/209#issuecomment-1328236139).


sorhawell:
uhh this was not the arm64-linux, but fix for some compilations on windows. Still thanks :)

PR #58, opened 2023-03-02
Refactor Makefiles

eitsupi:
Remove lines that are not considered to be working and formatting.

Since a new shell is executed for each line in Make, there is no point in changing directories with non-consecutive commands in a target, I think.

```sh
$ cat <<"EOF" >/tmp/Makefile
all:
	echo $(PWD)
	cd /tmp/
	echo $(PWD)
EOF
$ make -f /tmp/Makefile
echo /
/
cd /tmp/
echo /
/
```

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -10,32 +10,25 @@ all: C_clean
 $(SHLIB): $(STATLIB)
 
 $(STATLIB):
-
-	curdir=$(CURDIR)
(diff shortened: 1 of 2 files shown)


sorhawell:
Looks very fine,  I will just test in my afternoon if compatible with some local test procedure I use, and then approve.

sorhawell (approved):

PR #66, opened 2023-03-06
extendr_api 0.4

sorhawell:


eitsupi:
This PR break the Nix workflow added by #54.
If the Nix workflow cannot be maintained, I would suggest deleting all Nix related files for now.
I suppose we could consider doing that in another repository or doing it again when this repository is updated less frequently.

@Sicheng-Pan Any thoughts?

sorhawell:
I tried to read the docs for flake and nix and run an update of hashes. But I did not succeed. I probably bump rust-polars or something else once a week, so it seems a bit high work load to guarantee nix succeeds for any main build.

Maybe we can divert the nix/flake build to a branch an update when ever on feels like it.

Sicheng-Pan:
Currently it seems that the hash is not specified correctly.
```
error: hash mismatch in fixed-output derivation '/nix/store/y4lqwn6kbr1ylnhvvswkhcslq92c9n8f-extendr-d5eba3a.drv':
         specified: sha256-HOglEF9PLLV12uP3gWr/6pSZFfn38I+ATcJPBbGzpJI=
            got:    sha256-tbA8+aaGz2p71BfPSgPBRxqwgD+XHGoSrMfGM/RKZHo=
```
Simply replace the hash for `extendr-api-0.4.0` in `flake.nix` will resolve this issue.

Moreover, this is due to the fact that GitHub links are used in `Cargo.toml` instead of version numbers, which may not lead to reproducible behavior:
https://github.com/pola-rs/r-polars/blob/2db25f7f038dc8e0c8785a7d5b80c99f74363392/src/rust/Cargo.toml#L19

I know it might be preferable to use the latest commit, but I'm wondering if it is possible to simply use their latest release (i.e. `extendr-api = { version = "0.4.0", ... }`)? In that way you do not need to manually update the hashes to guarantee reproducible behavior, and the corresponding hash in `flake.nix` can be removed. The same goes for other hashes specified in `flake.nix`.

For Cachix and the corresponding workflow, we can ignore it as Cargo is not perfectly integrated with Nix, and it is not very efficient to develop with Nix for now. The following lines in `flake.nix` can be removed together with the Cachix workflow:
https://github.com/pola-rs/r-polars/blob/2db25f7f038dc8e0c8785a7d5b80c99f74363392/flake.nix#L51-L55

sorhawell:
> I know it might be preferable to use the latest commit, but I'm wondering if it is possible to simply use their latest release (i.e. extendr-api = { version = "0.4.0", ... })?

Some day when things are more stable that would be better.

In the case of extendr-api we use an "unstable" feature [error handling](https://github.com/extendr/extendr/pull/473) which has not been adopted in the main repo yet. The `r_result_list`-generic was a temporary workaround, but unsatisfying as it did support error-bubbling out-of-the-box. It seems we (extendr devs) are starting to agree on how to move forward, but it was not ready for 0.4.0 release. Maybe earliest in 0.5.0. I need to finalize the PR for extendr which could take maybe 2-5 full workdays.

In the case of rust-polars crate, r-polars depend on a few features that are not in the public rust-polars api because py-polars does so as well. Also the breaking changes in rust-polars can be rough to resolve. To pinpoint the exact commit in rust-polars is helpful, as r-polars need som new feature "x", but cannot yet handle breaking change "y"

I made support for cross compiled binaries for M1 macbooks, but that is still quite handheld due to some versioning compatibility stuff.
We also have the R.universe build on separate branch, which is different as it only supports Rust stable toolchain + other minor differences.

Since time is limited, we should not require during development that every external build system works for every commit, but move them to separate branches. Otherwise every developer must understand R.universe and nix to write a PR.

sorhawell:
@Sicheng-Pan would it be possible to do a nix install from a branch of r-polars, as for R.universe and long_arms_64 (cross comilation)?

Sicheng-Pan:
Currently the nix flake can be deployed on all platforms (except for Windows, obviously). I've tested it on M1 and x86_64 Linux. For sure you can put the flake on another branch (e.g. `nix`) and people can still use `nix develop github:pola-rs/r-polars/{branch}` to enter the development environment.

Issue #64, opened 2023-03-05
Conversion from (nano)arrow ?


eddelbuettel:
The Arrow C interface is bidirectional. Having the the _export_ to (nano)arrow Table and RecordBatch is useful. It would also be nice to vbe able to go the other way and take a Arrow schema and array_data pointer pair to instantiate a polars instance.  

Are there any plans for this?    (If there are any I apologize. I looked and only found _export_ but not _import_.)

sorhawell:
Hi Dirk, thank you for asking :) I think that would be the  [DataFrame.from_arrow()](https://github.com/pola-rs/polars/blob/master/py-polars/polars/convert.py#L308-L390) and that I need to implement.  I will try to do so right now.

sorhawell:
uhh I hit a steep bump on my learning path on pointer magic and the Arrow protocol. But now it works. It is 20 times faster than current full-copy rpolars `pl$DataFrame()` [The implementation is very analogues to py_polars](https://github.com/pola-rs/r-polars/blob/f300f8d9cdafba7a525526bee2909c4e54c609ee/src/rust/src/arrow_interop/to_rust.rs#L75) it calls arrow:::ExportArray for every chunk of every column. If chunks are very small, it might slow down a bit.

right now import speed is around 21ms for 600k row x 19 columns.

```R

> library(rpolars)
> library(arrow)
> library(nycflights13)
> 
> 
> rbr = as_record_batch_reader(nycflights13::flights)
> big_arrow_table = do.call(rbind,lapply(1:2, \(x) arrow::arrow_table(nycflights13::flights)))
> rbr = as_record_batch_reader(big_arrow_table)
> 
> dim(big_arrow_table)
[1] 673552     19
> 
> # via arrow API
> system.time({df = rpolars:::rb_list_to_df(rbr$batches(),rbr$schema$names)})
   user  system elapsed 
  0.020   0.001   0.022 
> 
> big_df = as.data.frame(big_arrow_table)
> 
> # via r-polars conversion full copy
> system.time({df_simple = pl$DataFrame(big_df)})
   user  system elapsed 
  0.381   0.007   0.387 
> 
```

eddelbuettel:
Nice. And you immediatly went for batches.

sorhawell:
[...]

PR #67, opened 2023-03-07
From arrow

sorhawell:


sorhawell:
enables arrow import #64

sorhawell:
needs a little more testing and tidying and then it should be good to go

PR #78, opened 2023-03-13
revert to nightly

sorhawell:


ritchie46:
I would always stay on nightly. We get SIMD only on nightly.

ritchie46:
> We have a release branch for R.universe where nightly and SIMD is disabled as it is 'stable' only.

Is that a requirement for R.universe? What is the reason for that?

jeroen:
Simd is usually not allowed on CRAN because the binaries that they publish are supposed to be portable, i.e work an all common hardware, including those without sims extensions. The same holds by the way if you would ever distribute software via a linux distro (e.g. deb or rpm), it has to work on all hardware.

In terms of the toolchains available:

On r-universe we build with the following versions of rust:
 - ubuntu linux: [mozilla rust-updates](https://launchpad.net/~ubuntu-mozilla-security/+archive/ubuntu/rust-updates?field.series_filter=jammy) (now 1.65.0)
 - mac: what comes preinstalled on GitHub actions, see [macos-12-Readme.md](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md)
 - windows: what comes preinstalled on GitHub actions, see [Windows2022-Readme.md](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md)

CRAN uses:
 - Debian: https://packages.debian.org/testing/rustc (now 1.63.0)
 - Fedora: https://src.fedoraproject.org/rpms/rust (now 1.68)
 - MacOS: whatever the sysadmin has installed (now 1.65)
 - Windows: whatever the sysadmin has installed (now  1.65.0)

sorhawell:
Uh these are the new versions on CRAN are here already.  Then I could start trying to submit to CRAN now.

PR #81, opened 2023-03-20
Add vignette and other doc improvements

grantmcdermott:
I hope this isn't too presumptuous. But this PR proposes various doc improvements:

- Streamlines the README (now built as an .Rmd file). Also updates badges.
- Adds an intro vignette that is a lengthed/adapted version of the "teaser" example that used to be in the README.
- Adds a NEWS file. I basically copied across the relevant version release notes, with some added tweaks that I could pick up from the old README.

I also fixed various typos, although may have introduced some new ones! Unfortunately, I couldn't do a test of the website build from this machine (I need to fix my Arrow config.) But I think the GHA workflow should take care of that anyway.


sorhawell:
Amazing ! Thank you so much @grantmcdermott .

I was thinking the previous readme was starting to look like a sleazy tabloid newspaper with large and small fonts mixed together :)

I will take a look at this tonight

DESCRIPTION
eitsupi, line 1: Maybe `VignetteBuilder: knitr` is needed here?

grantmcdermott: Resolved by https://github.com/grantmcdermott/r-polars/commit/e48dc0e20eda12f9a37e8a3a6f0b16b4771094e8

grantmcdermott:
Should be good to go now. We can revisit some of the doc/repo references once #82 is decided.

sorhawell:
Many thanks @grantmcdermott !!

PR #83, opened 2023-03-20
Fix typos

grantmcdermott:
Sorry, I missed these during the previous PR.

(Semi-related: I don't think the updated website is building automatically. Do we need to amend the pkgdown workflow action?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ docs
 .Rprofile
 check
 inst/doc
+.DS_Store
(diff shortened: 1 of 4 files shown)


sorhawell:
Currently pkgdown is always build, but only uploaded if it is a github binary release by [this logic gate](https://github.com/pola-rs/r-polars/blob/e515067dd603d3840c3d56c99628c9972ef5b642/.github/workflows/pkgdown.yaml#L34)

We could simply upload pkgdown docs on any commit instead. Maybe that makes more sense now where there are multiple release channels.

grantmcdermott:
Ah, I see.

I don’t want to distract too much from the main point of this PR (which is just about fixing some typos). But another nice option is to use the [automatic development mode](https://pkgdown.r-lib.org/articles/how-to-update-released-site.html?q=Devel#automatic-development-mode) in the pkgdown.yaml. So users can choose which version of the website to visit, e.g.

- https://vincentarelbundock.github.io/marginaleffects/ (stable)
- https://vincentarelbundock.github.io/marginaleffects/dev/ (dev)

sorhawell:
Many thanks @grantmcdermott :)

Issue #87, opened 2023-03-23
Using null_values argument in csv_reader() / and lazy_csv_reader() causes error


msummersgill:
Consider a minimal CSV as follows, saved to the working directory as "example.csv".

```
x,y
0,-999.25
```

```r
write.csv(data.frame(x = 1, y = -999.25), "example.csv", row.names = FALSE)
```
Reading it with the command below works as expected

```r
> pl$csv_reader("example.csv")
# polars DataFrame: shape: (1, 2)
# ┌─────┬─────────┐
# │ x   ┆ y       │
# │ --- ┆ ---     │
# │ i64 ┆ f64     │
# ╞═════╪═════════╡
# │ 0   ┆ -999.25 │
# └─────┴─────────┘
```
Specifying `null_values = "-999.25"` results in the error below. Behavior is not specific to the string `"-999.25"`, an argument of `"0"` gives the same result.

```r
> pl$csv_reader("example.csv",null_values = "-999.25")$collect()
# Error in (function (path, sep, has_header, ignore_errors, skip_rows, n_rows, :
# expected RNullValues
```

sorhawell:
@msummersgill thank you for point out this bug. I am also in the process of completely re-implementing the csv readers with better argument checks, error handling and unit tests to avoid something like this.

PR #89, opened 2023-03-23
fix csv null value

sorhawell:
#87


msummersgill:
I tested and confirmed the changes in this branch will resolve #87, thanks for the quick response!

PR #97, opened 2023-03-30
Add binary subnamespace Expr methods

xendai66:


sorhawell:
@xendai66  Awesome many thanks! I will take a look at it now.

src/rust/src/lazy/dsl.rs
```diff
@@ -2057,6 +2058,66 @@ impl Expr {
             .into())
     }
 
+   pub fn bin_contains(&self, lit: &[u8]) -> Self {
+        self.0.clone().binary().contains_literal(lit).into()
+    }
+
+    pub fn bin_starts_with(&self, sub: &[u8]) -> Self {
```
sorhawell, line 2065: you should use the robj_to! macro to improve the error messages
```rust
pub fn bin_starts_with(&self, sub: Robj) -> Result<Self, String> {
    Ok(self.0.clone().binary().starts_with(robj_to!(Raw, sub)?).into())
}
```

would look like this 

```R
 pl$col("a")$bin$starts_with( 77)
Error: the arg [sub] is not an R raw as required, but 77.0 
 when calling :
 unwrap(pl$col("a")$bin$starts_with(77))
```


to be fair the macro did support raw yet so I have added it in this PR suggestion branch, you can merge in the macro impl from this branch https://github.com/pola-rs/r-polars/commit/60abd7533214f851a2e7f6b6ef116111fcffa069


...remember to add a unwrap in R side impl ExprBin_starts_with

R/expr__binary.R
```diff
@@ -0,0 +1,75 @@
...
+
+
+#' contains
+#' @name ExprBin_contains
+#' @aliases expr_bin_contains
+#' @description R Check if binaries in Series contain a binary substring.
+#' @keywords ExprBin
+#' @param lit The binary substring to look for
+#' @return Expr returning a Boolean
+ExprBin_contains = function(lit) {
```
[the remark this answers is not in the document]
sorhawell: you do not need to prove polars is completely correct. Just check that if you change an arg input in implemented functions, the output is the correct behavior. Also check wrong argument type give error messages.

src/rust/src/lazy/dsl.rs
sorhawell: in the macro impl I suggest to use `Vec<u8>` instead of &[u8] for now for simplicity sake. To use a slice is theoretically faster, but unlikely in any practical use case.

sorhawell (changes requested):

sorhawell:
@xendai66 
Thank you, this looks very fine. I will merge it in Monday.

sorhawell:
many thx @xendai66   💯 👍

PR #103, opened 2023-04-11
DataFrame$tail()

vincentarelbundock:
This is a first attempt at a minimal commit. Is this the kind of contribution you would welcome? What should I do different to facilitate a merge?

``` r
library(rpolars)

pl$DataFrame(mtcars)$tail(2)
# polars DataFrame: shape: (2, 11)
# ┌──────┬─────┬───────┬───────┬─────┬─────┬─────┬──────┬──────┐
# │ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ ... ┆ vs  ┆ am  ┆ gear ┆ carb │
# │ ---  ┆ --- ┆ ---   ┆ ---   ┆     ┆ --- ┆ --- ┆ ---  ┆ ---  │
# │ f64  ┆ f64 ┆ f64   ┆ f64   ┆     ┆ f64 ┆ f64 ┆ f64  ┆ f64  │
# ╞══════╪═════╪═══════╪═══════╪═════╪═════╪═════╪══════╪══════╡
# │ 15.0 ┆ 8.0 ┆ 301.0 ┆ 335.0 ┆ ... ┆ 0.0 ┆ 1.0 ┆ 5.0  ┆ 8.0  │
# │ 21.4 ┆ 4.0 ┆ 121.0 ┆ 109.0 ┆ ... ┆ 1.0 ┆ 1.0 ┆ 4.0  ┆ 2.0  │
# └──────┴─────┴───────┴───────┴─────┴─────┴─────┴──────┴──────┘

tail(mtcars, 2)
#                mpg cyl disp  hp drat   wt qsec vs am gear carb
# Maserati Bora 15.0   8  301 335 3.54 3.57 14.6  0  1    5    8
# Volvo 142E    21.4   4  121 109 4.11 2.78 18.6  1  1    4    2
```

### DIFF
--- a/R/lazyframe__lazy.R
+++ b/R/lazyframe__lazy.R
@@ -222,7 +222,6 @@ LazyFrame_limit = function(n) {
 #'
 #' @return A new `LazyFrame` object with applied filter.
 LazyFrame_tail = function(n) {
-  if(!is.numeric(n)) stopf("tail: n must be numeric")
   unwrap(.pr$LazyFrame$tail(self,n))
 }
 
--- a/src/rust/src/lazy/dataframe.rs
+++ b/src/rust/src/lazy/dataframe.rs
@@ -4,6 +4,7 @@ use crate::rdatatype::new_join_type;
 use crate::utils::r_result_list;
 use crate::utils::try_f64_into_u32;
 use crate::utils::try_f64_into_usize;
+use crate::robj_to;
 use extendr_api::prelude::*;
 use polars::prelude as pl;
 
@@ -70,12 +71,8 @@ impl LazyFrame {
         )
     }
 
-    fn tail(&self, n: f64) -> List {
-        r_result_list(
-            try_f64_into_u32(n)
-                .map(|n| LazyFrame(self.0.clone().tail(n)))
-                .map_err(|err| format!("tail: {}", err)),
-        )
+    fn tail(&self, n: Robj) -> Result<LazyFrame, String> {
+        Ok(LazyFrame(self.0.clone().tail(robj_to!(u32, n)?)))
     }
 
     fn filter(&self, expr: &Expr) -> LazyFrame {
--- a/tests/testthat/test-dataframe.R
+++ b/tests/testthat/test-dataframe.R
@@ -429,5 +429,5 @@ test_that("to_Struct, unnest, to_frame, as_data_frame", {
 test_that("tail", {
   a = as.data.frame(pl$DataFrame(mtcars)$tail(6))
   b = tail(mtcars)
-  expect_true(all(a == b))
+  expect_equal(a, b, ignore_attr = TRUE)
 })
--- a/tests/testthat/test-lazy.R
+++ b/tests/testthat/test-lazy.R
@@ -100,4 +100,10 @@ test_that("lazy filter", {
 })
 
 
+test_that("tail", {
+  a = pl$DataFrame(mtcars)$lazy()$tail(6)$collect()$as_data_frame()
+  b = tail(mtcars)
+  expect_equal(a, b, ignore_attr = TRUE)
+})
+
 #TODO complete tests for lazy


sorhawell:
Hi @vincentarelbundock. Thank you so much for finding time for writing a translation :)

This is pretty alright. I would take the PR as it is right now. 

I will add some suggestions in the review, I came to appreciate, when I wrote the later translations.

src/rust/src/lazy/dataframe.rs
```diff
@@ -70,6 +70,14 @@ impl LazyFrame {
...
     }
 
+    fn tail(&self, n: f64) -> List {
+        r_result_list(
+            try_f64_into_u32(n)
+                .map(|n| LazyFrame(self.0.clone().tail(n)))
+                .map_err(|err| format!("tail: {}", err)),
+        )
+    }
+
```
sorhawell, line 80: this can now be expressed in a more ergonomic way. But essentially the same

robj_to! converts or write a suitable err msg, it also adds in the name of argument which failed

when using robj_to! I think it is fine to just skip testing of error handling on every single method, robj_to! is already tested plenty.

```rust
fn tail(&self, n: Robj) -> Result<LazyFrame, String> {
        Ok(LazyFrame(self.0.clone().tail(robj_to!(u32, n)?)))
}
```
err example
```R
> pl$DataFrame(iris)$lazy()$tail(n=-3)

Error: the arg [n] the value -3 cannot be less than zero
when calling :
pl$DataFrame(iris)$lazy()$tail(n = -3)
```

tests/testthat/test-dataframe.R
```diff
@@ -424,3 +424,10 @@ test_that("to_Struct, unnest, to_frame, as_data_frame", {
   expect_identical(df$as_data_frame(), df_e)
 })
 
+
+
+test_that("tail", {
+  a = as.data.frame(pl$DataFrame(mtcars)$tail(6))
+  b = tail(mtcars)
+  expect_true(all(a == b))
```
sorhawell, line 432: you could choose to use expect_identical(a,b) or expect_equal(a,b)

tests/testthat/test-dataframe.R
```diff
@@ -424,3 +424,10 @@ test_that("to_Struct, unnest, to_frame, as_data_frame", {
   expect_identical(df$as_data_frame(), df_e)
 })
 
+
```
sorhawell, line 427: you should add a similar test in test-lazy.R to formally test lazy tail

R/lazyframe__lazy.R
```diff
@@ -213,6 +213,18 @@ LazyFrame_limit = function(n) {
...
+#' @title Tail
+#' @description take last n rows of query
+#' @keywords LazyFrame
+#' @param n positive numeric or integer number not larger than 2^32
+#'
+#' @details any number will converted to u32. Negative raises error
+#'
+#' @return A new `LazyFrame` object with applied filter.
+LazyFrame_tail = function(n) {
+  if(!is.numeric(n)) stopf("tail: n must be numeric")
```
sorhawell, line 225: most R side type checks becomes redundant when using robj_to! and can just be dropped.

tests/testthat/test-dataframe.R
vincentarelbundock: done

R/lazyframe__lazy.R
vincentarelbundock: Done

src/rust/src/lazy/dataframe.rs
vincentarelbundock: thanks, this is a very nice shortcut.

vincentarelbundock:
Thanks for the hints. I think the latest commit resolves all outstanding issues.

I still don't quite understand what is going on under the hood, but I'll try to translate a few other functions. Will open a separate PR for those.

sorhawell:
> Thanks for the hints. I think the latest commit resolves all outstanding issues.
> 
> I still don't quite understand what is going on under the hood, but I'll try to translate a few other functions.

**Error handling in rpolars:**
It is not possible to raise R errors from rust side with extendr without bleeding memory. Instead an Result-enum `Result<LazyFrame, String>` is returned to R side as an R list `list(ok=NULL, err = "oups!")`. On R side typically the Result is just `unwrap()`'ed immediately to raise the error.
`r_result_list()` is the old less ergonomic way to convert Result to R list. Now rpolars use[ a PR branch of extendr](https://github.com/extendr/extendr/pull/473) where the conversion to R list will occur implicitly, as you have done here.

**robj_to!:**
robj_to! is just a macro which do custom conversions from R to rust types + error handling. The vanilla extendr conversions aren't bad, but not always ideal and error messages are very sparse. By using robj_to! all conversions and err handling can be modified throughout the code base at one single place.
robj_to! basically dispatches the appropriate conversion function and do error handling. There is a candidate master student project to upgrade the error handling this summer.

sorhawell:
many thanks @vincentarelbundock  for this excellent contribution :)

vincentarelbundock:
Thanks for the merge and the info, this is very useful.

Issue #104, opened 2023-04-11
S3 methods for simple operations


vincentarelbundock:
Thanks for the great package. Impressive work\!

I have been playing with `rpolars` over the past few days and have been thinking about how we could make the package more approachable for users who are familiar with base `R` but not the interface of this particular package. One thing that I feel would be very useful for inexperienced users would be to  ship a set of simple S3 methods that implement some of the most popular and standard base R operations on `rpolars` objects. For example:

``` r
library(rpolars)

head.DataFrame = function(x, n = 6L, ...) {
  x$limit(n = n)
}

df = pl$DataFrame(mtcars)

head(df)
# polars DataFrame: shape: (6, 11)
# ┌──────┬─────┬───────┬───────┬─────┬─────┬─────┬──────┬──────┐
# │ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ ... ┆ vs  ┆ am  ┆ gear ┆ carb │
# │ ---  ┆ --- ┆ ---   ┆ ---   ┆     ┆ --- ┆ --- ┆ ---  ┆ ---  │
# │ f64  ┆ f64 ┆ f64   ┆ f64   ┆     ┆ f64 ┆ f64 ┆ f64  ┆ f64  │
# ╞══════╪═════╪═══════╪═══════╪═════╪═════╪═════╪══════╪══════╡
# │ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ ... ┆ 0.0 ┆ 1.0 ┆ 4.0  ┆ 4.0  │
# │ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ ... ┆ 0.0 ┆ 1.0 ┆ 4.0  ┆ 4.0  │
# │ 22.8 ┆ 4.0 ┆ 108.0 ┆ 93.0  ┆ ... ┆ 1.0 ┆ 1.0 ┆ 4.0  ┆ 1.0  │
# │ 21.4 ┆ 6.0 ┆ 258.0 ┆ 110.0 ┆ ... ┆ 1.0 ┆ 0.0 ┆ 3.0  ┆ 1.0  │
# │ 18.7 ┆ 8.0 ┆ 360.0 ┆ 175.0 ┆ ... ┆ 0.0 ┆ 0.0 ┆ 3.0  ┆ 2.0  │
[...]

PR #107, opened 2023-04-12
S3 methods for DataFrame and LazyFrame: simplest

vincentarelbundock:
Issue https://github.com/pola-rs/r-polars/issues/104

Notes:

* S3 methods have `@noRd` tags because I did not want to pollute the manual. But for some reason skipping `nrow` and `ncol` led `check()` to issue a warning, so I included minimal docs for just those two.

TODO in a different PR:

  - [ ] brackets – lots of work\!
  - [ ] is.null (new method needed)
  - [ ] na.omit (new method needed)
  - [ ] colnames (new method needed)

This PR implements a few of the main base R S3 methods for `DataFrame` and `LazyFrame` objects. For example:

``` r
library(rpolars)
d = pl$DataFrame(mtcars)
dl = pl$DataFrame(mtcars)$lazy()

dim(d)
# [1] 32 11

head(dl, 3)
# [1] "polars LazyFrame naive plan: (run ldf$describe_optimized_plan() to see the optimized plan)"
#   SLICE[offset: 0, len: 3]
#     DF ["mpg", "cyl", "disp", "hp"]; PROJECT */11 COLUMNS; SELECTION: "None"

names(d)
#  [1] "mpg"  "cyl"  "disp" "hp"   "drat" "wt"   "qsec" "vs"   "am"   "gear"
# [11] "carb"

nrow(dl)
# [1] 32

ncol(dl)
# [1] 11

as.matrix(d) |> str()
#  num [1:32, 1:11] 21 21 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 ...
#  - attr(*, "dimnames")=List of 2
#   ..$ : NULL
#   ..$ : chr [1:11] "mpg" "cyl" "disp" "hp" ...
```


sorhawell:
>implement attributes and methods instead of LazyFrame()$collect()

I think e.g. nrow cannot be inferred without collect, however it could be implemented something like where only the count is returned. If polars will optimize away if possible.

```R
x$count()$collect()
```

vignettes/polars.Rmd
```diff
@@ -51,88 +59,106 @@ dependencies other than R itself. This makes it very fast and lightweight to
...
 
 ```
 # Python
-df.select([pl.col(["A", "B"]).sum()]))
+df.group_by("id").mean()
 
 # R
-df$select(pl$col(c("A", "B"))$sum())
+df$groupby("id")$mean()
 ```
```
eitsupi, line 89: Could you please split the code block between Python and R and set the info string?

like

````rmd
```python
df.method
```

```r
df$method
```
````

Comment (vincentarelbundock) [reply vignettes/polars.Rmd:86]: d5c256fdfa359a403bcce662a7cec7ff04b1de24

**Comment by vincentarelbundock** [comment]
I have completed all I planned to achieve in this PR, so feel free to take a look when you find some time. As always, I'm happy to make any changes you'd like.

The cool thing about my two recent PRs is that it allowed me to considerably simplify the Get Started vignette (@grantmcdermott may want to take a look). Here's a rendered version for your convenience:

https://arelbundock.com/polars.html

Here are the main things I changed in the vignette:

* Added a couple examples of standard base R functions like `head(df)`.
* The first Polars methods users now encounter are the ultra-simple and "clean" ones that we just merged, like `dat$tail()` and  `dat$mean()`.
* Introduce methods chaining using those same ultra-simple methods: `dat$tail(10)$mean()`. I used this example to replace the old Hello World. It was cute and funny, but overly complicated. I must admit that the first time I saw that I had a pretty bad reaction to the ugly syntax. In editing, I thought it was best for the first examples to be simple and actually useful.
* Push the discussion of data types to the end. This is fundamental and useful, but "Get Started" is also a "sales pitch" and I wanted to give users a payoff as fast as possible.

**Comment by sorhawell** [comment]
> For what it’s worth, if we didn’t care about overwriting the columns, the previous query could have been written more concisely as:

"For what it’s worth, the previous query could have been written more concisely as"
```R
dat = pl$DataFrame(mtcars)
dat$with_columns(
  pl$col(c("mpg", "hp"))$sum()$over("cyl")$prefix("sum_")
)
```

#BTW also possible but not fully stabilized, so do not recommend in tutorial

```R

# has currenyly a thread-safety user warning. But could be removed now.
dat$with_columns(
  pl$col(c("mpg", "hp"))$sum()$over("cyl")$map_alias(\(x) paste0("sum_",x))
)

# has not stabilized becaues does not work well with wild-cards like $all()
pl$set_polars_options(named_exprs = TRUE) # must be activted like this first
dat$with_columns(
  sum_mpg = pl$col("mpg")$sum()$over("cyl"),
  sum_hp  = pl$col("hp")$sum()$over("cyl")
)


#could be fun to try allow this also :)  as an opt-in feature. All column names converted to `pl$col(x)` and in scope of the method. Not sure it would work well for lazy API though.

dat$with_columns(
  sum_mpg = mpg$sum()$over(cyl),
  sum_hp  = hp$sum()$over(cyl)
)
```

sorhawell:
Half-way through the [revised tutorial ](https://arelbundock.com/polars.html). It has really been a pleasure to read. I agree with your changes. I will add a few syntax edits.

grantmcdermott:
> Half-way through the [revised tutorial ](https://arelbundock.com/polars.html). It has really been a pleasure to read. I agree with your changes. I will add a few syntax edits.

Agree, lots to like here. Will leave a few comments for you to think about.

vignettes/polars.Rmd
```diff
@@ -51,88 +56,105 @@ dependencies other than R itself. This makes it very fast and lightweight to
...
+head(dat, n = 2)
 ```
 
-Recall that Polars and R are both column-orientated. So you can think of
-`DataFrames` (data.frames) as being made up of a collection of `Series`
-(vectors). In fact, you can create a new Polars `DataFrame` as a mix of `Series`
-and/or regular R vectors.
+## Methods and pipelines
+
+Although some simple R functions work out of the box on **polars** objects, the full power of Polars is realized via _methods_. Polars methods are accessed using the `$` syntax. For example, to convert Polars `Series` and `DataFrames` back to standard R objects, we use the `$to_r_vector()` and `$as_data_frame()` methods:
```
grantmcdermott, line 137: It's been a while since I tried this, but is `as_data_frame` still a tibble-exported (or maybe arrow-exported) function? Just worth thinking ahead to whether users might run into a NAMESPACE clash.

vignettes/polars.Rmd
```diff
@@ -51,88 +56,105 @@ dependencies other than R itself. This makes it very fast and lightweight to
...
+```python
 # Python
-df.select([pl.col(["A", "B"]).sum()]))
+df.group_by("id").mean()
+```
 
+```r
 # R
-df$select(pl$col(c("A", "B"))$sum())
+df$groupby("id")$mean()
```
grantmcdermott, line 85: Not saying you should revert, but my motivation for including two columns in the original example was to show the users would also need to switch over to R's native `c()` method (i.e., in addition to switching to `$`).

vignettes/polars.Rmd
```diff
@@ -390,3 +441,21 @@ pl$DataFrame(iris)$select(
...
+```
+
+The full list of valid Polars types can be found by typing `pl$dtypes` 
+into your R console. These include _Boolean_, _Float32(64)_, _Int32(64)_,
+_Utf8_, _Categorical_, _Date_, etc. Note that some type names differ from what
+they are called in R (e.g., _Boolean_ in Polars is equivalent to `logical()` in
+R). This might occasionally require you to look up a specific type. But the good
+news is that **polars** generally does a good job of inferring types
+automatically.
+
```
grantmcdermott, line 461: On balance, I agree that moving types to the end makes sense for an intro vignette. But we've already included a type coercion above via `cast`. So it might be an idea to either flag it above, or at least mention something about type coercion here again.

grantmcdermott (commented):
Lots to like here @vincentarelbundock.

One thing I haven't mentioned as a specific comment---but I think is good for consistency---is sticking to the hard 80 character wrapping. I know that you're a fellow vim user and [this is the hook](https://vim.fandom.com/wiki/Format_only_long_lines) I always use ;-)

vignettes/polars.Rmd
```diff
@@ -17,12 +17,28 @@ options(rmarkdown.html_vignette.check_title = FALSE)
 
 ## What is Polars?
 
-[Polars](http://pola.rs) is among the
-[fastest](https://h2oai.github.io/db-benchmark/) data table query 
-libraries available in any language. The underlying computation engine is
-written in Rust and can be used directly from via the polars crate. Orginally Polars bindings were implemented for Python. The **polars** R-package also provides equivalent bindings from R. Within
-R, python and rust each polars package is just called polars. The implemention polars in the
-different languages can be called rust-polars, py-polars, r-polars, nodejs-polars ...
+Polars is [a lightning fast](https://h2oai.github.io/db-benchmark/) Data Frame
```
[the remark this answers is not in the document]
vincentarelbundock: I will update the link. And wow, DuckDB has really improved! Do you think we should change the 5-10 improvement relative to dplyr in light of those results? Some operations are 50x faster...

vignettes/polars.Rmd
vincentarelbundock: There's no clash between exported functions and polars methods, though, right? And this PR adds a proper `as.data.frame()` s3 for polars objects, which just calls `df$as_data_frame()` under the hood.

vignettes/polars.Rmd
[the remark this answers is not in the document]
grantmcdermott: Good question. We could hedge it with a classic "orders of magnitude(s) improvement".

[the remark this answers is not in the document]
vincentarelbundock: Good call.

vignettes/polars.Rmd
vincentarelbundock: Right that's useful info. But given that we haven't shown the user *anything* yet and it's their very first contact with **polars**, I feel that absolute simplicity and clarity serves our pitch better at this point. It's also not too hard to figure out `c()` in time when users try it out. That said, I don't feel extremely strongly about this, and could be swayed.

vincentarelbundock: Also, there is an example with `dat$select(pl$col(c("mpg", "hp")))` a little further down the page. So I think that's enough.

vincentarelbundock:
FYI, I believe I have addressed all the points raised in review so far. Feel free to make more requests if needed.

vignettes/polars.Rmd
grantmcdermott: Cool cool. Just wanted to check.

sorhawell:
Many thanks @vincentarelbundock  :)

Issue #111, opened 2023-04-13
Potential improvements in tests?


etiennebacher:
Hi @sorhawell, I've been looking a bit at the test suite of `polars` (since I'm not good enough in Rust for now ;)) and I think there are some things there that could be improved. Below is a small list of potential improvements, I prefer posting here first so that (1) we can discuss some of these before spending too much time on this, and (2) other people can contribute to it.

- [x] improve the test coverage. Currently (on my Ubuntu laptop), `covr::report()` says that there is 23.96% of coverage (maybe the "weird" syntax of functions with dollars in name makes coverage detection harder, but my point is that not everything is covered). To me, this is the most important potential improvement but also the one that will take the most time.
- [x] use [snapshots](https://testthat.r-lib.org/articles/snapshotting.html) instead of manually hardcoded expectations. Below is an example of expectation that could be replaced by a snapshot so that it's easier to update if the formatting ever changes.
https://github.com/pola-rs/r-polars/blob/28c3cfeaa2d0834b41e4539e647acec2683a95a2/tests/testthat/test-dataframe.R#L12-L25
- [x] remove uses of `testthat::` (since `testthat` is automatically loaded by `devtools::load_all()`)
- [x] similarly, `polars:::` shouldn't be needed in tests because `devtools::load_all()` also loads internal functions for tests 
- [x] use more suited `expect_` functions, e.g replace `expect_identical(whatever, FALSE)` by `expect_false(whatever)` 
- [x] add some regex to `expect_error()` or `expect_warning()` (arg `regex`) to ensure that the message is really what we expect

There are maybe more but I didn't review all tests for now. As you can see, apart from increasing the coverage, most of these suggestions are here to make the tests easier to read and therefore reduce the cognitive load when something breaks.

Let me know what you think ;)

sorhawell:
> says that there is 23.96% of coverage

I hoped we were in the >90% for public methods and functions. I suspect covr may include both e.g.
I rarely commit a method without an expectation

```R
polars:::wrap__Expr__abs
polars:::Expr_abs
Expr$abs
```

[The current public class methods and functions are about all these (except for sub namespaces):](https://rpolars.github.io/reference/pl.html)

I would be very useful with a covr workflow, and I do think we should aim for  100% coverage

sorhawell:
> use more suited expect_ functions, e.g replace expect_identical(whatever, FALSE) by expect_false(whatever)

[...]

PR #113, opened 2023-04-14
Examples for test improvements

etiennebacher:
@sorhawell Here's an example for #111. I modified two test files so that you can tell me if you're ok with this. Here are some notes:

* I moved some custom `expect_` from the `R` folder to `tests/testthat/helper.R` because this file is loaded every time you want to run the tests and it's where custom expectations and test fixtures should be (this also means that there's no need to use `polars:::` in the tests)
* I'm still not sure what `expect_grepl_error()` adds compared to `expect_error()`, maybe you have an example?
* I noticed a strange expectation when you test `rechunk` at the end of `test-series.R` (see comment in code)

Otherwise, mostly style and some things detailed in #111. Speaking about style, I preferred putting the chain `$` at the end of each line because of the nice indentation:
```r
pl$
  DataFrame(int_iris)$
  select(
    pl$col("Sepal.Length")$
      map(\(s) s)
  )$
  as_data_frame()
```
but maybe you have other preferences? Not crucial but I think it's better to stay consistent in docs / tests.

What do you think?


sorhawell:
> I preferred putting the chain $ at the end

Maybe what you suggest could be thought of as "dollar-sign-in-right-side" style. Looks good. I like aesthetically "left-side"-style but find it inconvenient for scripting and debugging. I usually gravitate towards "middle-style" or "ad-hoc-style" when scripting.

I think r-polars is too young to recommend a style yet. Write what you find is beautiful or practical and inspire the rest of us :)
py-polars also needs a parenthesis to do "left-side"-style and.

```R

#left-side style, (requires parenthesis), looks like spark when written in scala or rust method chaning I think
(
  pl$DataFrame(iris)
    $lazy()
    $select(
      pl$col("Species")$cast(pl$Utf8)$str$encode("base64")$alias("base64"),
      pl$all()
    )
    $collect()
)

#left-side style + also for column expressions
(
  pl$DataFrame(iris)
    $lazy()
    $select(
      pl$col("Species")
        $cast(pl$Utf8)
        $str$encode("base64")
        $alias("base64"),
      pl$all()
    )
    $collect()
)

#middle style
 pl$DataFrame(
    iris
  )$lazy(
  )$select(
    pl$col("Species")$cast(pl$Utf8)$str$encode("base64")$alias("base64"),
    pl$all()
  )$collect()
 
#ad-hoc style
pl$DataFrame(iris)$lazy()$select(
  pl$col("Species")$cast(pl$Utf8)$str$encode("base64")$alias("base64"),
  pl$all()
)$collect()

```

sorhawell:
> I'm still not sure what expect_grepl_error() adds compared to expect_error(), maybe you have an example?

I think I was dissatisfied with the call stack native R sometimes showed and I just added the call to error string. But then all regexpr matching are potential false-negatives. See this example. Probably my errors should have been more refined in the first place.


```R
#the error
pl$lit(1:4)$add("a")$to_r()

#test if "Arithmetic" is in the err message, hmm no failure, seems fine
expect_error( pl$lit(1:4)$add("a")$to_r(),regexp = "Arithmetic")


#on strange_fruit does not raise a failure, because "strange_fruit" is a part of the call
expect_error( pl$lit(1:4)$add("a")$to_r(),regexp = "strange_fruit")

#grepl error removes the "when calling ..." from message
polars:::expect_grepl_error(pl$lit(1:4)$add("a")$to_r(),"strange_fruit")

#also grepl error
polars:::expect_grepl_error(pl$lit(1:4)$add("a")$to_r(),c("Arithmetic","explicit"))
```

sorhawell:
nice spotted `expect_strictly_identical`is not needed anymore if testthat + waldo is upgraded

etiennebacher:
> See this example.

Then a simple regex does the job:

``` r
library(polars)
library(testthat)

# no error, no message
expect_error(
  pl$lit(1:4)$add("a")$to_r(),
  regexp = "Arithmetic .* explicit"
)
```
But in any case this kind of expectations will probably be revised after the rework on error handling this summer. If you're ok with the rest, you can merge this PR and I will make other ones later for the rest of tests

sorhawell:
great many thanks @etiennebacher  :)

vincentarelbundock:
> Maybe what you suggest could be thought of as "dollar-sign-in-right-side" style. Looks good.

FWIW, I think hat Etienne's right-side style has *a ton* of pedagogical/marketing value for a community that is ultra-familiar with `%>%` pipes. And it makes a lot of the code look nearly 1-to-1 with `dplyr`:

```r
df$
  filter()$
  select()$
  groupby()

df |>
  filter() |>
  select() |>
  group_by()
```

PR #120, opened 2023-04-17
Clean tests

### LINKED ISSUE #111: Potential improvements in tests?

etiennebacher:
Close #111 

Mostly styling, replacing some expectation functions by other which are more suited, and add some regex for errors whose message is unlikely to change. I gave up styling the chains (cf #113) because there are just too many and there's no automated way to do it (e.g `styler`).

I've been through half of the tests so far and the ~25% coverage reported in #111 seems largely underestimated to me (probably due to the strange construction of functions), there is an amazing amount of tests.

**Still WIP** (stopped at `test-expr_datetime.R`, included), do not pay attention to potential test failures for now


sorhawell:
@etiennebacher  bare with me, a real big personal hobby-horse coming up :)

I am personally a fan of `data.table` and love that the code base is written predominantly with the `=`assignment operator. I think some parts of the  R community have been pushing the `<-` to an extend, where it is implicitly hinted that the  `=` operator is an inherently defunct operator and a code smell.

I would be proud if some belittled R user one day would try to counter an argument by saying "it is really ok, see r-polars does it too".

How about tests are free game, but the translations predominantly uses `=` ? :)

I searched many time but ever only seen below two code examples to argue "`=`-bad", which I have not yet understood is an issue, because both behaviors are discouraged by the same style guides anyways.

```R
> x=1
> z <- y = x
Error in z <- y = x : could not find function "<-<-"
```
 and because of this
 ```R
 > f = function(x) print(x)
> f(x=2)
[1] 2
> f(x<-4)
[1] 4
> x
[1] 4
```


what do you think? :)

File: tests/testthat/test-bit64.R [hunk 1]
```diff
@@ -1,81 +1,76 @@
...
-    expect_grepl_error(
-      unwrap(polars:::test_robj_to_u32(bit64::as.integer64(-1))),
-      "the value -1 cannot be less than zero"
-    )
+  # NO NA
+  expect_error(
+    unwrap(test_robj_to_i64(bit64::as.integer64(NA))),
+    "scalar arguments do not support integer64 NA value"
+  )
+  expect_error(
```
sorhawell, line 50: If sytematically replacing the `expect_grepl_error` with `expect_error` one need to address that `expect_error` by it self cannot verify r-polars error messages, as the literally expected text will be included in the call also, and the returned error message can never fail. This silently breaks the usefulness of the test.

Options are: rework unwrap() and the provided context of error message or manually turn on/off the same option expect_grepl_error is doing automatically.

etiennebacher:
> How about tests are free game, but the translations predominantly uses = ? :)

I don't have any problems with using `=` instead `<-` in the code (I used to prefer `<-` because that's what I was taught but I saw many popular packages use `=` like `data.table` so I don't really care anymore ;) ).

The thing is simply that I ran `styler::style_dir()` on the tests to fix some styling (add a space after a comma, or before and after a `=`, etc.) and it also replaced `=` by `<-`. I will revert this change at the end, when this PR is ready for review.

tests/testthat/test-bit64.R
etiennebacher: I think I misunderstood you before, I didn't realize that:
```r
# NO NA
expect_error(
    unwrap(test_robj_to_i64(bit64::as.integer64(NA))),
    "hello there"
)
```
would not fail. I'll revert the changes made on `expect_grepl_error()`

eitsupi:
FYI, there is a well-known package `knitr` that uses `=` as the assignment operator, and I believe the `knitr` code is formatted with `formatR`.
https://github.com/yihui/formatR

sorhawell:
I do like formatters. I have not used  it in R yet. I will try it out :)

eitsupi:
I think that the `styler` is probably more popular, although it can probably be configured to use `=` for the assignment operator, but I rarely write settings for `styler` because I have to write the settings in R scripts.
https://stackoverflow.com/questions/50660401/use-instead-of-for-assignment-when-styling-r-code-with-styler

vincentarelbundock:
> I searched many time but ever only seen below two code examples to argue "`=`-bad", which I have not yet understood is an issue, because both behaviors are discouraged by the same style guides anyways.

@sorhawell Here's a video of Ross Ihaka saying we should all be using `=`. Thought you might get a kick out of that 😉 

https://www.youtube.com/watch?v=88TftllIjaY&t=2096s

eitsupi:
I prefer to let the auto-formatter (styler) do the work rather than obsessing over something.
I don't think there is much value in getting hung up on manual formatting here.

etiennebacher:
There's actually not much useful in this PR, most of it is styling and indeed it's better to let `styler` or something else do it. Probably better to close this, there are other priorities (but I let #111 open)

sorhawell:
@vincentarelbundock 

> Thought you might get a kick out of that

Good video also :) I have dream of moving to NZ for two years.

`<-` people and `=` people, all this infighting is unnecessary. We we have so much in common being "left assignment oriented", together we can agree that the "right assignment" `->` people are the odd ones!  😆

sorhawell:
@eitsupi 

I will try to use styler.equals + rstudio addin

https://github.com/Robinlovelace/styler.equals/issues/1

eitsupi:
> @eitsupi
> 
> I will try to use styler.equals + rstudio addin
> 
> [Robinlovelace/styler.equals#1](https://github.com/Robinlovelace/styler.equals/issues/1)

Wonderful! I could not find it.

PR #121, opened 2023-04-17
More `DataFrame` translations

vincentarelbundock:
Polars methods

- [x] `unique`
- [x] `fill_null`
- [x] `drop`
- [x] `drop_in_place`
- [x] `drop_nulls`
- [x] `frame_equal`
- [x] `shift`
- [x] `shift_and_fill`
- [x] `quantile`
- [x] `fill_nan`

S3 methods

- [x] `unique`
- [x] `na.omit`

Will not do this in the current PR after all:

- [x] `sort`
- [ ] `is_duplicated`
- [ ] `is_unique`


sorhawell:
this is amazing :) I didn't get to complete my review but it looks very good. I will continue tomorrow night.

sorhawell:
almost there still missing some tests and more examples

vincentarelbundock:
Do you plan to work on those or would you like me
to take a look.

No rush or pressure either way.

sorhawell:
> Do you plan to work on those or would you like me to take a look?

I would be happy for that.

By the way. What do you think of the wording of docs @param `by` and `...`? I think it is strange to an average R user, but fairly common for a python or rust user. I'm not sure how to state it intuitively.
Iterable Into<Expr> in practice means some R object that `lapply()` can iterate over and which is either already an Expr or can be converted with `pl$lit()` and/or `pl$col()`without errors.

vincentarelbundock:
I added some tests and lots of examples. I also tried to simplify the `by` argument documentation, but I don't think my version is great either. But I think the numerous examples should help clarify things for most users.

sorhawell:
@vincentarelbundock  great this is a huge contribution :) I will merge in after checks

sorhawell (approved):

sorhawell:
@vincentarelbundock likely after merging #140  and #146 into this PR failed. I tried to resolve it but think the solution drifted away from you vision of the s3 methods.

Can you solve the errors?

It might also be better for LazyFrame to not implement the "forbidden" collect methods nrow, dim, ... than throwing an R error. Some auto completion method in an IDE might call e.g. dim on any method.

vincentarelbundock:
Should be good to go now.

R/expr__expr.R
```diff
@@ -52,6 +52,16 @@ Expr_print = function() {
...
 
+#' @title as.list Expr
+#' @description wraps an Expr in a list
+#' @param x Expr
+#' @param ... not used
+#' @export
+#' @keywords Expr
+as.list.Expr = function(x, ...) {
+  list(x)
+}
```
eitsupi, line 63: I do not understand the need for this.

PR #122, opened 2023-04-17
mention r2u

sorhawell:


README.Rmd
```diff
@@ -108,6 +108,15 @@ require a different target or operating system build. Finally, see the bottom of
 this README for details on how to install rust to build from source (only
 relevant for developers, or users of unsupported operating systems).
 
+
+### r2u
+Speeding up your workflow? On Ubuntu, install polars + arrow from binaries and resolve
+system dependencies reliably and quickly with r2u ([see link for configuration](https://eddelbuettel.github.io/r2u/)).
+```r
```
vincentarelbundock, line 115: @eddelbuettel Does the user need to explicitly call `bspm::enable()` for this to work?

eddelbuettel: Yes and no. 

As you know, we detail a five step configuration at the r2u site, of the five steps one is getting the r2u apt repo, one is getting a current R, and so on.  Now _if_ you want to use the .deb binaries directly from R via `install.packages()` then yes, `bspm` is both needed and has to be enabled to use the repos information for binaries.  

So at is written here in the README.{md,Rmd} files you need it.  Hence the pointer to the docs.

eddelbuettel (approved):
This looks good to me.  (With the standard caveat that one has to have spent the minute setting up client side of a Ubuntu 22.04 system to be ready for r2u.)  

So I suggest to merge and :ship: this.

README.Rmd
vincentarelbundock: Yeah, that's kind of what I thought. In that context, I'd argue it probably makes sense to add `bspm::enable()` just before the `install.packages` call.

vincentarelbundock: Makes sense. thanks.

PR #125, opened 2023-04-18
feat: add knit_print method for DataFrame (experimental)

eitsupi:
Add the ability to output pretty HTML tables on knitr documents.


eitsupi:
I am confused about the errors in the test.
Locally the test does not fail.

R/pkg-knitr.R
```diff
@@ -0,0 +1,23 @@
+#' @export
+knit_print.DataFrame = function(x, ...) {
+  .env_formatting = Sys.getenv("POLARS_FMT_TABLE_FORMATTING")
+  .env_inline_data_type <- tolower(Sys.getenv("POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE"))
```
eitsupi, line 4:
```suggestion
  .env_inline_data_type = tolower(Sys.getenv("POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE"))
```

tests/testthat/test-knitr.R
```diff
@@ -0,0 +1,23 @@
+.knit_file <- function(file_name) {
+  file <- file.path("files", file_name)
+  output <- tempfile(fileext = "md")
```
eitsupi, line 3:
```suggestion
  file = file.path("files", file_name)
  output = tempfile(fileext = "md")
```

R/pkg-knitr.R
```diff
@@ -0,0 +1,23 @@
...
+
+  if (.env_formatting %in% c("", "ASCII_MARKDOWN") && .env_inline_data_type %in% c("", "true", "1")) {
+    Sys.setenv(POLARS_FMT_TABLE_FORMATTING = "ASCII_MARKDOWN")
+    Sys.setenv(POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE = "1")
+  }
+
+  out <- capture.output(print(x))
+
+  # Needs to insert a blank line
+  out <- c(out[1], "", out[-1]) |>
```
eitsupi, line 19:
```suggestion
  out = c(out[1], "", out[-1]) |>
```

R/pkg-knitr.R
```diff
@@ -0,0 +1,23 @@
...
+    Sys.setenv(POLARS_FMT_TABLE_FORMATTING = .env_formatting)
+    Sys.setenv(POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE = .env_inline_data_type)
+  })
+
+  if (.env_formatting %in% c("", "ASCII_MARKDOWN") && .env_inline_data_type %in% c("", "true", "1")) {
+    Sys.setenv(POLARS_FMT_TABLE_FORMATTING = "ASCII_MARKDOWN")
+    Sys.setenv(POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE = "1")
+  }
+
+  out <- capture.output(print(x))
```
eitsupi, line 16:
```suggestion
  out = capture.output(print(x))
```

sorhawell:
> I am confused about the errors in the test. Locally the test does not fail.

I think I have had that error before also, I can try fiddle with it this evening.

sorhawell:
@eitsupi I also updated the snapshot, which had slightly different spacings on my machine. At worst polars could have some machine-specific inference of line widths which is good for users and bad for testing. In such case, maybe we can use a DataFrame with only 1-3 columns as test case.

eitsupi:
Thanks!

Definitely a small data frame seems to be the right target for snapshot testing here. I will update.

eitsupi:
Sorry, but I found that the current approach will fail if there is a column of list or long column name.
We can check with `dplyr::starwars` for example.

```r
> Sys.setenv(POLARS_FMT_TABLE_FORMATTING = "ASCII_MARKDOWN")

> Sys.setenv(POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE = "1")

> pl$DataFrame(dplyr::starwars)
shape: (87, 14)
| name (str)     | height (i32) | mass (f64) | hair_color | ... | species (str) | films          | vehicles       | starships     |
|                |              |            | (str)      |     |               | (list[str])    | (list[str])    | (list[str])   |
|----------------|--------------|------------|------------|-----|---------------|----------------|----------------|---------------|
| Luke Skywalker | 172          | 77.0       | blond      | ... | Human         | ["The Empire   | ["Snowspeeder" | ["X-wing",    |
|                |              |            |            |     |               | Strikes Back", | , "Imperial    | "Imperial     |
|                |              |            |            |     |               | "Rev...        | Speede...      | shuttle"]     |
| C-3PO          | 167          | 75.0       | null       | ... | Droid         | ["The Empire   | []             | []            |
|                |              |            |            |     |               | Strikes Back", |                |               |
|                |              |            |            |     |               | "Att...        |                |               |
| R2-D2          | 96           | 32.0       | null       | ... | Droid         | ["The Empire   | []             | []            |
|                |              |            |            |     |               | Strikes Back", |                |               |
|                |              |            |            |     |               | "Att...        |                |               |
| Darth Vader    | 202          | 136.0      | none       | ... | Human         | ["The Empire   | []             | ["TIE         |
|                |              |            |            |     |               | Strikes Back", |                | Advanced x1"] |
|                |              |            |            |     |               | "Rev...        |                |               |
| ...            | ...          | ...        | ...        | ... | ...           | ...            | ...            | ...           |
| Poe Dameron    | null         | null       | brown      | ... | Human         | ["The Force    | []             | ["T-70 X-wing |
|                |              |            |            |     |               | Awakens"]      |                | fighter"]     |
| BB8            | null         | null       | none       | ... | Droid         | ["The Force    | []             | []            |
|                |              |            |            |     |               | Awakens"]      |                |               |
| Captain Phasma | null         | null       | unknown    | ... | null          | ["The Force    | []             | []            |
|                |              |            |            |     |               | Awakens"]      |                |               |
| Padmé Amidala  | 165          | 45.0       | brown      | ... | Human         | ["Attack of    | []             | ["H-type      |
|                |              |            |            |     |               | the Clones",   |                | Nubian        |
|                |              |            |            |     |               | "The Ph...     |                | yacht",       |
|                |              |            |            |     |               |                |                | "Naboo s...   |
```

Perhaps we may need to write out the HTML in the same way as python-polars...
https://github.com/pola-rs/polars/blob/81b4102ac03f79361645e709f05061d9bb876a51/py-polars/polars/dataframe/_html.py#L48-L115

eitsupi:
Some additional functions need to be implemented.

- `[.DataFrame` (Like https://github.com/apache/arrow/blob/0bf777a5952be012e41f5b1ad443d4fec38e6f5a/r/R/arrow-tabular.R#L105)
- `dtype_strings` (maybe -> `typeof()`)
- `get_fmr` (maybe -> `as.charactor()`)

sorhawell:
@eitsupi 
It might be little slower then other polars functions, but probably fast enough for limited printing.
https://pola-rs.github.io/polars/polars/series/trait.SeriesTrait.html#tymethod.get
"Get a single value by index. Don’t use this operation for loops as a runtime cast is needed for every iteration."


I can try to rewrite into a vectorized private version.

R/pkg-knitr.R
```diff
@@ -0,0 +1,157 @@
...
+#' (https://github.com/pola-rs/polars/blob/4ffcb7461302a580255a9d910d70f1f7b2508675/py-polars/polars/dataframe/_html.py)
+#' for HTML outputs.
+#' @name knit_print.DataFrame
+#' @param x a polars DataFrame to knit_print
+#' @param ... additional arguments, not used
+#' @keywords DataFrame
+#' @export
+knit_print.DataFrame <- function(x, ...) {
+  .print_opt = getOption("polars.df_print", "auto")
+  if (.print_opt != "default" && (knitr::is_html_output() || (.print_opt == "html"))) {
```
vincentarelbundock, line 13: do we need to check if `knitr` is installed?

eitsupi: As I understand it, this is a function of knitr and it is never used if knitr is not installed, so no check is needed.
I believe nanoarrow does the same thing already.

https://github.com/pola-rs/r-polars/blob/71e4e58d645ca5dcc11a9916803c70f90e1a481e/R/zzz.R#L135-L139

eitsupi:
I think it works well now.
For example, if we run `rmarkdown::render("vignettes/polars.Rmd")`, we will get an HTML file that looks like this

image

It would be more useful to be able to control the table size with options, etc., but I think we can merge this for now.

sorhawell (commented):
@eitsupi 
colname and value(s) are right-aligned, whereas colname and type are left-aligned. Would it look good to right-align all three (colname, type, value(s))?

eitsupi:
> @eitsupi colname and value(s) are right-aligned, whereas colname and type are left-aligned. Would it look good to right-align all three (colname, type, value(s))?
> 
> ... or left align

Good point, that bothered me too.

But I think this looks exactly the same as the current py-polars.
For example, this is what happens when we render the following Rmd file.

````rmd
---
title: test
output:
  github_document:
    df_print: paged
---

```{r}
#| include: false
nycflights13::flights[1, ] |> arrow::write_parquet("flights.parquet")
devtools::load_all()
```

```{python}
import polars as pl

pl.read_parquet("flights.parquet")
```

```{r}
polars::scan_parquet("flights.parquet")$collect()
```
````

image
image

(Only the R's have `[` and `]` escaped with a backslash.......needs to be fixed.)

On the other hand, the following qmd file is right-aligned when rendered by Quarto CLI (1.3, prerelease version).

````qmd
---
title: test
format:
  html:
    self-contained: true
    df-print: paged
---

```{r}
#| include: false
nycflights13::flights[1, ] |> arrow::write_parquet("flights.parquet")
devtools::load_all()
```

```{python}
import polars as pl

pl.read_parquet("flights.parquet")
```

```{r}
polars::scan_parquet("flights.parquet")$collect()
```
````

image
image

So this is more like a CSS problem of `rmarkdown` package.

File: R/pkg-knitr.R [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+}
+
+#' Generate HTML table from a DataFrame
+#' @param x DataFrame
+#' @param max_cols an integer of maximum number of columns to display
+#' @param max_rows an integer of maximum number of rows to display
+#' @return A string of HTML code
+#' @examples
+#' to_html_table(mtcars, 3, 3)
+#' @export
```
vincentarelbundock, line 41: Do we want to export this function? It's great to have internally, but do we want to commit to maintaining a table-making functions that users may end up using long term?

vincentarelbundock: Also, out of curiosity, what made you decide against `knitr::kable(format="html")`? I guess the output wouldn't match pypolars exactly, but there would be no maintenance and some "guarantees" that it plays well with Rmarkdown and Quarto, since so many users depend on it and it's well tested upstream.

To be clear, I'm not advocating; just curious.

eitsupi: > Do we want to export this function? It's great to have internally, but do we want to commit to maintaining a table-making functions that users may end up using long term?

Good point. We probably don't need to export this.
(Users who want to use it can use it even if it is not public)

> what made you decide against knitr::kable(format="html")?

Short answer, simply because I like python-polars HTML tables.

I found the following document I recently created with Quarto to be a very good python polars table display.
https://eitsupi.github.io/querying-with-prql/indexing#filtering-rows

- Show types
- Show shape
- Automatically collapses the middle and displays the edges of the table

BTW, the ability to detect the `df_print` option and switch to output using `kable` can be implemented in the future.

vincentarelbundock: yeah, that document looks awesome. Let me know if/when you'd need me to install and try your branch.

eitsupi: Thanks.
You do not need you to check the output locally as we can check output from the snapshot tests.

R/pkg-knitr.R
```diff
@@ -0,0 +1,170 @@
...
+#' @param x DataFrame
+#' @param max_cols an integer of maximum number of columns to display
+#' @param max_rows an integer of maximum number of rows to display
+#' @return A string of HTML code
+#' @examples
+#' to_html_table(mtcars, 3, 3)
+#' @noRd
+#' @importFrom utils getFromNamespace
+to_html_table <- function(x, max_cols = 75, max_rows = 40) {
+  escape_html <- getFromNamespace("escape_html", "knitr")
```
vincentarelbundock, line 44: Are you 100% sure that no user will ever want to execute this function without `knitr` installed? If that's not the case, we should have a conditional as described here: https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Suggested-packages

sorhawell (commented):
looks fine to me
Worked as expected on my machine

sorhawell (approved):

R/pkg-knitr.R
```diff
@@ -0,0 +1,170 @@
...
+#' - `options(polars.df_print = "html")` for the HTML table.
+#' @name knit_print.DataFrame
+#' @param x a polars DataFrame to knit_print
+#' @param ... additional arguments, not used
+#' @keywords DataFrame
+#' @export
+knit_print.DataFrame <- function(x, ...) {
+  .print_opt <- getOption("polars.df_print", "auto")
+  .rmd_df_print <- knitr::opts_knit$get("rmarkdown.df_print")
+  if (.print_opt == "html" ||
```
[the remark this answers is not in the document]
eitsupi: Thanks, updated.

R/pkg-knitr.R
eitsupi: Makes sense. I add lines to check which knitr is installed.

vincentarelbundock (approved):
Looks good to me. This is a very nice improvement!

PR #126, opened 2023-04-18
credit Ritchie as author

sorhawell:
Some auto generated citation snippets from this polars package display me as the sole author of polars, which is not right.

I have added Ritchie as first author.

btw "cre" means package maintainer

[I skimmed this guide](https://journal.r-project.org/archive/2012-1/RJournal_2012-1_Hornik~et~al.pdf)

Any suggestions on how to depict authorship should freely be debated among the contributors. I have no strong opinions on this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -3,8 +3,9 @@ Title: Polars ported to R
 Version: 0.5.0
 Depends: R (>= 4.1.0)
 Imports: utils, codetools
-Authors@R: 
-    person("Soren", "Welling", , "sorhawell@gmail.com", role = c("aut","cre"))
+Authors@R:
+  c(person("Ritchie", "Vink", , "ritchie46@gmail.com", role = c("aut")),
+    person("Soren", "Welling", , "sorhawell@gmail.com", role = c("aut","cre")))
 Description: This package ports the polars data query library to R.
 License: MIT + file LICENSE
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
I like this guide a lot.

My sense is that practices vary *a lot* from project to project. `ctb` is a pretty "cheap" recognition because it doesn't show up in the package citation, and thus it does not really "dilute" credit of the main authors. Still, listing everyone who makes typo corrections as `ctb` can become unwieldy if/when the project becomes popular. 

That said, even very popular packages like `dplyr` tend to be "top-heavy", so finding a threshold isn't that difficult. See: https://github.com/tidyverse/dplyr/graphs/contributors You'll eventually have to draw a line somewhere, and it is certainly be an arbitrary and unsatisfactory process.

In the end, I don't think you should sweat this too much. It doesn't matter much.

eitsupi (approved):

eitsupi:
See also extendr/rextendr#236

sorhawell:
> See also [extendr/rextendr#236](https://github.com/extendr/rextendr/issues/236)

@eitsupi good point.  I will add to cran check list to resolve rust crate licensing. I think we currently depend on +100 crates, I far from know them all by name ^^ . Thank goodness for automated license tools.

PR #128, opened 2023-04-18
WIP improve roxygen usage field for methods

sorhawell:
This PR investigates how to improve .Rd-files and other docs for polars methods and especially the usage field.

### DIFF
--- a/R/dataframe__frame.R
+++ b/R/dataframe__frame.R
@@ -530,6 +530,7 @@ DataFrame_lazy = "use_extendr_wrapper"
 #' @return DataFrame
 #' @aliases DataFrame_clone
 #' @keywords  DataFrame
+#' @usage <DataFrame>$clone()
 #' @examples
 #' df1 = pl$DataFrame(iris);
 #' df2 =  df1$clone();
--- a/man/DataFrame_clone.Rd
+++ b/man/DataFrame_clone.Rd
@@ -4,7 +4,7 @@
 \alias{DataFrame_clone}
 \title{Clone a DataFrame}
 \usage{
-DataFrame_clone()
+<DataFrame>$clone()
 }
 \value{
 DataFrame


sorhawell:
@etiennebacher  did I get this right?

I remember now that if I write usage field like `<DataFrame>$clone()`, then pkgdown will try to do some fancy parsing of code and fail as it does not understand extendr-classes. Then it skips the method completely and does not include it in Reference.

I tried to modify pkgdown source code but failed to solve the issue. I personally found the parsing part in pkgdown written very convoluted, so maybe I missed the point.

```R
> pkgdown::build_site(lazy = TRUE)
Warning: Failed to parse usage:

<DataFrame>$clone()

-- Installing package into temporary library --------------------------------------------------
== Building pkgdown site =======================================================
Reading from: '/Users/sorenwelling/Documents/projs/r-polars'
Writing to:   '/Users/sorenwelling/Documents/projs/r-polars/docs'
-- Initialising site -----------------------------------------------------------
-- Building home ---------------------------------------------------------------
Reading 'NEWS.0.md'
Reading 'LICENSE.md'
Writing '404.html'
-- Building function reference -------------------------------------------------
-- Building articles -----------------------------------------------------------
-- Building news ---------------------------------------------------------------
Writing 'sitemap.xml'
-- Building search index -------------------------------------------------------
== DONE ========================================================================
-- Previewing site ----------------------------------------------------------------------------
```

etiennebacher:
> did I get this right?

yes the changes are the same as what I did

> I tried to modify pkgdown source code but failed to solve the issue. I personally found the parsing part in pkgdown written very convoluted, so maybe I missed the point.

I don't think this would be the solution because it would require maintaining a forked `pkgdown` with this change. I'm also not sure the `pkgdown` team would like to support the kind of docs that `polars` has. The code of `polars` is a bit strange in the sense that it looks like R6 but it's not R6 so I don't think they would put additional effort to support this use case.

> also CRAN check throws a warning

Hum, I don't know how to fix this one (and also the CRAN check stalls locally because it takes forever to build the package for me). If we use `mkdocs` one solution would be to replace the usage section in the markdown files rather than in the R docs or in the Rd files. This sounds a bit clumsy though.

---

I made a branch where I put some work on how to make `mkdocs` work with `polars` automatically: https://github.com/etiennebacher/r-polars/tree/mkdocs-test. 

Here's the corresponding live demo: https://etiennebacher.github.io/testRTD/site

If you want to try it locally, just get my branch and run `altdoc::preview()` (I'm not sure why but sometimes it just freezes in RStudio, but it runs fine in VS code). A few things to note:

* there's a script `docs/make-docs.R` that takes care of converting Rd files to markdown, organizing the reference section and writing `mkdocs.yml` (which is where the structure of the website is done)
* I modified a few Rd files: I added some "general classes" Rd files for `GroupBy` for example and I added some `@usage` section to see how it works 
* all the docs are self-contained in "docs"

There are still things to improve: evaluate examples in "Reference", deal with the usage sections, etc. 

I think there's no perfect solution: 

* `pkgdown` won't give us enough flexibility to deal with the organization of the reference page and the usage section
* `mkdocs` is more flexible but will require more tweaking and custom scripts. However, once these are done and rules are well established on how the docs for each function should be written, then it *should* work smoothly

Before diving too much into the work on `mkdocs`, I'd like to know what you prefer (so that we don't just waste time if we end up with `pkgdown`). Also pinging @eitsupi @vincentarelbundock @grantmcdermott to have your opinion on the docs

vincentarelbundock:
@etiennebacher I'll admit that I was very skeptical at first, because I thought there was value in having a familiar website like pkgdown.

However, now that I see your example, I think it's an absolute no-brainer. The proposed website is **much** better.

Given the non-standard nature of the package, the absolute most important part is the Reference section. In pkgdown, it is basically unreadable. In your version, it is a joy to navigate.

FWIW, I don't think @Usage is a big deal if the @Examples are rendered.

I would have some more detailed comments eventually, but I guess you're just looking for a high level view now.

Awesome stuff!

eitsupi:
@etiennebacher The side bar looks very good!
Indeed, the current pkgdown reference page is too long and difficult to read. The sidebar would definitely help.

I am a big fan of the pkgdown ecosystem cross-linking, but I felt that your demo site looks better so far since r-polars is not linked from anywhere and should be less likely to link anywhere.

As a minor point, I noticed that the font settings may not be appropriate. please compare the following sections of the get start article.

image

image

grantmcdermott:
Agree: The altdoc formatting works much better here.

Fwiw I don’t know if I’d hold out _too_ much hope for pkgdown supporting R6-esque methods in the near future. Packages like [torch](https://torch.mlverse.org/docs/reference/index.html) have had to confront this issue for a while and comprise and overlapping set of maintainers. So I kinda feel it would have happened by now. Maybe the introduction of S7 will change things, but it’s [not yet clear](https://github.com/RConsortium/OOP-WG/issues/202) that it will support encapsulated OOP anyway.

sorhawell:
@etiennebacher this is amazing :) I'm all in. This will promote r-polars so much better.

If code builds needs to be hosted and maintained by the r-polars community in parallel repositories, then we do have the organization https://github.com/rpolars to park any supporting repositories to not spam the pola-rs.

etiennebacher:
Great, now that this is decided, I think you can close this PR @sorhawell and I'll make an issue with a TODO list and I'll make a draft PR with my branch for  `mkdocs`

Issue #117, opened 2023-04-15
`.to_r_vector()` -> `.as_vector()`


vincentarelbundock:
`.as_data_frame()` vs. `to_r_vector()` seems inconsistent. Would you consider renaming this method to `.as_vector()` or creating an alias? I'm asking because this is what I expected and it took me a few minutes to figure out what was going on.

sorhawell:
You're right this is inconsistent.

the `to_xyz` style originates from [py-polars](https://pola-rs.github.io/polars/py-polars/html/search.html?q=to_) and rust. I think we should keep the `to_xyz` to be consistent with the other x-polars. However we could add in as aliases  `as_xyz` for any `to_xyz` method or use as via the S3-generic.


How about ?

`as_data_frame()` / `to_data_frame()`
`as_vector()` / `to_vector()`
`as_list` / `to_list`


btw also exists,  to_r() in some classes  . My idea with this was to lower mental strain while polars-scripting to just convert to the most meaningful R type.

vincentarelbundock:
I like this. But maybe just advertise one of the as the "standard" polars approach. The to_ syntax seems perfectly fine to me as long as it's documented and used consistently in the vignettes.

A lot of the polars design choices seem to be guided by keeping the syntax close to python and rust, and I think that's a strength. My point was just that we should pick one and be consistent.

All good.

eitsupi:
FYI, the arrow R package has `to_arrow()` and `to_duckdb()`.
https://arrow.apache.org/docs/r/reference/index.html#connections-to-other-systems

PR #132, opened 2023-04-19
`to_vector()` and `to_data_frame()`

vincentarelbundock:
Closes https://github.com/pola-rs/r-polars/issues/117

* `to_r_vector()` -> `to_vector()`
* `as_data_frame()` -> `to_data_frame()`

I added aliases to both functions with `@noRd` tags to preserve backward compatibility, and I have added ultra-minimalist tests of `as_data_frame()` and `to_r_vector()` to make sure we don't break old code.

This is a lot of changes, but it was basically just a search and replace with minimal other changes.


vincentarelbundock:
FWIW, @sorhawell suggested `$to_data_frame()` in the issue linked above, and I liked that suggestion a lot, which is why I implemented that. I personally don't think it's confusing, and would prefer that to `to_r_data_frame()`, which I find verbose and not too pretty.

Not a big fan of `to_r()` because wouldn't be certain what kind of R object this returns.

eitsupi:
> Not a big fan of `to_r()` because wouldn't be certain what kind of R object this returns.

I agree with that, but python-polars has `to_pandas` and `to_arrow`......

I looked again and found that python-polars has `polars.Series.to_frame`.
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.to_frame.html

Considering this, I still think the name `$to_data_frame` is not a good choice here.

vincentarelbundock:
> I agree with that, but python-polars has `to_pandas` and `to_arrow`......

Right. Note that these are not `to_python_pandas` and `to_python_arrow`, which would be analogous to `to_r_data_frame`.

In any case, I made my pitch and am now happy to defer this decision to the package creator(s).

vincentarelbundock:
Cool. I think I'm still unconvinced, but I agree with you that this is not such a big deal.

sorhawell:
many thx @vincentarelbundock

PR #138, opened 2023-04-20
docs: Switch website from pkgdown to mkdocs-material with altdoc

### LINKED ISSUE #137: Implement `mkdocs`
See progress here: https://etiennebacher.github.io/r-polars/site

TODO list for `mkdocs` implementation:

### Questions

- [x] Do we keep a `vignettes` folder and automatically convert vignettes to markdown for the docs or do we write vignettes/articles directly in the docs? -> write vignettes as usual in the `vignettes` folder because all the docs generation should be done automatically
- [x] Code autolinking? It is not possible to autolink `polars` functions because they are not exported by `polars` (e.g `polars::DataFrame()` errors). This is also a problem on the `pkgdown` website. It is possible to autolink functions from other packages but do we actually use other packages in docs? -> let's put this aside for now because the package doesn't have deps anyway

### Formatting

- [x] Deal with strange formatting of tables (see README or Get started) happening only in mkdocs
- [x] use same color theme as the py-polars docs? -> theming can be dealt with later, it's only a few CSS variables to change
- [x] same logo?

### To fix in R docs

- [ ] Some minor formatting issues in the "Reference" pages, e.g there's the word "data" before the page title sometimes. This requires using `@docType` (see template below)
- [ ] Clean docs in R files so that they work without problems with `mkdocs` (see template below)
- [ ] Finish categorizing docs: how do we classify `Expr_` fns, or `arr_` fns?

### Other 

- [x] Evaluate examples in "Reference" pages
- [x] See if it is possible to replace usage section in md files (and not in R files so that R CMD check doesn't complain about those) -> not specific to `mkdocs` (also happens with `pkgdown`) so lower priority
- [ ] Make a yaml for CI, maybe render the "site" folder in another branch (like `pkgdown`) -> almost done but I can't push to the branch `gh-pages` because my fork (and hence the bot) doesn't have the rights (https://github.com/pola-rs/r-polars/actions/runs/4762862052/jobs/8465590223?pr=138#step:16:32)

### Outside this package

- [x] Fix `altdoc` to allow custom scripts to be ran with `altdoc::update_docs()`
- [x] Add functionality in `altdoc` to parse NEWS and link issues, PR, contributors

---

#### General class (`DataFrame`, `LazyFrame`, etc.)

Template:

```
#' Title + description
#'
#' @params
#'
#' @return 
#'
#' @rdname CATEGORY_class <----------- IMPORTANT
#'
#' @examples

foo = function()
```

Example:
```
#' Create polars DataFrame
#' 
#' Some description if needed...
#'
#' @return data.frame
#'
#' @rdname DataFrame_class <----------- IMPORTANT
#'
#' @examples
#' df = pl$DataFrame(iris[1:3,])

DataFrame = function(...) {
  ...
}
```

#### Method (`get_columns()`, etc.)

Template for method:

```
#' Title + description
#'
#' @params
#'
#' @return 
#'
#' @rdname CATEGORY_function_name <----------- IMPORTANT
#'
#' @examples

foo = function()
```

Example:

```
#' Return polars DataFrame as R data.frame
#' 
#' Some description if needed...
#'
#' @param ... any args pased to as.data.frame()
#'
#' @return data.frame
#'
#' @rdname DataFrame_as_data_frame <----------- IMPORTANT
#'
#' @examples
#' df = pl$DataFrame(iris[1:3,])
#' df$as_data_frame()

DataFrame_as_data_frame = function(...) {
  ...
}
```

#### Other rules

If the function is just a character vector then add `@docType NULL`. Example:

```r
#' Get columns (as Series)
#' ...
#' @docType NULL <------------------ IMPORTANT

DataFrame_get_columns = "use_extendr_wrapper"
```

If the general class has no specific function, it is still required to create an Rd file for it with `NULL`:
```r
#' Use groups on a Polars lazy DataFrame
#' @name LazyGroupBy_class <----------- IMPORTANT
#'
#' @return A lazy grouped DataFrame
NULL
```

etiennebacher:
Close #137 

See progress here: ~~https://etiennebacher.github.io/r-polars/site~~ No longer available (I removed the files so that this can be merged)


eitsupi:
Just a question: do we need to commit this large number of Markdown files?

etiennebacher:
All markdown files are used when we run `mkdocs build` (called by `altdoc::preview()`) so that it can build the `docs/site` folder. I'm still thinking of the best way to write a github action that would create all the docs automatically without having to commit anything other than in the usual R development workflow (like `pkgdown`).

etiennebacher:
But for now I commit everything because I put this branch on github pages so that we can have a live demo

eitsupi:
Thanks, it make sense now.

etiennebacher:
Note to self: we still need a `docs` folder that contains `mkdocs.yaml` and a subfolder `docs` containing custom CSS and logo because that's crucial to structure our docs. What we *don't* need in this folder are:

* the files for "Reference" because they can be generated automatically
* the vignettes because can be converted automatically -> caching them is not a good idea because even if the vignette in itself doesn't change, the functions used inside can change
* the `site` subfolder because it's generated by `mkdocs build`

etiennebacher:
Looks like the CI for documentation is ready, it only errors because I don't have the rights to create a branch on this repo

eitsupi:
Could the generated files be added to gitignore and then deleted?

etiennebacher:
I think the preview is still useful for now so if I delete the files we won't have it anymore. Do we do all the cleaning in R files here before merging?

sorhawell:
> Do we do all the cleaning in R files here before merging?

I didn't quite get this question. How can I help to bring this live :) what buttons should I press to give branch permission?

Currently pkgdown pushes to rpolars github.io with `secrets.DEPLOY_DOCS` token. Could you use that from the new workflow?

.github/workflows/documentation.yaml
```diff
@@ -0,0 +1,109 @@
+# This workflow is a mix of:
+# - https://github.com/pola-rs/r-polars/blob/main/.github/workflows/check.yaml
+# - https://github.com/r-lib/actions/blob/v2-branch/examples/pkgdown.yaml
+# - https://squidfunk.github.io/mkdocs-material/publishing-your-site/?h=deploy#material-for-mkdocs
+
+name: mkdocs
+
+on:
+  push:
+    branches: main
```
eitsupi, line 10: This will always push from the main branch.
pkgdown automatically determines whether to update the root or the `dev` directory by the version number of the R package, but this trigger may be inappropriate if there is no such mechanism.

.github/workflows/documentation.yaml
```diff
@@ -0,0 +1,109 @@
...
+  pull_request:
+    branches: main
```
eitsupi, line 12: I think we need to be careful not to push websites by PR.

.github/workflows/documentation.yaml
```diff
@@ -0,0 +1,109 @@
...
+        ##### Build documentation and put it on "gh-pages" branch
+
+      - name: Build docs folder
+        run: |
+          altdoc::update_docs(custom_reference = "docs/make-docs.R")
+        shell: Rscript {0}
+
+        # https://www.mkdocs.org/user-guide/deploying-your-docs/
+      - name: Build site and deploy to GitHub pages
+        run: cd docs && mkdocs gh-deploy --force
```
eitsupi, line 109: Note that we are not currently using the gh-pages branch of this repository.

https://github.com/pola-rs/r-polars/blob/2b04d3445dde103ff41099b91a7846eb4cc7f0af/.github/workflows/pkgdown.yaml#L107-L115

.github/workflows/documentation.yaml
etiennebacher: To me this means that the CI will run when there's a push on `main`, isn't that what we want? Also for now I don't think it's possible to have versioned documentation. I think `mkdocs` allows this but I didn't look into it and maybe it should be done in another PR later

.github/workflows/documentation.yaml
etiennebacher: Same as above, we want the CI to trigger on pull requests to see if a specific PR creates an error in the documentation. However, we shouldn't push the changes indeed so we need to add `github.event_name != 'pull_request'` at the end of the CI to avoid committing every time there's a PR

.github/workflows/documentation.yaml
```diff
@@ -97,13 +100,28 @@ jobs:
...
+        with:
+          ssh-private-key: ${{ secrets.DEPLOY_DOCS }}
+
         # https://www.mkdocs.org/user-guide/deploying-your-docs/
       - name: Build site and deploy to GitHub pages
-        run: cd docs && mkdocs gh-deploy --force
+        if: github.event_name != 'pull_request'
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          clean: false
```
eitsupi, line 123: Perhaps if this is not set to `true`, the pkgdown files will continue to remain.

```suggestion
          clean: true
```

In this case, we also need to make sure that files like `.nojekyll` continue to remain.

.github/workflows/documentation.yaml
eitsupi: I don't think it is necessary to necessarily keep multiple versions of a document.

@sorhawell Do you have any idea which version of the document you would like to publish?
I think a while ago only tags would have triggered document updates.

etiennebacher:
> Currently [pkgdown pushes to rpolars github.io with](https://github.com/pola-rs/r-polars/blob/71e4e58d645ca5dcc11a9916803c70f90e1a481e/.github/workflows/pkgdown.yaml#L105) secrets.DEPLOY_DOCS token. Could you use that from the new workflow?

We now updated the workflow to also push to this repo. I removed most of the docs files (that will be automatically generated by the workflow) so once @eitsupi's comment is addressed I think we can merge this

.github/workflows/documentation.yaml
sorhawell: Before I only pushed docs for every binary release, but I think that was too rare. I think it makes sense to push docs at every push to main. At a later stage we could freeze the docs for each released version, similar to rust docs, but that is a luxury feature I think. Can a user who install polars from X commit build the docs herself locally?

etiennebacher: Yes they would only need to run `altdoc::update_docs(custom_reference = "docs/make-docs.R")` (at the condition that they have everything needed to build the package).

But the [docs versioning](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=versioni) is also a feature of `mkdocs`, we could enable it later

eitsupi: I agree that it is not really necessary to keep the documentation for the latest release since this package is not yet a CRAN release.

> they would only need to run `altdoc::update_docs(custom_reference = "docs/make-docs.R")` (at the condition that they have everything needed to build the package).

Perhaps it would be worthwhile to manage jobs on this using Makefile, etc. I have created an issue #173

eitsupi:
Thank you for tackling this tough task!
I would like to try a local build later and push some minor updates.

.github/workflows/documentation.yaml
etiennebacher: I never used Makefiles so I don't really understand what it would bring here. Is it to run `devtools::document` + `devtools::check` + `altdoc::update_docs(custom_reference = "docs/make-docs.R")` all at once? If so, how do you run it locally?

vincentarelbundock: You just type `make` in a bash/zsh terminal. Once you are ready, just post different code blocks with the different workflows that we need (series of commands) in the Makefile Issue, and one of us (I can) will write one.

eitsupi:
I have made some minor updates.

- Rbuildignore and gitignore some files
  - Since `mkdocs.yml` is both the template and an auto generated file, CI was set up to copy the template `mkdocs.orig.yml` to `mkdocs.yml` just before executing `altdoc::update_docs()`.
    Perhaps it would be worth adding a new feature that could specify a template for `altdoc::update_docs()`
- To check the results of the CI run, I overwrote the workflow definition file with the existing pkgdown one.

If there are no problems with this change, it is working well and can be merged. Great job!

image

docs/mkdocs.yml
```diff
@@ -0,0 +1,520 @@
...
+repo_url: https://github.com/pola-rs/r-polars
+repo_name: polars
+plugins:
+  - search
+extra_css:
+  - stylesheets/extra.css
+nav:
+  - Home: README.md
+  - Get started: articles/polars.md
+  - Reference:
```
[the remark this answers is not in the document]
eitsupi: Thanks, as I mentioned in my comment (https://github.com/pola-rs/r-polars/pull/138#issuecomment-1520184506), I think it is better to separate the templates from the generated files, given that they may be updated automatically at build time.

etiennebacher:
Looks great, thanks for cleaning all of this @eitsupi, it's much better to keep a clear template like this. One thing: should we rename `pkgdown.yml` to something else?

Otherwise, if tests pass, I'm ok to merge

eitsupi:
> One thing: should we rename `pkgdown.yml` to something else?

My understanding is that CI will not run if renamed in this PR.
It would make sense to change this in a follow-up PR after merging this.

eitsupi (approved):

sorhawell:
Looking forward to see the new docs. pkgdown was a good stepping stone, but just deprecate it whenever ready :)

etiennebacher:
Looks like it worked! :tada: :tada: 

I'll make another PR soon with various tweaks

vincentarelbundock:
This is *amazing* work and a massive improvement over `pkgdown`. Thanks!

eitsupi:
Awesome!

(I forgot to remove the dev doc badge from README...)

PR #140, opened 2023-04-20
`[`.DataFrame

vincentarelbundock:
This PR implements the `[.DataFrame` accessor with associated tests and informative errors.

I think that all we would need to make this work for `LazyFrame` is a `LazyFrame$columns` attribute, so I left in some code that can be uncommented to make that work when `LazyFrame.columns` becomes available. 

I also added an example to the Get Started vignette.

``` r
library(polars)

df = pl$DataFrame(mtcars)

df[, "mpg"]
# shape: (32, 1)
# ┌──────┐
# │ mpg  │
# │ ---  │
# │ f64  │
# ╞══════╡
# │ 21.0 │
# │ 21.0 │
# │ 22.8 │
# │ 21.4 │
# │ ...  │
# │ 15.8 │
# │ 19.7 │
# │ 15.0 │
# │ 21.4 │
# └──────┘

df[1:4, c("hp", "wt")]
# shape: (4, 2)
# ┌───────┬───────┐
# │ hp    ┆ wt    │
# │ ---   ┆ ---   │
# │ f64   ┆ f64   │
# ╞═══════╪═══════╡
# │ 110.0 ┆ 2.62  │
# │ 110.0 ┆ 2.875 │
# │ 93.0  ┆ 2.32  │
# │ 110.0 ┆ 3.215 │
# └───────┴───────┘

i = c(TRUE, TRUE, rep(FALSE, nrow(df) - 2))
j = c(TRUE, TRUE, rep(FALSE, ncol(df) - 2))
df[i, j]
# shape: (2, 2)
# ┌──────┬─────┐
# │ mpg  ┆ cyl │
# │ ---  ┆ --- │
# │ f64  ┆ f64 │
# ╞══════╪═════╡
# │ 21.0 ┆ 6.0 │
# │ 21.0 ┆ 6.0 │
# └──────┴─────┘
```


eitsupi:
Thanks!
I need this in https://github.com/pola-rs/r-polars/pull/125#issuecomment-1516259235

vincentarelbundock:
The CI error appears due to Github Actions, and seems unrelated to the actual content of this PR:

```sh
  Warning message:
  ! Wrong OS or architecture, pak is probably dysfunctional.
    Call `pak_update()` to fix this. 
  Error in dyn.load(dll_file) : 
    unable to load shared object '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/RtmpieUNnz/file10f614337c71/glue/libs//glue.so':
    dlopen(/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/RtmpieUNnz/file10f614337c71/glue/libs//glue.so, 0x0006): tried: '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/RtmpieUNnz/file10f614337c71/glue/libs//glue.so' (not a mach-o file), '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/RtmpieUNnz/file10f614337c71/glue/libs/glue.so' (not a mach-o file)
  Calls: <Anonymous> ... load_private_packages -> load_private_package -> dyn.load
  Execution halted
  Error: Process completed with exit code 1.
```

sorhawell:
Every 15th run or so has some fluke in a dependency installation, anything before check. I pressed the 'rerun failed job' button in the action page.

grantmcdermott:
Forgive the minor hijacking of this PR, but two typos that might as well get fixed in the process of merging Vincent’s nice new additions.

1. In the vignette, the Python method should also be `groupby` (no underscore): https://github.com/pola-rs/r-polars/blob/4806147b638f06145276d5384768ca5fd178186d/vignettes/polars.Rmd#L87
2. We use “=“ for assignment everywhere else (and I think this is a good thing given it makes it easy to visually compare with Python), so I think the lone “<-“ in the README here should be changed. https://github.com/pola-rs/r-polars/blob/4806147b638f06145276d5384768ca5fd178186d/README.Rmd#L117

sorhawell:
@vincentarelbundock  a rabbit hole for another time :)

[not long time ago I made a POC repo on dressing up any rust type as an R vector](https://github.com/sorhawell/extendr_usize_vector/blob/main/R/VecUsize.R). Here I implement `usize_vector` + just enough S3 generics to make it compatible with base `data.frame`. We could one day also implement similar S3 methods for polars Series to make it behave like any other R vector. 

Polars/ r-arrow Series are ~4 times faster for most arithmetic operations. If implemented natively multi-threaded it could even become faster.

``` r
library(helloextendr)

#make
x = usize_vec(c(1:3 ,NA_integer_))

#index it and print
print(x[1:3])
#> [1] "usize vector [ 1, 2, 3  ]"

length(x)
#> [1] 4

#put in a data.frame
df = data.frame(
  usize_vec = x,
  i32_vec = seq_along(x) #seq_along just needs length to work
)

#subset combined vector
df[2:4,]
#>   usize_vec i32_vec
#> 2         2       2
#> 3         3       3
#> 4        NA       4
```

<sup>Created on 2023-04-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

vincentarelbundock:
That's interesting. We could squeeze them into `data.frame()`, but I'm not sure exactly which operations will be speeded-up. I think +-/* are S4 in R, and I'm not sure we can define new methods for `PolarsSeries` R6.

eitsupi:
#140

PR #142, opened 2023-04-20
Expand `path` in `pl$lazy_csv_reader`

vincentarelbundock:
1-line commit to ensure that `pl$lazy_csv_reader` and friends expand paths.

``` r
library(polars)
write.csv(airquality, "~/Downloads/air.csv")

pl$csv_reader("~/Downloads/air.csv")
# Error: in rlazy_csv_reader: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" }) 
#  when calling :
```

### DIFF
--- a/R/csv.R
+++ b/R/csv.R
@@ -83,6 +83,8 @@ lazy_csv_reader = function(
 
   #capture all args and modify some to match lower level function
   args = as.list(environment())
+  
+  args[["path"]] = path.expand(path)
 
   #overwrite_dtype: convert named list of DataType's to DataTypeVector obj
   if(!is.null(args$overwrite_dtype)) {


sorhawell:
awesome!  maybe add a single little test with a tilde path

sorhawell (changes requested):
awesome maybe add a single test which uses a tilde path

vincentarelbundock:
> awesome maybe add a single test which uses a tilde path

I thought about it, but I'm actually not so sure how to do that. I've had some pretty bad experience trying to include write-to-file tests in CRAN packages.

And I think in general it is not "polite" for a test suite to write in a user's home directory without warning and explicit prompting.

sorhawell (approved):

sorhawell:
good point :)

One day someone could write an integration test I guess.

PR #143, opened 2023-04-20
feat: add an internal method `Series$get_fmt`

eitsupi:
Needed for #125


eitsupi:
Am I possibly mistaken about exporting methods?
I set `@noRd`, but can the user still easily access this method?

sorhawell:
`@noRd` is also fine by me. I have so far tagged internal methods/function with `#' @keywords internals.` Then the user can look up the method if they are curious in the [internals section](https://rpolars.github.io/reference/index.html#internals) https://rpolars.github.io/reference/index.html#internals.

R/series__series.R
```diff
@@ -251,6 +251,15 @@ Series_rem = function(other) {
   .pr$Series$rem(self, wrap_s(other))
 }
 
+#' @param index index of the Series (0-base)
+#' @param str_length length of string to show
+#' @return a string
+#' @examples
+#' pl$Series(c("foo", "bar"))$get_fmt(index = 1, str_length = 3)
+#' @noRd
+Series_get_fmt = function(index, str_length) {
```
sorhawell, line 260: if we define this function `get_fmt()` it will be public. If we want it to stay private we can just, skip this definition, and access from anywhere else internally with `.pr$Series$get_fmt`. Currently `.pr$...` is exported such that users could access the private methods, but we could remove this export such that a user must write `polars:::.pr$...` to use them.

sorhawell:
@eitsupi
It might be little slower then other polars functions, but probably fast enough for limited printing.
https://pola-rs.github.io/polars/polars/series/trait.SeriesTrait.html#tymethod.get
"Get a single value by index. Don’t use this operation for loops as a runtime cast is needed for every iteration."

I can try to rewrite into a vectorized private version.

R/series__series.R
sorhawell: we can also just remove the  `Series_` prefix, then it will not be exported as a public method

sorhawell:
@eitsupi 

I take back what I said about vectorized formatter. rust `get_fmt()` methods looks good, I could not come with anything faster. Probably way faster than R terminal printing.

I just deleted the public method and updated the test, as I suggest you could use the .pr$Series$get_fmt() in knitr functions or any other function.

If you need a derived internal R function, you can do as [here and just remove the class prefix.](https://github.com/pola-rs/r-polars/blob/d5120b77c75bc356335e9eb4f4657e06c423a665/R/expr__expr.R#L523)

I think here, it is OK in rust, that get_fmt() can panic as it is for internal use and it should be fairly light weight.

sorhawell (approved):

eitsupi:
Thanks for updates!

PR #147, opened 2023-04-20
feat: dimnames for DataFrame

eitsupi:


R/s3_methods.R
```diff
@@ -117,6 +117,14 @@ ncol.DataFrame = function(x) x$height
 #' @noRd
 names.DataFrame = function(x) x$columns
 
+#' @export
+#' @noRd
+row.names.DataFrame = function(x) as.character(seq_len(nrow(x)))
```
sorhawell, line 122: fine to have row.names and dimnames. polars does not store row.names. Currently when converting DataFrame to data.frame, the row.names are just set as sequence starting from 1:...
```R
> rdf = data.frame(a=1:3, row.names = c("a1","a2","a3"))
> row.names(pl$DataFrame(rdf)$to_data_frame())
[1] "1" "2" "3" #not a1 a2 a3
```

sorhawell (approved):

R/s3_methods.R
eitsupi: Yes.

Sorry I forgot to mention, but this is the behavior copied from `arrow::Table`.

eitsupi:
Thanks for your review!

Issue #146, opened 2023-04-20
Some methods execute `LazyFrame$collect()`


eitsupi:
https://github.com/pola-rs/r-polars/blob/4806147b638f06145276d5384768ca5fd178186d/R/s3_methods.R#L23
I think this is very dangerous because collect might perform a large number of calculations.

See `arrow::Dataset`. `arrow::Dataset` does not have methods which needs collect.
https://github.com/apache/arrow/blob/0bf777a5952be012e41f5b1ad443d4fec38e6f5a/r/R/dataset.R

...Or, `tbl_lazy` returns NA for the number of rows that cannot be determined until collect is executed.
https://github.com/tidyverse/dbplyr/blob/505ca7004ce339f14cfc54aa714c5492d5bc2042/R/tbl-lazy.R#L41-L43

vincentarelbundock:
That's a very good point. I can re-write those methods to tell users that they must ˋcollect()` before calling ˋdim()` and friends.

sorhawell:
I agree. I like in polars that only collect triggers collect. In spark it is less obvious I think.

vincentarelbundock:
Can be closed.

PR #148, opened 2023-04-20
Issue Some methods execute `LazyFrame$collect()` #146

vincentarelbundock:
#146

The only S3 methods which now call collect are `as.data.frame()` and `as.matrix()`. In those cases, the user makes an explicit choice to bring everything in memory as R objects, so it seems fine to me.

In all other cases, an informative error is raised.


vincentarelbundock:
I assume you get the notifications, but tagging @eitsupi just in case.

sorhawell:
looks good to me

Issue #153, opened 2023-04-21
"linker `x86_64-w64-mingw32.static.posix-gcc.exe` not found" error on Windows build


eitsupi:
It seems to be occurring after #121 was merged, but #121 seems to be unrelated to this error, so I do not know what is causing it.

```log
── R CMD build ─────────────────────────────────────────────────────────────────
* checking for file 'D:\a\r-polars\r-polars/DESCRIPTION' ... OK
* preparing 'polars':
* checking DESCRIPTION meta-information ... OK
* cleaning src
* installing the package to build vignettes
      -----------------------------------
* installing *source* package 'polars' ...
** using staged installation
** libs
rm -Rf polars.dll ./rust/target/x86_64-pc-windows-gnu/release/libpolars.a entrypoint.o
gcc  -I"C:/R/include" -DNDEBUG     -I"c:/rtools42/x86_64-w64-mingw32.static.posix/include"     -O2 -Wall  -std=gnu99 -mfpmath=sse -msse2 -mstackrealign  -Wall -pedantic -fdiagnostics-color=always -c entrypoint.c -o entrypoint.o
mkdir -p ./rust/target/libgcc_mock && touch ./rust/target/libgcc_mock/libgcc_eh.a
# CARGO_LINKER is provided in Makevars.ucrt for R >= 4.2
#--target-dir ./rust/target
echo "RPOLARS_RUST_SOURCE is set to /d/a/r-polars/r-polars/src/rust"
RPOLARS_RUST_SOURCE is set to /d/a/r-polars/r-polars/src/rust
export CARGO_TARGET_X86_64_PC_WINDOWS_GNU_LINKER="x86_64-w64-mingw32.static.posix-gcc.exe" && \
	export LIBRARY_PATH="${LIBRARY_PATH};/c/Users/runneradmin/AppData/Local/Temp/RtmpaSK6DM/Rbuildb78576a20b3/polars/src/./rust/target/libgcc_mock" && \
	cargo +nightly-gnu build --target=x86_64-pc-windows-gnu --lib --release --manifest-path="/d/a/r-polars/r-polars/src/rust/Cargo.toml"
   Compiling proc-macro2 v1.0.50
   Compiling unicode-ident v1.0.6
   Compiling quote v1.0.23
error: linker `x86_64-w64-mingw32.static.posix-gcc.exe` not found
  |
  = note: program not found
error: could not compile `proc-macro2` (build script) due to previous error
[...]

PR #154, opened 2023-04-21
hotfix windows 4.3 rtools

sorhawell:
hotfixing  #153


sorhawell:
before disabling windows check ,it might be enought to just point to rtools 43 when using R 4.3 ^^

PR #167, opened 2023-04-22
Update renv + change roxygen2 and nanoarrow

sorhawell:
update renv version
add styler.equals (install from my fork which also has an addin for Rstudio)

set roxygen2  not bleeding edge (9000) for reproducibility  (willing to undo this, if anyone is against)
set nanoarrow not bleeding edge (9000) I could compile bleeding on my mac for some reason

rextendr as the only currently stays 9000

renv is not mandatory
it is suggestion of package versions for development environment

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,7 +10,7 @@ Description: This package ports the polars data query library to R.
 License: MIT + file LICENSE
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3.9000
+RoxygenNote: 7.2.3
 URL: https://github.com/pola-rs/r-polars
     https://rpolars.r-universe.dev/polars
 Suggests: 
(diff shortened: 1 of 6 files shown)


eitsupi:
Related to this, the style of describing dependent packages for development in the DESCRIPTION file seems to be gradually spreading.

For example, my prqlr DESCRIPTION file says like this.

```dcf
Suggests:
    knitr (>= 1.38),
    rmarkdown,
    DBI,
    glue,
    RSQLite,
    tidyquery,
    nycflights13,
    dplyr,
    testthat (>= 3.0.0),
    patrick,
    withr
Config/Needs/dev:
    devtools,
    rextendr,
    lintr,
    styler,
    processx,
    jsonlite,
    purrr,
    RcppTOML
Config/Needs/website:
    pkgdown,
    rextendr
```

https://github.com/eitsupi/prqlr/blob/b6b822765dcddb47a92de0f047a0b6e071a7fc5d/DESCRIPTION

These all packages can be installed with something like `pak::local_install_deps(dependencies = c("all", "Config/Needs/dev", "Config/Needs/website"))`.

`renv::install` also seems to have recently added support for these (rstudio/renv#1138). (However, I still use pak because it works better as far as I have tried)

sorhawell:
@eitsupi 

Thx. I will try not using renv for some time and only pak to get better feel of it.

PR #170, opened 2023-04-23
Series$mean median std var

vincentarelbundock:
New minimalist implementations of a few Series methods.


R/extendr-wrappers.R
```diff
@@ -1,7 +1,4 @@
 # Generated by extendr: Do not edit by hand
-
-# nolint start
-
```
eitsupi, line 4: Could you use the dev version rextendr?
(Installs from GitHub or R-universe)

vincentarelbundock: Sure, no problem: 007e675a755cc56d1d57ea70518ca1063818d43d

R/series__series.R
```diff
@@ -679,6 +705,26 @@ Series_max = "use_extendr_wrapper"
...
 Series_min = "use_extendr_wrapper"
 
+#' @title Var
+#' @description Aggregate the columns of this Series to their variance values.
+#' @keywords Series
+#' @param ddof integer Delta Degrees of Freedom: the divisor used in the calculation is N - ddof, where N represents the number of elements. By default ddof is 1.
+#' @return A new `Series` object with applied aggregation.
+#' @examples pl$Series(1:10)$var()
+Series_var = function(ddof = 1) {
+  unwrap(.pr$Series$var(self, ddof))
```
sorhawell, line 715: we can add a context in unwrap like this
`unwrap(.pr$Series$var(self, ddof), "in $var()")`

my thinking behind adding context to errors is that it can otherwise be hard to locate the source of an error inside a long chain of several methods. If new users get confused, they might give up on polars prematurely. I have not consistently added context to all methods, but I think it is helpful.

vincentarelbundock: e7721fb018caaa05968d25bbe90166c2217437c2 adds context to these and several others where it was originally missing.

src/rust/src/series.rs
```diff
@@ -465,6 +474,14 @@ impl Series {
         self.0.sum_as_series().into()
     }
 
+    pub fn std(&self, ddof: u8) ->  Series {
```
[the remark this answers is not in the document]
vincentarelbundock: Got it. e7721fb018caaa05968d25bbe90166c2217437c2 also wraps the output in `OK()`:

```rust
pub fn var(&self, ddof: Robj) ->  Result<Series, String> {
  Ok(self.0.var_as_series(robj_to!(u8, ddof)?).into()) 
}
```

sorhawell (approved):

PR #172, opened 2023-04-24
LazyFrame_join_asof + DataFrame_join_asof

### LINKED ISSUE #151: support join_asof
I'm trying to benchmark rpolars (r package) vs. other implementations of 'asof' time series joins using the R package as opposed to through reticulate - as presented by @bwlewis

https://bwlewis.github.io/duckdb_and_r/asof/asof.html

However I do not see the join_asof function exposed for DataFrame. Is this planned in the future?  Thanks!

sorhawell:
resolve #151


sorhawell:
currently misses some more tests + DataFrame_join_asof

inst/misc/examples/example.R
```diff
@@ -48,7 +48,7 @@ pf$select(
   pl$col("Sepal.Width")$sum()$over("Species")$alias("sw_sum_over_species"),
   pl$col("Sepal.Length")$sum()$over("Species")$alias("sl_sum_over_species"),
   "Petal.Width"
-)$to_data_frame() %>% head
+)$to_data_frame() |> head
```
etiennebacher, line 51: you need to add `()` for functions with native pipe

eitsupi:
Could you resolve the conflicts?

R/lazyframe__lazy.R
```diff
@@ -625,4 +625,101 @@ LazyFrame_sort = function(
...
+#'     'on' key is less than or equal to the left's key.
+#'
+#'   - A "forward" search selects the first row in the right DataFrame whose
+#'     'on' key is greater than or equal to the left's key.
+#'
+#' The default is "backward".
+#' @keywords LazyFrame
+#' @return joined LazyFrame
+#' @examples #
+LazyFrame_join_asof = function(
```
etiennebacher, line 691: This function looks super useful, but is also a bit hard to grasp in my opinion (I had to spend a few minutes on the simple examples in the tests to understand the results). I think it's important to add several examples here. There could even be a small vignette for this function to explore all of its parameters (we could also compare it with the inequality joins of `dplyr` or `data.table` to make the transition a bit easier).

sorhawell: Thx. I will add some more examples to docs immediately. Also I forgot to add in the eager DataFrame version.

sorhawell: I have added a handful of examples showcasing args: on, by and  tolerance. I suggest to skip vignette for now as @rdavis120 in issue #151 wanted join_asof to exactly compare with other libraries.

sorhawell:
@etiennebacher will you be my reviewer again? :)

etiennebacher:
Sure, but I don't have time for a proper review now, I can probably do it in the next day or two.

Just one thing I noticed is that the docs for `join_asof` is the same for `DataFrame` and `LazyFrame`. You could write them only for one of the two and use the roxygen tag `@inheritParams` in the other one to automatically duplicate the docs. That way you also ensure that future changes will be synchronized

sorhawell:
@etiennebacher I added inherits Param. I will start using that a lot more. Thx

R/lazyframe__lazy.R
```diff
@@ -625,4 +625,132 @@ LazyFrame_sort = function(
 }
 
 
+#' join_asof
```
etiennebacher, line 628: same as above

etiennebacher (changes requested):
Looks good to me for the R part, I can't tell for the Rust part but I guess "if it compiles it should be fine" <sup>TM</sup> 😄

etiennebacher:
The docs workflow fails because I pushed to the wrong branch earlier, I tried to remove the commit (it was [b42c05e](https://github.com/pola-rs/r-polars/pull/172/commits/b42c05ef6bd34a2ef1e592b5b4b327482d74f80b)) but apparently it didn't work. Maybe you can just go back to the old code in `docs.yml`

etiennebacher (approved):
I'm fine with this, looks like a very useful function, thanks

etiennebacher:
One last thing, could you add a NEWS bullet point?

PR #177, opened 2023-04-26
dev: add Makefile to run some tasks

### LINKED ISSUE #173: Set up Makefile or something to run tasks
Other repositories in this organization seem to use Makefiles to define tasks.

By the way, I prefer Task rather than Make (like https://github.com/eitsupi/prqlr/blob/b6b822765dcddb47a92de0f047a0b6e071a7fc5d/Taskfile.yml), which I think is easier to use than Make in terms of built-in documentation and Windows support with built-in sh.
https://taskfile.dev/

eitsupi:
Close #173


vincentarelbundock:
haha, I was working on this just now. You beat me by a few minutes.  :)

Makefile
```diff
@@ -0,0 +1,60 @@
...
+	$(VENV_BIN)/pip install --upgrade mkdocs-material
+
+.PHONY: requirements-rs
+requirements-rs:
+	rustup toolchain install nightly
+	rustup default nightly
+
+.PHONY: build
+build: ## Compile polars R package and generate Rd files
+	Rscript -e 'rextendr::document()'
```
vincentarelbundock, line 40: I would expect this target or `build-all` to call `R CMD build .`

eitsupi: That makes sense.

I expect the target we use most routinely to update README.md, and `build-all` was intended to be the target for that.
What name would you suggest? Simply `all`?

vincentarelbundock:
install target may also be useful:

```Makefile
install:
	Rscript -e 'devtools::install(pkg = ".", dependencies = TRUE)'
```

Makefile
vincentarelbundock: Either that or `build-readme`, `build-pkg`, `build-doc` and then `build` or `build-all` does them all. `all` feels a bit too generic to me, but I don't have a strong view on this. Whatever you pick will be fine.

eitsupi: The point here is that README can only be updated after the R package has been built.

If we expect that `build` to update README, we need to update README in `build` itself, not in the dependencies of `build`.

Perhaps we need to define a new target with dependencies on `build`, `test`, and `README.md`.

vincentarelbundock: Ah yes, I see. As I said, this is mostly dev-facing, so I don't think we need to worry too much about user interface or naming here. Whatever you think is best!

vincentarelbundock:
> I am wondering about the need to add this as I believe this is a very common function and is executed infrequently during development.

Not a big deal, but I personally very often call `make install` in my other packages because I always want to dogfood the latest version in day to day work.

vincentarelbundock (approved):
This is really great! Looks good to me for a merge.

sorhawell:
It think windows devs need to addgnu target`rustup target add x86_64-pc-windows-gnu`

sorhawell:
this is fantastic :)

sorhawell (approved):

eitsupi:
> not entirely sure but I think requirement-rs may need to have some OS specific branching of install. On windows msvc is default.

Good point, we definitely need to specify GNU nightly on Windows!

eitsupi:
I don't know much about Windows so I'll merge this for now. (I don't know if Windows users will want to use this, since this Makefile only works on Bash.)

If anyone is using Windows, updates are welcome.

vincentarelbundock:
Thanks for your work on this. Should be very useful!

Issue #149, opened 2023-04-20
Open discussion on renaming polars classes


sorhawell:
I suggest to prefix all polars classes `Polars`, examples would be `PolarsDataFrame` `PolarsSeries` `PolarsDataType`

Pros:
 - Reduces risk of namespace collision with other R packages
 - Improves readability of rust code as currently native rust-polars struct and wrapper structs are named the same:
 
 readability example of current naming: rust-polars `LazyFrame` in a wrapper struct named `LazyFrame`. This can be a confusing to read, also the compiler error can be confusing.
<img width="396" alt="image" src="https://user-images.githubusercontent.com/8015271/233480120-48175a53-81cb-451d-9151-14f4d6839a9e.png">

Cons:
 - class/struct names become a bit long
 
 What naming do you (anyone) prefer/suggest? Should we rename? Prefixed RPolars?
 
 
Bonus info
 - Camel case is the only (unless explicitly disabled) allowed naming style for structs on rust side.
 - Currently, extendr can only name R side auto generated class the same as rust side wrapper structs.
 - Since r-polars only uses 2 structs + 1 class from rust-polars to public R, whereas py-polars use 2 structs + 2 classes, the final  public R side class name most be the same as the rust side wrapper struct. (**footnote)

(**the long footnote)
py-polars has 4 classes/structs from rust to python. DataFrame(rust-polars struct) -> PyDataFrame(wrapper struct) -> PyDataFrame (py03 internal class) -> DataFrame(py-polars public class). The vast majority of py-polars wrapper structs and wrapper classes are the thinnest possible.
r-polars only has 3 classes/struct from rust to python:  DataFrame(rust-polars struct) -> DataFrame(wrapper struct) -> DataFrame(extendr class). The R side DataFrame class though have both a set of public methods and a private method set. An early version of r-polars had 4 classes like py-polars and was using R6. That led to more conversions needed at runtime, more boiler plate code, and class instances was no longer just an R type ExternalPtr. Also, I felt it was sometimes cumbersome to make R6 behave exactly like py-polars syntax. Instead I decided to just implement extra base R S3 generic methods for the already auto-generated extendr-classes and add public method functions.

vincentarelbundock:
FWIW, I agree that this is a useful change: I can imagine `DataFrame` vs. `data.frame` being confusing for some users.

> I suggest to prefix all polars classes `Polars`, examples would be `PolarsDataFrame` `PolarsSeries` `PolarsDataType`

[...]

PR #178, opened 2023-04-26
chore: rename the crate and lib to r-polars

eitsupi:
Related to #149

Rename the internal Rust crate from `polars` to `r-polars` for clarity.


sorhawell (approved):
ahh I was fiddling with this, didn't know it needed to be an `_` and not a `-` in makevars files.

eitsupi:
> ahh I was fiddling with this, didn't know it needed to be an _ and not a - in makevars files.

I didn't know that too, but rextendr know that. (`rextendr::use_extendr` has options to rename Rust side)

PR #181, opened 2023-04-27
Series aggregation function return R scalar value

sorhawell:


sorhawell:
@vincentarelbundock   I agree I would have preferred methods stick to returning same class if not very obvious otherwise. I guess py-polars convert and return py-value i the length of the output can only be one.

I think r-polars can deviate from py-polars if makes a lot sense. Some py-polars Series methods have mutable behaviour. I have disabled that as Default, e.g. Series_rename. 

This PR intends to match py-polars  Series methods.


tests not updated yet.

sorhawell:
@eitsupi or @vincentarelbundock  can you see why this fails for ubuntu? Maybe I will try to redo the PR incrementally to locate my error

eitsupi:
I don't know why, but I rerun the jobs and they are all green now.

vincentarelbundock (approved):

PR #185, opened 2023-04-28
Fix doc ordering and usage section

etiennebacher:
Mentioned in #182:

* fix ordering (previously the description came after the "Returns" section)
* fix usage section, e.g `DataFrame_shift_and_fill(fill_value, periods = 1)()` -> `<DataFrame>$shift_and_fill(fill_value, periods = 1)`. 

*Note:* I had to hardcode the usage of `pl$DataFrame()` in `docs/make-docs.R` because it doesn't appear in the Rd file because of the `$`.

Small preview:

image

image

Issue #182, opened 2023-04-27
Docs improvements


etiennebacher:
Sorry for yet another (but hopefully final) todo list for docs. I prefer this rather than listing issues / suggestions in PR comments. @sorhawell @grantmcdermott @vincentarelbundock @eitsupi please edit this post or add comments if you find some issues while browsing the docs or have some suggestions on how to improve them.

- [x] Make a reference landing page 
- [x] Fix section order in reference pages (e.g the description comes after the "Returns" section), reported in https://github.com/Genentech/rd2markdown/issues/30
- [x] Split `Expr` docs into several subsections as in [py-polars](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/aggregation.html)
- [x] Footnote hyperlinks are not displaying properly. (Search for "[^1]" in the [vignette](https://rpolars.github.io/articles/polars/).)
- [ ] Table formatting works in some setups but not all (e.g mobile + laptop on Ubuntu)
- [x] suffix all roxygen2 r-polars classes `#' @name` with `_class` to let docs auto discover the classes
- [x] make a tab "About" that contains license + citation + developer info
- [x] Replace `DataFrame_clone()` by `<DataFrame>$clone()` in "Usage" section (and same for all other functions)
- [ ] Use [`@inherit`](https://roxygen2.r-lib.org/reference/tags-reuse.html) and friends to automatically use the same docs for LazyFrame and DataFrame
- [x] Add `@doctype NULL` and `@format NULL` everywhere the function is just a character `"use_extendr_wrapper"`. This is to avoid the useless "Format" section (like [here](https://rpolars.github.io/reference/Expr_product/)) that is usually for data.

vincentarelbundock:
Thanks for all your work on this Etienne! FYI, when PR #175 is merged, I intend to take a close look at all the documentation. I'll try to address some of the issues listed here.

grantmcdermott:
Added a minor point about footnotes. But in general just want to say that the docs are looking fantastic. Thanks Etienne!

sorhawell:
added _class tagging of all roxygen2 docs of polars classes. I'll happily do that :)
It looks like I have 1-3 weeks nearly full time polars starting next week.

vincentarelbundock:
> added _class tagging of all roxygen2 docs of polars classes. I'll happily do that :) It looks like I have 1-3 weeks nearly full time polars starting next week.

[...]

PR #186, opened 2023-04-28
Documentation tweaks

vincentarelbundock:
Preview: http://arelbundock.com/polars

This PR is a refactor of the `make_docs.R` script. Roughly speaking, the current script works like this:

1. Read `.Rd` files
2. Convert `.Rd` files to `.md` files using the `rd2markdown` package.
3. Use `mkdocs` to convert those `.md` files back to `.html` files and build the website.

The multiple conversions are a bit convoluted and risky, as some things may be lost during translation. This workflow also relies on the `rd2markdown` package, which is only available and apparently has bugs: not all documentation section are displayed, and the order is shuffled.

This PR changes the workflow to:

1. Read `.Rd` files.
2. Convert `.Rd` files to `.html` using base R `tools::Rd2HTML`.
3. Embed this html code directly in the `.md` files that `makedocs` reads.

With a few tweaks, this PR gives us:

- [x] Code simplification: `make_docs.R` is 40% shorter.
- [x] No more dependencies on `cli`, `tinkr`, `magrittr`, `stringr`, `fs`, and the problematic `rd2markdown`.
- [x] All sections of the `.Rd` files are displayed in the correct order.
- [x] Nice display of the Arguments section in a table.
- [x] Polars User Guide page.
- [x] About page.
- [x] Simple (naive?) way to control order of Reference menu.
- [x] Nested menu for `Expr`
- [x] Functions are sorted in sections using function names or `@name` tags.

TODO:

- [x] Usage as `<DataFrame>$columns`
- [x] Argument names are sometimes wrapped when they should not be.
- [ ] Code highlighting of the Usage section.
- [x] Go through the reference menu to identify problematic cases.
- [x] Large space in Examples code block
- [x] Missing usage in `pl_DateTime` and others
- [ ] Landing page
- [x] the header for each reference should be the title of the Rd file and not the function name. For example, here (https://arelbundock.com/polars/reference/DataFrame_columns/) we should remove the "columns" at the top of the page (like how it currently is in the deployed docs)
  
I would appreciate any and all comments


vincentarelbundock:
tag @etiennebacher who might be interested.

etiennebacher:
Impressive work, this refactoring looks really nice. I'll dive in the code a bit later but a few visual remarks so far:

* the header for each reference should be the title of the Rd file and not the function name. For example, here (https://arelbundock.com/polars/reference/DataFrame_columns/) we should remove the "columns" at the top of the page (like how it currently is)
* there's a large space at the beginning of the code block for examples. This is due to the output of `downlit` that needs to be tweaked a bit (I did that in a commit in #138 but I don't remember which one exactly)
* would be nice to replace e.g `DataFrame_columns()` by `<DataFrame>$columns()` in "Usage". You can take some code from #185 for this
* the "Usage" section is not displayed for some pages (e.g https://arelbundock.com/polars/reference/pl_Datetime/)

I love the subsections for `Expr`. Do you want to add a landing page in this PR as well or is it for later?

vincentarelbundock:
@etiennebacher I updated the TODO list, checked off a few things, and updated the preview website.

I don't know how to change the header, but I think I actually like having the function up top.

I looked at a couple cases of missing Usage sections, and it looks like those don't have a Usage in the .Rd file itself. I guess we could add them manually in roxygen2, but I don't think the workflow is dropping them.

etiennebacher:
Additional remarks:

1. The description text before the usage section is too big, it's the same size as section titles. 
2. For highlighting the usage section, it would have to be done before fixing the function names because `downlit` doesn't work with the `<>`:
``` r
downlit::highlight("<DataFrame>$clone()")
#> [1] NA
```
3. Regarding sections in the sidebar, I think they should be ordered alphabetically and that we should rename the `Polars` section ("Operations" maybe?).
4. I really prefer having the function title as header instead of the function name (using title as header is also what `pkgdown` does).

Finally, is it fine to push some commits in your branch or do you prefer handling this yourself?

vincentarelbundock:
@etiennebacher I just pushed a few changes and have updated the preview: https://arelbundock.com/polars

>     1. The description text before the usage section is too big, it's the same size as section titles.

I think this is fixed now, yes?

>     2. For highlighting the usage section, it would have to be done before fixing the function names because `downlit` doesn't work with the `<>`:

Makes sense, but frankly this doesn't feel high priority to me, and I don't really have the energy.

>     3. Regarding sections in the sidebar, I think they should be ordered alphabetically and that we should rename the `Polars` section ("Operations" maybe?).

I think we should try to follow the Python/Rust section ordering, which is not alphabetical. "Operations" has the same problem as "Polars": both are too generic. Eventually, we probably want to be even closer to Python/Rust API reference, and sort the grab-bag of current "Polars" section in Input/Output, Functions, Extending the API. See: https://pola-rs.github.io/polars/py-polars/html/reference/

>     4. I really prefer having the function title as header instead of the function name (using title as header is also what `pkgdown` does).

Yes, sorry, I misunderstood the problem. You are 100% right. I already pushed a change for that, and we have proper titles now.

> Finally, is it fine to push some commits in your branch or do you prefer handling this yourself?

Yep, that's fine. I'm traveling next week, so won't be able to work on this PR anymore. I can either merge it nearly as-is, or hand it off to you to complete.

Thanks again for putting together the website. It looks super cool.

etiennebacher:
Re title and description: yes all good now

Re code highlighting: completely agree, low priority thing

Re section organization: yes I'll try to match the organization of the Python docs more closely. Once this is done, the "Polars" section will no longer be needed 

I'd prefer working on this on a branch of `main` instead of on your fork so I can review this now and merge it. Can you fix the conflicts and request my review?

And thanks also for improving the website, adding keywords and fixing the docs is really tedious work

etiennebacher (changes requested):
Actually I noticed a few things that should be fixed before merging:

* evaluation of some examples fail, e.g https://arelbundock.com/polars/reference/DataFrame_join/
* we lost the link to the source file between title and description sections
* the "About" tab doesn't need to be a vignette, we can just make it a permanent file in `docs`

@vincentarelbundock the first point is the most important and I can take care of it in this branch before deploying the docs.

vincentarelbundock:
> Actually I noticed a few things that should be fixed before merging:
> 
>     * evaluation of some examples fail, e.g https://arelbundock.com/polars/reference/DataFrame_join/

Good catch! The problem is that I was not careful enough in doing the `<DataFrame>` sub here:
https://github.com/pola-rs/r-polars/pull/186/files#diff-ad7021ea0176aafa82074234706eef90248e9f733c463313f0dcd98f86d3f211R64

Should apply more narrowly to just the Usage section.

> @vincentarelbundock the first point is the most important and I can take care of it in this branch before deploying the docs.

Thanks! I appreciate that. Feel free to push directly to the branch. (I need to disconnect now.)

etiennebacher (approved):
Thanks a lot @vincentarelbundock ! This ticks several boxes in #182

PR #188, opened 2023-05-01
Include a performance vignette?

### LINKED ISSUE #176: Add example for memory and speed gain when using lazy?
So far, the "Get started" vignette explains that using lazy frames can lead to large gains in speed and memory, but the example doesn't really show the extent of these gains (which is acknowledged in the vignette).

Here's a small example of how `polars` can reorganize the query in order to optimize it (much faster and more memory-efficient to do filter-sort than sort-filter). Should something like this be included in the vignette or elsewhere?

``` r
library(polars)
library(ukbabynames)

# make the dataset
test <- data.frame(
  x = sample(ukbabynames::ukbabynames$name, 5*1e7, TRUE),
  y = sample(1:1000, 5*1e7, TRUE),
  z = sample(1:100, 5*1e7, TRUE)
)
readr::write_csv(test, "foo.csv")

eager <- pl$read_csv("foo.csv")
lazy <- pl$read_csv("foo.csv", lazy = TRUE)

lazy_query <- lazy$
  sort(pl$col("x"))$
  filter(
    pl$col("y") > 50 & 
      pl$col("x")$is_in(pl$lit(c("John", "Mary")))
  )

lazy_query$describe_plan()
#>   FILTER [([(col("y")) > (50f64)]) & (col("x").is_in([Series]))] FROM
#>     SORT BY [col("x")]
#>       CSV SCAN foo.csv
#>       PROJECT */3 COLUMNS
lazy_query$describe_optimized_plan()
#>   SORT BY [col("x")]
#>     CSV SCAN foo.csv
#>     PROJECT */3 COLUMNS
#>     SELECTION: [([(col("y").cast(Float64)) > (50f64)]) & (col("x").is_in([Series]))]

bench::mark(
  eager = eager$
    sort(pl$col("x"))$
    filter(
      pl$col("y") > 50 & 
      pl$col("x")$is_in(pl$lit(c("John","Mary")))
    ),
  lazy = lazy_query$collect()
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 eager        11.32s   11.32s    0.0883    13.4KB        0
#> 2 lazy          1.17s    1.17s    0.858       624B        0
```

<sup>Created on 2023-04-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

etiennebacher:
@sorhawell @eitsupi @vincentarelbundock @grantmcdermott I started a vignette on `polars` performance, not to compare it to other packages but rather to present a few "good practices" (?) to use its full capabilities:

* lazy eval > eager eval because 1) it doesn't load data in memory and 2) it optimizes the queries under the hood before applying them to the data
* better to use `polars`' built-in functions rather than passing R functions to a `DataFrame` 
* streaming data? It's something I've seen mentioned in the docs but I didn't explore this yet

I'm still a beginner in `polars` and in data wrangling with larger-than-RAM data so there might be some things to correct/complete here. Also, I mostly wrote this for me to have some explanations somewhere and because it might be useful if I end up teaching this, but it doesn't have to be included as a vignette.  

What do you think about this? 

Close #176


eitsupi:
Since it maybe difficult to run benchmarks on CI, I think we need to investigate how other repositories include benchmarks in their articles.
I don't know of many examples, but something like https://github.com/tidyverse/vroom for example.

performance.md
```diff
@@ -0,0 +1,227 @@
...
+the base R function `grepl()`. However, using the built-in function is
+much faster:
+
+``` r
+bench::mark(
+  contains = df_test$with_columns(
+    pl$col("country")$str$contains("na")
+  ),
+  grepl = df_test$with_column(
+    pl$col("country")$map(\(s) { # map with a R function
```
vincentarelbundock, line 207: I think it's important to make that point, but also expect that a number of readers will find this comparison unfair. Isn't it like applying `grepl()` to every element individually instead of the vector? That's against the R mantra of vectorizing operations...

sorhawell:
> May be relevant: https://github.com/pola-rs/tpch

I was playing around a bit with tpch about August last year.  But a lot of features were missing in r-polars back then. Some of test datasets did not compile out-of-the-box and required some hand held fixing on my machine. Maybe tpch has become more ergonomic now.

I think a tpch benchmark would be the ultimate approval that r-polars is on par with py-polars-

performance.Rmd
```diff
@@ -0,0 +1,184 @@
...
+read it lazily with `lazy_csv_reader()`:
+
+```{r echo = FALSE, eval = !exists("data/test.csv")}
+set.seed(123)
+test = data.frame(
+  country = sample(countries, 1e7, TRUE),
+  x = sample(1:100, 1e7, TRUE),
+  y = sample(1:1000, 1e7, TRUE)
+)
+readr::write_csv(test, "data/test.csv")
```
sorhawell, line 82: maybe readr should be added to suggested

performance.Rmd
```diff
@@ -0,0 +1,184 @@
...
+
+lazy_query
+```
+
+However, this doesn't do anything to the data until we call `collect()` at the 
+end. We can now compare the two approaches:
+
+```{r}
+bench::mark(
+  eager = df_test$
```
sorhawell, line 120: reading the csv every time in eager mode, might be a more fair comparison

```r
bench::mark(
  eager = {
    df_test = csv_reader("data/test.csv")
    df_test$
    sort(pl$col("country"))$
    filter(
      pl$col("country")$is_in(pl$lit(c("United Kingdom", "Japan", "Vietnam")))
    )
  },
  lazy = lazy_query$collect(),
  iterations = 10
)
```

performance.Rmd
```diff
@@ -0,0 +1,184 @@
...
+  lazy = lazy_query$collect(),
+  iterations = 10
+)
+```
+
+
+On this very simple query, using lazy execution instead of eager execution lead
+to a 1.7-2.2x decrease in time. 
+
+So what happened? Under the hood, `polars` reorganized the query so that it filters
```
sorhawell, line 134: ... reorganized the query **so that it filters rows already while reading the csv into memory**, and then sorts the remaining data. This can be seen by comparing ...

performance.md
sorhawell: I was surprised Series can be passed directly to grepl, I guess that is all the sweet S3 methods at play :)

sorhawell: map has a very slow cousing called apply which should never be used except for in group_by calls.
If no group_by apply will iterate and convert every element individually

```R
grepl_scalarized = df_test$limit(1e6)$with_column( #10 x fewer
    pl$col("country")$apply(\(str) { # map with a R function
      grepl("na", str)
    }, return_type = pl$Boolean)
  )
```

in a group_by context apply iterate over the group, and perform a map on each group. If using map in a group_by context, it ignores the grouping and transform the raw column.

sorhawell: btw the map could be written with explicit conversions, depending on style


```R
  grepl_explicit = df_test$with_column(
    pl$col("country")$map(\(s) { # map with a R function
      pl$Series(grepl("na", s$to_vector())) #explicit conversion
    })
  ),

```

sorhawell:
@etiennebacher  I'm very positive about this draft :)

performance.Rmd
vincentarelbundock: Probably should just use the base `write.csv` instead. No reason to load a new dependency just for writing a CSV...

vincentarelbundock:
Maybe you can add something ultra simple but fun like this, and then point readers to the DuckDB benchmarks for more serious stuff. The idea is would be to give readers an early "Wow!".

``` r
library(bench)
library(dplyr)
library(polars)
library(data.table)

N = 1e7
df = data.frame(matrix(runif(25 * N), nrow = N))
df$letters = sample(letters, N, replace = TRUE)

df_dt = data.table(df)
df_pl = pl$DataFrame(df)

# comparison
bench::mark(
    "base" = by(df, df$letters, \(x) colMeans(x[, -26])),
    "dplyr" = df %>% group_by(letters) %>% summarise_all(mean),
    "data.table" = df_dt[, lapply(.SD, mean), by = "letters"],
    "polars" = df_pl$groupby("letters")$mean(),
    check = FALSE,
    relative = TRUE
)
#   expression   min median `itr/sec` mem_alloc `gc/sec`
#   <bch:expr> <dbl>  <dbl>     <dbl>     <dbl>    <dbl>
# 1 base       16.1   15.6       1        8241.      Inf
# 2 dplyr       9.66   9.35      1.67     3826.      Inf
# 3 data.table  2.29   2.26      6.92      634.      NaN
# 4 polars      1      1        15.5         1       NaN
```

grantmcdermott:
I didn't want to spam everyone's inbox&mdash;so let me know if other would like to join too&mdash;but I invited @vincentarelbundock and @etiennebacher to a private repo that houses (an adapted subset of) benchmarks that I keep for myself on some common data tasks across a variety of languages and libraries. Feel free to poke around etc. I also have timings for larger datasets, but this ends up being a bottleneck for some languages (_cough Stata cough_).

grantmcdermott:
> I'm still a beginner in polars and in data wrangling with larger-than-RAM data so there might be some things to correct/complete here. 

AFAIK r-polars does not support streaming yet. See the py-polars handbook for some [simple examples](https://pola-rs.github.io/polars-book/user-guide/concepts/streaming/). _tl;dr_ just end your query with `collect(steaming=True)`.

performance.Rmd
etiennebacher: sure, I just used `readr` because it much faster than `write.csv` but it's not necessary, I removed it

performance.Rmd
etiennebacher: done

performance.Rmd
etiennebacher: done

performance.md
etiennebacher: nice, thanks for the details

etiennebacher:
Thanks all, FYI if you're interested feel free to push changes directly to this PR

performance.Rmd
```diff
@@ -0,0 +1,227 @@
...
+  library(dplyr)
+  library(polars)
+  library(data.table)
+})
+
+N = 1e7
+df = data.frame(matrix(runif(25 * N), nrow = N))
+df$letters = sample(letters, N, replace = TRUE)
+
+df_dt = setDT(df)
```
vincentarelbundock, line 22: Out of principle, this should probably be `data.table(df)` to avoid changing `df` itself, and the benchmark would probably be fairer without `setkey()` since the other methods do not benefit from this pre-processing.

eitsupi:
I am reluctant to make comparisons with `dplyr` or `data.table` here.
(Is there any reason to include `dplyr` and `data.table` but not Acero (`arrow`) and `duckdb`?)

vincentarelbundock:
> I am reluctant to make comparisons with `dplyr` or `data.table` here.

Well, one goal here is obviously to convince R users that it's worth it for them to try out `polars`. One way to do that is to show that it'll be faster than what they currently use, and 98% of R users currently rely on `base`, `dplyr`, or `data.table`. So in pure "marketing" terms, it seems pretty important to have this there. And if we clearly note that these are not extensive rigorous benchmarks, and point to the DuckDB page, then it is "honest" marketing that we can feel good about.

>  Is there any reason to include `dplyr` and `data.table` but not Acero (`arrow`) and `duckdb`?

No, I think those should be added too if it's easy. I would also be curious to know what people think the benefits of `polars` are over `duckdb` (assuming the performance is similar.

eitsupi:
For example, when I tried Acero and duckdb in my environment, I got the following results.
However, duckdb converts the results to R DataFrame, which may not be fair compared to Acero and polars, which require an additional cost when converting to DataFrame.

``` r
library(bench)
library(data.table)
library(dplyr, warn.conflicts = FALSE)
library(arrow, warn.conflicts = FALSE)
library(duckdb)
#> Loading required package: DBI
#>
#> Attaching package: 'duckdb'
#> The following object is masked from 'package:dplyr':
#>
#>     sql
library(polars)

N = 1e7
set.seed(1)
df = data.frame(matrix(runif(25 * N), nrow = N))
df$letters = sample(letters, N, replace = TRUE)

df_dt = data.table(df)
at = as_arrow_table(df)
df_pl = pl$DataFrame(df)

con = DBI::dbConnect(duckdb::duckdb(), ":memory:")
duckdb_register(con, "df", df)

# comparison
bench::mark(
    "data.table" = df_dt[, lapply(.SD, mean), by = "letters"],
    "Acero" = at |> group_by(letters) |> summarise(across(!letters, ~ mean(.x, na.rm = TRUE))) |> compute(),
    "duckdb" = duckdb::sql("FROM df SELECT letters, avg(COLUMNS(x -> NOT suffix(x, 'letters'))) GROUP BY letters", con),
    "polars" = df_pl$groupby("letters")$mean(),
    check = FALSE,
    relative = TRUE
)
#> # A tibble: 4 × 6
#>   expression   min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <dbl>  <dbl>     <dbl>     <dbl>    <dbl>
#> 1 data.table  6.68   6.29      1      35566.       NaN
#> 2 Acero       1.39   1.31      4.81     821.       Inf
#> 3 duckdb      1.71   1.61      3.89       1        NaN
#> 4 polars      1      1         6.29      56.2      NaN

sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.0 (2023-04-21)
#>  os       Ubuntu 22.04.2 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Etc/UTC
#>  date     2023-05-05
#>  pandoc   3.1.2 @ /usr/bin/ (via rmarkdown)
#>
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version  date (UTC) lib source
#>  arrow       * 11.0.0.3 2023-03-08 [1] RSPM
#>  assertthat    0.2.1    2019-03-21 [1] RSPM
#>  bench       * 1.1.2    2021-11-30 [1] RSPM
#>  bit           4.0.5    2022-11-15 [1] RSPM (R 4.3.0)
#>  bit64         4.0.5    2020-08-30 [1] RSPM (R 4.3.0)
#>  cli           3.6.1    2023-03-23 [1] RSPM
#>  data.table  * 1.14.8   2023-02-17 [1] RSPM
#>  DBI         * 1.1.3    2022-06-18 [1] RSPM (R 4.3.0)
#>  digest        0.6.31   2022-12-11 [1] RSPM
#>  dplyr       * 1.1.2    2023-04-20 [1] RSPM (R 4.3.0)
#>  duckdb      * 0.8.0    2023-05-05 [1] https://duckdb.r-universe.dev (R 4.3.0)
#>  evaluate      0.20     2023-01-17 [1] RSPM
#>  fansi         1.0.4    2023-01-22 [1] RSPM
#>  fastmap       1.1.1    2023-02-24 [1] RSPM
#>  fs            1.6.2    2023-04-25 [1] RSPM (R 4.3.0)
#>  generics      0.1.3    2022-07-05 [1] RSPM (R 4.3.0)
#>  glue          1.6.2    2022-02-24 [1] RSPM
#>  htmltools     0.5.5    2023-03-23 [1] RSPM
#>  knitr         1.42     2023-01-25 [1] RSPM
#>  lifecycle     1.0.3    2022-10-07 [1] RSPM
#>  magrittr      2.0.3    2022-03-30 [1] RSPM
#>  pillar        1.9.0    2023-03-22 [1] RSPM
#>  pkgconfig     2.0.3    2019-09-22 [1] RSPM
#>  polars      * 0.6.0    2023-05-04 [1] local
#>  profmem       0.6.0    2020-12-13 [1] RSPM
#>  purrr         1.0.1    2023-01-10 [1] RSPM
#>  R.cache       0.16.0   2022-07-21 [1] RSPM
#>  R.methodsS3   1.8.2    2022-06-13 [1] RSPM
#>  R.oo          1.25.0   2022-06-12 [1] RSPM
#>  R.utils       2.12.2   2022-11-11 [1] RSPM
#>  R6            2.5.1    2021-08-19 [1] RSPM
#>  reprex        2.0.2    2022-08-17 [1] RSPM
#>  rlang         1.1.0    2023-03-14 [1] RSPM
#>  rmarkdown     2.21     2023-03-26 [1] RSPM
#>  sessioninfo   1.2.2    2021-12-06 [1] RSPM
#>  styler        1.9.1    2023-03-04 [1] RSPM
#>  tibble        3.2.1    2023-03-20 [1] RSPM
#>  tidyselect    1.2.0    2022-10-10 [1] RSPM (R 4.3.0)
#>  utf8          1.2.3    2023-01-31 [1] RSPM
#>  vctrs         0.6.2    2023-04-19 [1] RSPM
#>  withr         2.5.0    2022-03-03 [1] RSPM
#>  xfun          0.39     2023-04-20 [1] RSPM
#>  yaml          2.3.7    2023-01-23 [1] RSPM
#>
#>  [1] /usr/local/lib/R/site-library
#>  [2] /usr/local/lib/R/library
#>
#> ──────────────────────────────────────────────────────────────────────────────
```

<sup>Created on 2023-05-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

eitsupi:
Here are what I consider to be the rough pros and cons:

## Acero

### Pros

- In many cases, high performance can be achieved using the dplyr syntax as is.

### Cons

- Winodw functions are not supported.

## DuckDB

### Pros

- Queries can be written in SQL and are highly portable.
- dbplyr allows the dplyr syntax to be used almost verbatim.

### Cons

- It adds SQL-extended syntax such as COLUMNS and EXCLUDE, but does not support column selection or column renaming as flexibly as dplyr (or polars).

## Polars

### Pros

- It is considered one of the fastest DetaFrame libraries.

### Cons

- It is necessary to learn its own syntax.
- There are many breaking changes.

eitsupi:
I really don't think it is a good idea to make a bare dplyr comparison here, because in my opinion dplyr users can achieve higher speeds by simply switching to data.table (via dtplyr) or Acero (via arrow) or duckdb (via dbplyr) as a backend.
(Note that these different backends are described in dplyr's README)

vincentarelbundock:
Why don't we just add a dtplyr example, then?

One other benefit of polars, I think, is the parallelism in syntax across R, python, and Rust, which facilitates multilingual projects and teams

tdhock:
I would recommend using asymptotic benchmarks, which means measuring time and memory for data size N values increasing on a log scale. I have a package https://github.com/tdhock/atime that makes this easy. These benchmarks are much more convincing than single N, which can be misleading (which N is relevant to test may depend on the particular problem/hardware so much more informative/convincing to see result for several N).

etiennebacher:
I don't think we should include benchmarks with `data.table`/`dplyr`/`arrow`, etc., or at least not in this vignette (so we could change its name).

To me, the objective of this vignette is not to compare `polars` to other packages or tools because our benchmarks will never be as comprehensive as the ones run by `duckdb`. Also, if we start doing this, then we must make a lot of choices (including those discussed above): should we count data reading in the timing? should we use keyed `data.table`s? should we compare to `arrow`, `duckdb`? how many observations should we keep? etc. 

I think it's more important here to focus on how one can use `polars`' full capabilities, because it's not something that the average R user knows (e.g I guess few R users know the difference between eager and lazy execution). If we want to advertise the speed of `polars`, couldn't we just take a graph from `duckdb`'s benchmarks?

--- 

@tdhock thanks for the link, I think [`bench::press()`](https://bench.r-lib.org/reference/press.html) does something similar?

tdhock:
yes, bench::press does something similar, and that is discussed in the Related work section of the atime README, https://github.com/tdhock/atime#related-work
```
[bench](https://cloud.r-project.org/web/packages/bench/)::press does something similar, and is more flexible because it can do multi-dimensional grid search (not only over a single size N argument as atime does). However it can not store results if check=FALSE, results must be equal if check=TRUE, and there is no way to easily specify a time limit which stops for larger sizes (like seconds.limit argument in atime).
```

grantmcdermott:
One other (tbc?) Pro for Polars is that multithreading automatically works on MacOS.

I might be missing something about the R-universe build process, but enabling multithreading in other high performance R libraries can be a bit of a pain. That's because these are C/C++ based and the OpenMP toolchain has to be installed and then linked to manually. (Basically, you have to specify a bunch of, e.g., C++ flags in your Makevars and then build from source instead of installing binaries.)

eitsupi:
> I might be missing something about the R-universe build process

Do you mention that SIMD is disabled in the R-universe builds?
https://github.com/pola-rs/r-polars/pull/78#issuecomment-1479702330

sorhawell:
> Do you mention that SIMD is disabled in the R-universe builds?

We do not :/ but should. I know of no benchmarks so far revealing, when it matters.

etiennebacher:
Can someone review the content of this vignette? I included the `.md` file so that it's easier to review from Github but we'll need to remove it before merging since it's not expected by R CMD check.

vignettes/performance.Rmd
```diff
@@ -0,0 +1,223 @@
...
+  sort_filter = {
+    tmp = test[order(test$country), ]
+    subset(tmp, country %in% c("United Kingdom", "Japan", "Vietnam"))
+  },
+  filter_sort = {
+    tmp = subset(test, country %in% c("United Kingdom", "Japan", "Vietnam"))
+    tmp[order(tmp$country), ]
+  }
+)
+```
```
eitsupi, line 71: I don't know if it is a good idea to do the benchmarking on a GitHub runner every time it is deployed.

eitsupi (approved):
Thanks for working on this!
Sorry for the late review.

vignettes/performance.Rmd
etiennebacher: True, maybe I can cache the results of the benchmarks that are not really related to `polars` (like this one) but those that compare eager and lazy queries we want their timings to be up to date, right?

etiennebacher: In the end I just put all current results and do not evaluate the benchmarks. They should probably be updated from time to time

etiennebacher:
Thanks for the review @eitsupi

PR #190, opened 2023-05-02
docs: regen some Rd files by released version of roxygen2

eitsupi:
It seems that the Rd files were not updated in the previous PR, so I regenerated these.


eitsupi:
I think the test failure is due to waldo updates (0.5.0).

> NA_character_ and "NA" are no longer treated as equal (https://github.com/r-lib/waldo/issues/162).
https://github.com/r-lib/waldo/releases/tag/v0.5.0

sorhawell (approved):

eitsupi:
Thanks for checking this!

Issue #135, opened 2023-04-19
scan_parquet very slow with large datasets from VS Code


grantmcdermott:
Has anyone else experienced this? I've confirmed that the same thing is happening for me on both Mac and Linux (Arch).

I first noticed it when scanning a large number of parquet files. However,  I don't think the number of files is a problem _per se_. It's rather the total size. For the below example, I'm scanning 1 month of NYC taxi data that I downloaded from AWS S3 like so:

```sh
mkdir -p taxi-data/year=2012
aws s3 cp s3://ursa-labs-taxi-data/2012/01 taxi-data/year=2012/month=01 --recursive
```

Running through a regular R terminal:

```r
library(polars)

tic = proc.time()
nyc = scan_parquet("/Users/gmcd/taxi-data/*/*/*.parquet")
(toc = proc.time() - tic)
#>  user  system elapsed 
#> 0.009   0.003   0.012
```

Running through VS Code:

```r
library(polars)

tic = proc.time()
nyc = scan_parquet("/Users/gmcd/taxi-data/*/*/*.parquet")
(toc = proc.time() - tic)
#>   user  system elapsed 
#> 29.834  11.322   7.774 
```

(So, a ~650x slowdown).

What's really weird is that this slowdown disappears if I collect immediately. (Which you'd obviously think would be _more_ expensive.) Here's the same chunk again from VS Code, except I chain `$collect()` at the end of the call.

```r
library(polars)

tic = proc.time()
nyc = scan_parquet("/Users/gmcd/taxi-data/*/*/*.parquet")
(toc = proc.time() - tic)
#>  user  system elapsed 
#> 5.593   1.953   0.893 
```

(Huh? Slowdown disappears.)

Some additional notes:

- It's not just scanning itself: Computation operations are also affected if they are applied to scanned-but-not-yet-collected objects. E.g. running`nyc$groupby("passenger_count")$agg(pl$col("fare_amount")$mean())$collect()` adds another 10 seconds to the VS code chunk, but a fraction of a second for the R terminal.
   - At the same time, if I run the same command on a collected dataset, then performance is back to where it should be.

- The problem appears to get increasingly worse with the size of datasets. E.g. Downloading another month's worth of taxi data and running the same chunk from VS code more than doubles the wait time.

- Looking at `htop` while this is all going on shows that my memory swap partition is being crushed (completely used up) and all cores are active throughout.

[...]

PR #193, opened 2023-05-03
debug/profile polars

sorhawell:
In issue #135 we had some non obvious run-away S3 methods hampering performance. To assist future debugging/profiling such issues, I suggest to add a simple polars class call-stack viewer / profiler.

The little amount of extra indirection/overhead to activate this, is `if(polars_optenv$debug_polars) { #debug stuff }` in every public/private method-call and pl$-function call.

It may also improve bug-reports. It may help with further development of S3-methods.

## use case
```R
pl$set_polars_options(debug_polars=TRUE) #enable

# run some polars code
> pl$DataFrame(iris)$select("Species","Sepal.Length")$limit(10)
[9999ms]
pl$DataFrame() -> [1.973ms]
   .pr$DataFrame$new_with_capacity() -> [1.188ms]
   .pr$DataFrame$set_column_from_robj() -> [2.131ms]
   .pr$DataFrame$set_column_from_robj() -> [1.701ms]
   .pr$DataFrame$set_column_from_robj() -> [1.585ms]
   .pr$DataFrame$set_column_from_robj() -> [1.453ms]
   .pr$DataFrame$set_column_from_robj() -> [2.558ms]
DataFrame$select() -> [0.953ms]
ProtoExprArray$push_back_rexpr() -> [1.961ms]
pl$col() -> [1.382ms]
   .pr$Expr$col() -> [0.9389ms]
ProtoExprArray$push_back_rexpr() -> [0.885ms]
pl$col() -> [1.932ms]
   .pr$Expr$col() -> [1.98ms]
   .pr$DataFrame$select() -> [2.111ms]
DataFrame$limit() -> [1.115ms]
DataFrame$lazy() -> [0.458ms]
LazyFrame$limit() -> [0.2651ms]
   .pr$LazyFrame$limit() -> [0.355ms]
LazyFrame$collect() -> [0.3009ms]
   .pr$LazyFrame$collect() -> [0.6909ms]
DataFrame$print() -> [1.349ms]
   .pr$DataFrame$print() -> shape: (10, 2)
┌─────────┬──────────────┐
│ Species ┆ Sepal.Length │
│ ---     ┆ ---          │
│ cat     ┆ f64          │
╞═════════╪══════════════╡
│ setosa  ┆ 5.1          │
│ setosa  ┆ 4.9          │
│ setosa  ┆ 4.7          │
│ setosa  ┆ 4.6          │
│ …       ┆ …            │
│ setosa  ┆ 4.6          │
│ setosa  ┆ 5.0          │
│ setosa  ┆ 4.4          │
│ setosa  ┆ 4.9          │
└─────────┴──────────────┘
```


etiennebacher:
There's sth wrong with the first timing, it can't be 9999ms

sorhawell:
> There's sth wrong with the first timing, it can't be 9999ms

It is a very naive timer which computes  `this_time - last_time`, where only last_time is stored in `runtime_state$last_subtime`. The first time will always be 9999ms or how long time since last timer call. Maybe I could wrap all methods with a timerfunction, but it would become more complicated.

sorhawell:
I tried to write a scoped profiler, which starts timing before executing any method and stops after. It also kept track of the polars method call-stack. However, this made the class system code more complicated/ugly and probably more difficult to maintain. Probably if this simple `debug_polars` is not telling enough, then `profvis` and/or `browser()` are better alternatives.

etiennebacher:
Thanks, there's no need to complexify all of this. Maybe it's just possible to remove the first timing which is always `[9999ms]`? Like replace it with `[TIME]` or sth else?

Otherwise it's not a big deal, it's already useful as is

sorhawell:
@etiennebacher it looks like this now. I think this option could be quite helpful.

```R
> pl$DataFrame(iris)$lazy()$select(pl$col("Sepal.Length")$mean())$collect()
[TIME? ms]
pl$DataFrame() -> [0.483ms]
   .pr$DataFrame$new_with_capacity() -> [0.344ms]
   .pr$DataFrame$set_column_from_robj() -> [0.2971ms]
   .pr$DataFrame$set_column_from_robj() -> [0.262ms]
   .pr$DataFrame$set_column_from_robj() -> [0.253ms]
   .pr$DataFrame$set_column_from_robj() -> [0.2599ms]
   .pr$DataFrame$set_column_from_robj() -> [0.298ms]
DataFrame$lazy() -> [0.237ms]
LazyFrame$select() -> [0.21ms]
pl$col() -> [0.164ms]
   .pr$Expr$col() -> [0.237ms]
Expr$mean() -> [0.242ms]
ProtoExprArray$push_back_rexpr() -> [0.18ms]
   .pr$LazyFrame$select() -> [0.206ms]
LazyFrame$collect() -> [0.169ms]
   .pr$LazyFrame$collect() -> [0.7479ms]
DataFrame$print() -> [0.174ms]
   .pr$DataFrame$print() -> shape: (1, 1)
┌──────────────┐
│ Sepal.Length │
│ ---          │
│ f64          │
╞══════════════╡
│ 5.843333     │
└──────────────┘
> 
```

sorhawell:
That is a bit confusing, it arises from this is the defalt param.

> #' @param return_replaced_options return previous state of modified options
#' Convenient for temporarily swapping of options during testing.
#'

I have made the previous state invisible.


If #201  is accepted then `debug_polars` can also be activated like this

```R
pl$options$debug_polars(TRUE)
```

etiennebacher (approved):
I don't know in which order you want to merge your PRs but it looks good to me,

Do you want to mention somewhere in the docs that users can use this option if they feel that `polars` is abnormally slow?

sorhawell:
> Do you want to mention somewhere in the docs that users can use this option if they feel that polars is abnormally slow?

I have added a mention in the bottom of README and in NEWS

README.Rmd
```diff
@@ -274,3 +274,43 @@ cd r-polars
 - The `RPOLARS_RUST_SOURCE` environment variable allows **polars** to recover the Cargo cache even if source files have been moved. Replace with your own absolute path to your local clone!
 - `filter_rcmdcheck.R` removes known warnings from final check report.
 - `unlink("check")` cleans up.
+
+### Misc
+
+If experiencing unexpected sluggish performance, when using polars in a given IDE, we like to here about it. You can try to activate `pl$set_polars_options(debug_polars = TRUE)` to profile what methods are being touched (not necessarily run) and how fast. Below is printed an example of good behavior.
```
etiennebacher, line 283:
```suggestion
If you experience unexpected sluggish performance, when using polars in a given IDE, we'd like to hear about it. You can try to activate `pl$set_polars_options(debug_polars = TRUE)` to profile what methods are being touched (not necessarily run) and how fast. Below is an example of good behavior.
```

etiennebacher (approved):

PR #195, opened 2023-05-04
fix dl-link macos readme

sorhawell:


sorhawell:
@eitsupi  currently the github dl for macos does not work as filename is change to xxyz_darwin20_xyz instead of 17.
This PR changes url in README such that no mac users do not get tricked and give up.

eitsupi (approved):

eitsupi (changes requested):
We should update the Rmd file.

eitsupi (approved):
Thanks!

PR #200, opened 2023-05-08
Docs: misc fixes

etiennebacher:
* Fix examples that failed because `<` and `>` were not parsed correctly (for example [filter](https://rpolars.github.io/reference/DataFrame_filter/))
* in Reference, auto collapse of sections in sidebar if we open another section
* Replace `<h3>` tags by `##` so that subsections appear in the TOC on the right of docs. Example:

image


eitsupi (approved):
Thanks!

Issue #95, opened 2023-03-26
Refactor options.R into Config.R based on py-polars


sorhawell:
R uses options.R to handle R specific polars settings. It should be refactored to behave like py-polars Config + R options. If no other reason, the state should be kept in env-vars. py-polars Config has some methods for using scoped config ( `with Config as cfg: ...` ). Nice if that could be translated according to some R syntax.

py-polars
https://github.com/pola-rs/polars/blob/master/py-polars/polars/config.py

polars core re mapping
https://github.com/pola-rs/polars/blob/8763812c8b83e5ebd6e044523f7c543a5473a705/polars/polars-core/src/config.rs#L2

PR #201, opened 2023-05-08
Refactor groupby + defualt_maintain_order option

sorhawell:
This PR refactors GroupBy methods to latest parameter conversion pattern.

and

add global option defualt_maintain_order. The default is FALSE. 
```R
#' @rdname polars_options
#' @name defualt_maintain_order
#' @details sets maintain_order = TRUE as default
#' implicated methods/functions are currently: DataFrame_GroupBy + LazyFrameGroupby.
#' @param defualt_maintain_orderr bool, default = FALSE
polars_optenv$defualt_maintain_order = FALSE #set default value
```

example:

```R
#' Lazy_groupby
#' @description apply groupby on LazyFrame, return LazyGroupBy
#' @keywords LazyFrame
#' groupby on LazyFrame.
#'
#' @param ... any single Expr or string naming a column
#' @param maintain_order Nullable bool,  should an aggregate of groupby retain order of groups?
#' FALSE = random, slightly faster, but not dertimistic order. Default is NULL means FALSE, unless
#' `pl$set_polars_options(default_maintain_order = TRUE)`, then NULL mean TRUE.
#'
#' @return A new `LazyGroupBy` object with applied groups.
LazyFrame_groupby = function(..., maintain_order = NULL) {
  largs  = list2(...)
  maintain_order = maintain_order %||% polars_optenv$default_maintain_order
  unwrap(.pr$LazyFrame$groupby(self,largs,maintain_order))
}
```


sorhawell:
@eitsupi did you imagine a maintain_order option like this?

```R
pl$set_polars_options(defualt_maintain_order = TRUE)
```

eitsupi:
Thanks for working on this.

> @eitsupi did you imagine a maintain_order option like this?
> 
> ```r
> pl$set_polars_options(default_maintain_order = TRUE)
> ```

I was assuming that the settings would be made by the `options` function, which is common in R.
Such as the one used below in this package.

https://github.com/pola-rs/r-polars/blob/198992c8dfbefd6620a27a7126ec22f08a4adf0b/R/pkg-knitr.R#L19

Something like that:

```r
options(polars.groupby_maintain_order = TRUE)
```

vincentarelbundock:
You could also call ˋgetOption()` right in the function signature as the default argument value, which is nice because it is kind of self-documenting...

eitsupi:
We can also create a dedicated function like `readr::should_show_types()` and get option in that function.
https://github.com/tidyverse/readr/blob/9558d899b7e8c592be339bfc31ec35593ccf20ca/R/read_delim.R#L190-L203
https://github.com/tidyverse/readr/blob/9558d899b7e8c592be339bfc31ec35593ccf20ca/R/utils.R#L34-L56

sorhawell:
@vincentarelbundock  @eitsupi 

It seemed getOption was designed for read from R options(). r-polars currently save to private environment `polars_optenv` [but should be upgraded to store state in environment variables](https://github.com/pola-rs/r-polars/issues/95), which are also easier to read from rust side.

Do you have examples of R packages that read or write options as environment variables?

I tinkered  and made a safe simple interface `pl$options$` which behaves and looks like a list, but it is not possible to set invalid values.

```R
# polars options access with `pl$options`
pl$options
[1] "pl$options (polars_options, list-like with value validation):"
$default_maintain_order
[1] FALSE

$do_not_repeat_call
[1] FALSE

$strictly_immutable
[1] TRUE

$named_exprs
[1] TRUE

$no_messages
[1] FALSE
pl$options$strictly_immutable
[1] TRUE
pl$options$default_maintain_order
[1] FALSE
# cannot write wrong arguments to options
tryCatch({pl$options$default_maintain_order <- 42}, error = function(err) cat(as.character(err)))
Error: option: default_maintain_order  must satisfy requirement named is_bool with function
 function (x) {
    is.logical(x) && length(x)==1 && !is.na(x)
  }
<bytecode: 0x7feed1ff7750>
<environment: namespace:polars>
´´´

grantmcdermott:
> Do you have examples of R packages that read or write options as environment variables?

Not sure if this is quite what you mean or are looking for, but **collapse** implements options via a special environment contained in its own package namespace (in part to avoid the overhead of `getOptions`). https://sebkrantz.github.io/collapse/reference/collapse-options.html

sorhawell:
@grantmcdermott , thx that was interesting :) it is quite similar to what we have now.

I think if prefer to merge this PR for now and return when refactoring the internal option-state in to environement_variable as py-polars does it.

sorhawell:
> 
> Something like that:
> 
> ```r
> options(polars.groupby_maintain_order = TRUE)
> ```

Oh missed this comment. Sry. I wrote the options handler/checker around a separate environment `polars_optenv`. Since it is getting refactored into be based on envvars as py-polars, I think this is ok temporarily.

I ended up cleaning up the GroupBy class, so eventhough this PR continues a home cooked option-system, it still has good value, I think.

eitsupi (approved):
Thanks. I agree with merging this for now.

PR #202, opened 2023-05-10
make build now load library arrow and nanoarrow to emsure roxygen2 renders correctly

sorhawell:
Maybe someone else have been annoyed by nanoarrow.Rd pops up alot if library `search()` does not include arrow/nanoarrow when roxygen2 is called first time in session.

```R
9 |    | \method{as_nanoarrow_array_stream}{DataFrame}(x, ..., schema = NULL)
  | 19 | as_nanoarrow_array_stream.DataFrame(x, ..., schema = NULL)

21 |   | \method{infer_nanoarrow_schema}{DataFrame}(x, ...)
  | 21 | infer_nanoarrow_schema.DataFrame(x, ...)

23 |   | \method{as_arrow_table}{DataFrame}(x, ...)
  | 23 | as_arrow_table.DataFrame(x, ...)

25 |   | \method{as_record_batch_reader}{DataFrame}(x, ..., schema = NULL)
  | 25 | as_record_batch_reader.DataFrame(x, ..., schema = NULL)
```

I suggest to add to makefile `git_untrack` and `git_retrack` which ignores any changes to a list of known annoying files.

Are other suggestions for how to fix this?


eitsupi:
I did not understand what was going on here.
Perhaps it makes more sense to force on the Makefile to load nanoarrow and arrow before generating the Rd file?

sorhawell:
> Perhaps it makes more sense to force on the Makefile to load nanoarrow and arrow before generating the Rd file?

Thx, that was a better solution. I have added nanoarrow and arrow to `make build`.

sorhawell:
this git diff became rather messy, I will redo this PR

PR #207, opened 2023-05-10
try preload nanoarrow before building polars

sorhawell:
this PR replaces #202 

loads nanoarrow and arrow in `make build` inorder for roxygen2 render nanoarrow.Rd correctly


eitsupi (approved):
Thanks!

PR #212, opened 2023-05-11
try smorimoto/tune-github-hosted-runner-network

sorhawell:
https://github.com/smorimoto/tune-github-hosted-runner-network

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -36,6 +36,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      
+      - name: Tune GitHub-hosted runner network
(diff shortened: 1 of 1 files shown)


sorhawell:
In this [issue thead](https://github.com/actions/runner-images/issues/1187) a "TCP/UDP offload" is recommended to counter flaky network errors. The configuration vary a bit depending on OS. This action perform the configuration on either 3 OS: win, mac, linux.

Two runs had no fails so far. I guess we have to merge, to try it out to see. My impression is that every 2nd or 3rd run would have some job failing at the moment. If this drops to every 10th or never, I would count it as an success.

eitsupi (approved):

Issue #214, opened 2023-05-13
Set linter and auto formatter


eitsupi:
I tried `styler.equals`, but unfortunately, this currently seems to do little more than modify the assign operator.
So perhaps `styler.equals::style_pkg()` should be executed after `styler::style_pkg()`.

sorhawell:
I also ran regular styler first and then styler.equals in this Rstudio add-in.

https://github.com/sorhawell/styler.equals/blob/main/R/saveAndStylerEqualsAddin.R

Robinlovelace:
Finally seeing this conversation. Based on what is in styler I'm unsure why `styler.equals` is failing to style packages. Glad you have found out issues with it, any ideas on solutions welcome...

eitsupi:
> Based on what is in styler I'm unsure why `styler.equals` is failing to style packages.

Thanks for the reply. It seems you've pinpointed the cause in https://github.com/Robinlovelace/styler.equals/issues/3#issuecomment-1556025539.

etiennebacher:
@eitsupi I think this can be closed?

Robinlovelace:
Just got a notification on this. Seems it can be closed to me. On a different topic, I still didn't get `styler.equals` to work : ( any ideas how I can fix it? Thanks!

etiennebacher:
I didn't understand this was the point of this issue, I thought `task format-` and the new linter CI were enough

lorenzwalthert:
> Just got a notification on this. Seems it can be closed to me. On a different topic, I still didn't get styler.equals to work : ( any ideas how I can fix it? Thanks!

Maintainer of {styler} here, I can maybe help you if you open an issue in your repo and explain the problem to me.

Robinlovelace:
Hey @lorenzwalthert I know you have already helped but would you be willing to provide some pointers on the shelved `styler.equals` package? I sense that could be an upstream fix, not 100% sure though. Thanks and apologies for jumping in on this thread.

etiennebacher:
Thanks for the offer, but AFAIK the only thing reason this issue is open is because we need to setup a CI for styler. Currently we use `styler` and modify one of its transformers to prefer `=` over `<-` (see [this script](https://github.com/pola-rs/r-polars/blob/main/dev/styler_utils.R)). Locally we just have to run `task format-r`

eitsupi:
@Robinlovelace Thank you for working on that!
But as stated above, this repository is already not using `styler.equals`, so don't worry about me.

[...]

PR #215, opened 2023-05-13
formatting: Auto formatting by styler

eitsupi:
Related to #214

TODO

- [x] Add formatting task to the Makefile

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,6 +36,7 @@ Config/Needs/website:
     yaml
 Config/Needs/dev:
     devtools,
+    styler,
     robinlovelace/styler.equals,
     rextendr
 Config/testthat/edition: 3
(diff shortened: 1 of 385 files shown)


sorhawell:
looks good to me.

I shortened an example which due to styling become wider than 100 chars.

maybe extendr-wrapper.R is better to not style as it will a constant battle between rextendr and styler.equals. On the other hand it will be very revealing when I forget to use the auto formatter :) So maybe we try like this. 

Very cool you implemented diff check to no styler all files every time.

sorhawell (approved):

eitsupi:
I update the Makefile to allow formatting based on the difference from the main branch via `make fmt GIT_DIF_TARGET=main`.
However, this could take a very long time and probably requires setting up a workflow in CI to be able to detect formatting forgetfulness.

PR #217, opened 2023-05-13
fix: pin the rust toolchain version

### LINKED ISSUE #216: rust polars doesn't support the latest Rust nightly toolchain
Build currently fails due to compatibility with 2023-05-13 nightly toolchain.
https://github.com/pola-rs/r-polars/blob/6a3522d11ddbbba9dc739918e9ee4d1ca2c10bcf/src/rust/rust-toolchain#L2
https://github.com/pola-rs/r-polars/blob/6a3522d11ddbbba9dc739918e9ee4d1ca2c10bcf/Makefile#L34-L36

Definitely need to specify a date like this.
https://github.com/pola-rs/polars/blob/64dabba165a811179197687f193037daba1103c1/rust-toolchain.toml#L2

Unfortunately, there does not seem to be a command to install the toolchain by referencing a rust-toolchain file at this time (rust-lang/rustup#2686), so it appears that the date needs to be entered in the two places above.

eitsupi:
Fix #216

### DIFF
--- a/src/rust/rust-toolchain
+++ b/src/rust/rust-toolchain
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2023-04-11"
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -44,8 +44,8 @@ jobs:
         run: print(list.files("..",recursive = TRUE,full.names=TRUE))
         shell: Rscript {0}
 
-      - run: echo "rust-toolchain=nightly" >>"$GITHUB_ENV"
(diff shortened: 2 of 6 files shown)


src/rust/rust-toolchain
eitsupi, line 1: @sorhawell This file is used for something now?
I am wondering if we can consolidate everything into a Makefile and remove this.

sorhawell: No I don't think the file has any use anymore. We can set toolchain via make and yaml.

sorhawell (approved):

PR #219, opened 2023-05-13
pin rust night in release.yaml (follow up #217)

### LINKED ISSUE #216: rust polars doesn't support the latest Rust nightly toolchain

sorhawell:
Close #216

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -33,4 +33,3 @@ runs:
         with:
           shared-key: build
           workspaces: ./src/rust/
-          cache-on-failure: "true"
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -1,0 +1,36 @@
+name: setup
+description: Setup build environment (except R)
+
+runs:
+  using: composite
+  steps:
+      - name: Install ubuntu SYSTEM REQUIREMENTS
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          sudo apt-get update \
+          && sudo apt-get install -y \
+            libfontconfig1-dev \
+            libfreetype6-dev \
+            libfribidi-dev \
+            libharfbuzz-dev \
+            libcurl4-openssl-dev \
+            libgit2-dev \
+            libicu-dev \
+            libjpeg-dev \
+            libpng-dev \
+            libtiff-dev \
+            libxml2-dev \
+            libssl-dev
+
+      - name: Set up Rust
+        shell: bash
+        run: |
+          make requirements-rs
+
+      - name: Cache Rust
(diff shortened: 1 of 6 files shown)


sorhawell:
@eitsupi 
I could not find anywhere in docs to pin rust. Is it always using a cached version of compilation only? Would that always work?

sorhawell:
#216

eitsupi:
> I could not find anywhere in docs to pin rust. Is it always using a cached version of compilation only? Would that always work?

Just using `make` command here.

https://github.com/pola-rs/r-polars/blob/bd2e7b9d3c2ea83ace051bca6ad220fcf690e88c/.github/workflows/docs.yaml#L90-L91

I have updated this PR to use `make` in other workflows.
Now we only need to specify the toolchain version in one place.

eitsupi:
@sorhawell Are the following sections still needed?
My understanding is that we are using `Swatinem/rust-cache` and don't need it anymore.
https://github.com/pola-rs/r-polars/blob/f95214fb4cd49674d975b6c216400d369a038cb7/src/Makevars#L23-L29

eitsupi (approved):
New cache workflow seems working fine, so merging now.

sorhawell:
@eitsupi I think long time ago  R CMD check gave warnings for illegal files in the target lib.  It is probably not needed anymore. If we run into that again we can rewrite it accordingly.

eitsupi:
Sorry, I was mistaken.
Perhaps what is needed is the following process.
It may easier to detect the `NOT_CRAN` environment variable and change the behavior.
I believe the latest rextendr's template does that.

https://github.com/r-rust/hellorust/blob/05a148ff4128a2480d3519a19bd5cbcd9efcbe2e/src/Makevars#L16-L19

```makefile
$(STATLIB):
	PATH="${PATH}:${HOME}/.cargo/bin" cargo build --release --manifest-path=myrustlib/Cargo.toml
	rm -Rf $(CARGOTMP) || true #CRAN wants us to remove "detritus"
	rm -Rf $(LIBDIR)/build || true
```

PR #223, opened 2023-05-16
Add a reference page for docs

etiennebacher:
Related to #182.

I'm really unsure about what we should put here. The [py-polars reference page](https://pola-rs.github.io/polars/py-polars/html/reference/index.html) just lists the sections and subsections of the sidebar, which I don't find super useful. I think it would be better to explain a bit the organization of the docs, especially for expressions because I find it very confusing. 

Also the syntax is quite strange for an average R user so I tried to explain a bit how to think about it (knowing that I don't understand all the syntax myself yet).

Feel free to modify / push to this branch, I just opened this to get the ball rolling.

Pinging @vincentarelbundock because I think you had the idea of making a reference page


docs/docs/reference_home.md
```diff
@@ -1,3 +1,36 @@
 # Reference
 
-Here's the reference homepage.
+`polars`' functions are divided in several categories.
+
+- functions that can be applied directly on `DataFrame` and `LazyFrame` (or
```
vincentarelbundock, line 5: Should we use the word "methods" everywhere?

vincentarelbundock:
Thanks Etienne. I agree that R users are likely to be unfamiliar with this programming style and that it makes sense to be a bit verbose there. I didn't have concrete ideas, but I like the direction you're going into.

etiennebacher:
Still draft. I added some details and changed the structure a bit (the [polars talk at PyCon 2023](https://youtu.be/s4vsucVSsL8?t=20185) was helpful for me to understand and if you're interested it starts at 5:25:00)

There are probably mistakes in the terminology, what I called "data structures" would probably make someone else scream

etiennebacher:
I added a bit more details, I think we "just" need to finish putting all the functions in the section "Polars" in their correct subsection in "Expr".

Are you ok with the reference page?

docs/docs/reference_home.md
```diff
@@ -1,3 +1,135 @@
...
+
+A context simply is the type of data modification that is done. There are 3 types
+of contexts:
+
+* select and modify columns with `select()` and `with_columns()`;
+* filter rows with `filter()`;
+* group and aggregate rows with `groupby()` and `agg()`
+
+Inside each context, you can use various **expressions**. Some expressions cannot
+be used in some contexts. For example, in `with_columns()`, you can only apply
```
sorhawell, line 40: of curiosity what are the general rules for other contexts? Is it the same for filter and groupby, but not for agg ?

sorhawell:
Nice intro!

I added some more examples, and rewrote "subnamespace" introduction to not call them "types" as one subnamespace `meta` is not a type.

etiennebacher:
Nice, thank you! I did minor changes and I think this is now ready for review

docs/docs/reference_home.md
```diff
@@ -1,3 +1,36 @@
...
+
+* Array: `arr`
+* DateTime: `dt`
+* String: `str`
+[... TODO]
+
+For example, suppose we want to use the function `concat()`. We could do this as
+follows:
+
+```r
```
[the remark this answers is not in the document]
etiennebacher: I think having one per data type would be a lot, given that there is already a lot of explanations

eitsupi:
> trigger CI

FYI, closing and reopening the PR can trigger the CI, so there is no need to do a commit.

etiennebacher:
Good to know, but I think it makes 2 notifs by mail instead of one

sorhawell (approved):
many thanks @etiennebacher  🚀

docs/docs/reference_home.md
```diff
@@ -1,3 +1,183 @@
...
+shape: (3, 4)
+┌─────┬──────────────────────┬──────────────┬────────────┐
+│ cyl ┆ mpg                  ┆ mpg_sliced   ┆ mpg_summed │
+│ --- ┆ ---                  ┆ ---          ┆ ---        │
+│ f64 ┆ list[f64]            ┆ list[f64]    ┆ f64        │
+╞═════╪══════════════════════╪══════════════╪════════════╡
+│ 4.0 ┆ [22.8, 24.4, … 21.4] ┆ [22.8, 24.4] ┆ 293.3      │
+│ 8.0 ┆ [18.7, 14.3, … 15.0] ┆ [18.7, 14.3] ┆ 211.4      │
+│ 6.0 ┆ [21.0, 21.0, … 19.7] ┆ [21.0, 21.0] ┆ 138.2      │
+└─────┴──────────────────────┴──────────────┴────────────┘
```
eitsupi, line 80: I think the R code and the results are mixed up and not very good.
Could this file be changed to Rmd?

I believe it is possible by adding a target building this file to the Makefile and gitignore this file.

PR #237, opened 2023-06-08
Revert commits pushed to the main branch by accident

eitsupi:
See https://github.com/pola-rs/r-polars/commit/08ce8c45c2a903e375ae0d344e0065e36ee4f2b0#r117064940

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: polars
 Title: Polars ported to R
-Version: 0.6.1.9000
+Version: 0.6.1
 Depends: R (>= 4.1.0)
 Imports: utils, codetools
 Authors@R:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,3 @@
-# polars (development version)
-## What's changed
-
-- `<DataFrame>$unique()` and `<LazyFrame>$unique()` gain a `maintain_order` argument (#237).
-
 # polars 0.6.1
 ## What's changed
  - use `pl$set_polars_options(debug_polars = TRUE)` to profile/debug method-calls of a polars query (#193)
--- a/R/dataframe__frame.R
+++ b/R/dataframe__frame.R
@@ -427,16 +427,11 @@ DataFrame_drop_nulls = function(subset = NULL) {
 #' @title DataFrame_unique
 #' @description Drop duplicate rows from this dataframe.
 #' @keywords DataFrame
-#'
-#' @param subset string or vector of strings. Column name(s) to consider when
-#'  identifying duplicates. If set to NULL (default), use all columns.
+#' @param subset string or vector of strings. Column name(s) to consider when identifying duplicates. If set to NULL (default), use all columns.
 #' @param keep string. Which of the duplicate rows to keep:
 #' * "first": Keep first unique row.
 #' * "last": Keep last unique row.
 #' * "none": Don’t keep duplicate rows.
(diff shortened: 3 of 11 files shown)


etiennebacher:
the problem is that `melt` and some others were already merged in `main` but I forgot to update my `main` before merging which is why I had to fix the conflict. Since the CI was fine with my 2 commits, can't we just let them there?

eitsupi:
> the problem is that `melt` and some others were already merged in `main` but I forgot to update my `main` before merging which is why I had to fix the conflict. Since the CI was fine with my 2 commits, can't we just let them there?

I cannot determine what has been changed because at first glance I do not know what has been changed

I am fine either way, but I believe that submitting it again as a Pull Request is the more reliable way.
If we follow the changes later, we will not really know what happened.

sorhawell:
That resolve conflicts have tricked me before also.

Ok so we remove option to override in branchrules. And I guess if any admin really really need to override some day, then admin must temporaily toggle the branch rule.

etiennebacher:
I will redo a clean PR with these changes, it's not too big anyway

etiennebacher:
But @eitsupi I think reverting this goes back too early, it was already 0.6.1 before my first commit.

eitsupi:
> But @eitsupi I think reverting this goes back too early, it was already 0.6.1 before my first commit.

@etiennebacher I'm sorry but I don't understand what you are talking about. It would be very helpful if you could do this.

sorhawell:
@etiennebacher just point to a PR which is a fair try. Then we can all review it and add any changes. If not exactly the same so be it.

eitsupi:
Would it be better to force push to the main branch?

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: polars
 Title: Polars ported to R
-Version: 0.6.1.9000
+Version: 0.6.0.9000
```
eitsupi, line 3: @etiennebacher I see, perhaps you have a point about here?
I really dont know how to undo this other than force push.

etiennebacher:
Thank you @eitsupi I think the last force-push solved it, this PR now goes back to the state before my last 2 commits on `main`. I think this can be merged and I will make a new PR with my 2 commits

eitsupi:
Ok, I finally could back to 0.6.1 with this command:

```sh
git diff 08ce8c45c2a903e375ae0d344e0065e36ee4f2b0..ddf5510b47ada712ff7838cac74bfbf3f5a226dd >to_fix.diff
mv to_fix.diff /tmp/to_fix.diff
git apply /tmp/to_fix.diff
```

4b897bcd67a36f831b4707fac7a629789c72ddb8 is completely identical to 0.6.1

```sh
git diff 4b897bcd67a36f831b4707fac7a629789c72ddb8..ddf5510b47ada712ff7838cac74bfbf3f5a226dd
```

etiennebacher (approved):

etiennebacher:
FYI I think a simpler solution would have been `git reset ddf5510b47ada712ff7838cac74bfbf3f5a226dd` (last "good" commit)

eitsupi:
Thanks @etiennebacher, It would be sufficient to simply create a PR on the branch where you revert the commit that merged this.

eitsupi:
> FYI I think a simpler solution would have been git reset ddf5510b47ada712ff7838cac74bfbf3f5a226dd (last "good" commit)

My understanding is that `git reset` only moves the tree and does not create a new commit.

PR #238, opened 2023-06-10
Add arg `maintain_order` to `unique()`

etiennebacher:
Follow-up of #237

### DIFF
--- a/tests/testthat/test-dataframe.R
+++ b/tests/testthat/test-dataframe.R
@@ -598,6 +598,21 @@ test_that("unique", {
   expect_false(all(x == y))
 })
 
+test_that("unique, maintain_order", {
+  df = pl$DataFrame(
+    x = rep(c("A", "B", "C"), 2),
+    y = rep(c("A", "B", "C"), 2)
+  )
+  # with maintain_order = FALSE, the order could still be maintained once in
+  # a while due to chance so we repeat this test several times
+  for (i in 1:5) {
+    expect_equal(
+      df$unique(maintain_order = TRUE)$to_data_frame()$x,
+      c("A", "B", "C")
+    )
+  }
+})
+
 test_that("as_data_frame (backward compatibility)", {
   w = as.data.frame(pl$DataFrame(mtcars)$to_data_frame())
   x = as.data.frame(pl$DataFrame(mtcars)$as_data_frame())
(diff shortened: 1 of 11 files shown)


tests/testthat/test-lazy.R
```diff
@@ -247,6 +247,21 @@ test_that("unique", {
...
+    y = rep(c("A", "B", "C"), 2)
+  )
+  # with maintain_order = FALSE, the order could still be maintained once in
+  # a while due to chance so we repeat this test several times
+  for (i in 1:5) {
+    expect_equal(
+      df$lazy()$unique(maintain_order = TRUE)$collect()$to_data_frame()$x,
+      c("A", "B", "C")
+    )
+  }
```
eitsupi, line 262: I don't see the validity of repeating 5 times here.
Is it possible to reduce the chance factor by fixing seed or setting some environment variable?

tests/testthat/test-dataframe.R
```diff
@@ -598,6 +598,21 @@ test_that("unique", {
...
+    y = rep(c("A", "B", "C"), 2)
+  )
+  # with maintain_order = FALSE, the order could still be maintained once in
+  # a while due to chance so we repeat this test several times
+  for (i in 1:5) {
+    expect_equal(
+      df$unique(maintain_order = TRUE)$to_data_frame()$x,
+      c("A", "B", "C")
+    )
+  }
```
eitsupi, line 613: ditto

eitsupi (commented):
Looks good, thanks.

One comment about the tests.

tests/testthat/test-lazy.R
etiennebacher: I don't know how to do that, @sorhawell do you have some idea?

sorhawell: We could test on  a column of e.g. a hundred unique elements, then the risk if randomly getting the order right by chance would be astronomically low `1 / 100! = 1E-157`.

sorhawell:
I will add some minor changes to rust code tonight

tests/testthat/test-lazy.R
etiennebacher: indeed, that was a very easy solution :smile:

sorhawell (approved):

PR #246, opened 2023-06-20
fix release pipeline misses devtools

sorhawell:
fix release pipeline misses devtools and fails

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,7 +53,7 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           extra-packages: any::rcmdcheck
-          needs: check
+          needs: dev
           pak-version: devel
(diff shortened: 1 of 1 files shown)


sorhawell:
add run release on any PR
install dev packages for release build

.github/workflows/release.yaml
```diff
@@ -5,6 +5,8 @@ on:
     branches: [main]
     tags:
       - "**" # Push events to every tag including hierarchical tags like v1.0/beta
+  pull_request:
+    branches: [main]
```
eitsupi, line 9:
```suggestion
```

Since `workflow_dispatch` can be executed by specifying a branch, there is no need to trigger `pull_request` here.
You can test manually with `workflow_dispatch`.

eitsupi (approved):
Thanks. One minor update.

PR #248, opened 2023-06-21
disable docs deployment if not DEPLOY_DOCS key is present in github SECRETS

sorhawell:
Any fork having a branch called main will currently try to deploy docs without the DEPLOY_DOCS key and fail.
That is annoying for any external developer.

This PR aborts deploying docs if not DEPLOY_DOCS key are actively stored in fork repo.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -63,7 +63,7 @@ jobs:
       - uses: webfactory/ssh-agent@v0.6.0
         env: 
           DEPLOY_DOCS: ${{ secrets.DEPLOY_DOCS  }}
-        if: ${{ (github.event_name != 'pull_request') && (env.DEPLOY_DOCS != '') }} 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -61,13 +61,17 @@ jobs:
         run: make docs
 
       - uses: webfactory/ssh-agent@v0.6.0
-        if: github.event_name != 'pull_request'
+        env: 
+          DEPLOY_DOCS: ${{ secrets.DEPLOY_DOCS  }}
+        if: ${{ (github.event_name != 'pull_request') && (env.DEPLOY_DOCS != '') }}
```
eitsupi, line 66: I am concerned that if I put a secret in an environment variable and then use that environment variable elsewhere, that secret will show up. Wouldn't it be better to check the repository owner than that?

eitsupi (approved):

Issue #243, opened 2023-06-13
Allow `$columns` for `LazyFrame`


etiennebacher:
Currently, `pl$DataFrame(mtcars)$lazy()$columns` errors: 
``` r
polars::pl$DataFrame(mtcars)$lazy()$columns
#> Error: syntax error: columns is not a method/attribute of the class LazyFrame 
#>  [TRUNCATED]
```

It is possible in Python Polars: https://pola-rs.github.io/polars/py-polars/html/reference/lazyframe/api/polars.LazyFrame.columns.html#polars.LazyFrame.columns

etiennebacher:
This is now possible thanks to #250 by @Sicheng-Pan

PR #249, opened 2023-06-22
Implement methods to extract schema from LazaFrame

Sicheng-Pan:
This PR aims to implement `schema`, `columns`, `dtypes` for `LazaFrame`, as requested by #243


Sicheng-Pan:
I'm re-opening this in #250 as the commit history in this one is too much spam

PR #250, opened 2023-06-22
Implement LazyFrame$schema

Sicheng-Pan:
This PR aims to implement `schema`, `columns`, and `dtypes` for `LazyFrame`, as requested by #243


Sicheng-Pan:
@sorhawell It seems that the assignment symbol is changed from `=` to `<-` somehow in the wrapper generated by `extendr`. Any ideas how to configure this so that we can keep using `=` for assignment?

sorhawell:
> @sorhawell It seems that the assignment symbol is changed from `=` to `<-` somehow in the wrapper generated by `extendr`. Any ideas how to configure this so that we can keep using `=` for assignment?

my bad, I auto-styled all files in your previous PR extendr-wrappers.R should not be styled.

#251 should revert this, feel free to overrride `=` immediately

sorhawell:
@Sicheng-Pan I recommend that you work and submit PRs from individual feature branches in your fork. I got this advice from the `rlang` maintainer. It will make it easier for you to work on simultanous PRs and a reviewer like me will not mess up your main branch (and all branches thereof) by pushing changes to your PR.

sorhawell:
looks good :) 
I changed to derive columns and dtypes from internal wrapper to show "in $columns" and not "in $schema". It should also be possible to derive from a public method but then something like this pattern should be used

```R
  self$schema |> 
    result() |> 
    unwrap("in $column():") |>
    names()
 ```
 
 however currently then err will have two times the same call-info +  two times a `"in $xxx" ` info.
 
 
 I was contemplating to let RPolarsErr have the type
 
 ```rust
 struct RPolarsErr {
 ctx_chain: VecDeque<Rctx>,
 call: Option<Rctx>,
 wherein: Option<Rctx>
}
```

then any error can only have one call and one wherein. When deriving from another public method, the call and wherein will just be overwritten if an error.

sorhawell:
@vincentarelbundock  this will allow us to bring back names s3 method for LazyFrame without a collect and performance drop :)

sorhawell (approved):
Thanks !

vincentarelbundock:
> @vincentarelbundock this will allow us to bring back names s3 method for LazyFrame without a collect and performance drop :)

This is awesome! I haven't time to do anything about this recently, and I probably won't have more time in the next couple months. Will probably check in then to see what's to do. Sorry...

PR #251, opened 2023-06-22
Chore: revert styler on extender-wrappers.R

sorhawell:
I mistakenly applied styler to auto-generated extendr-wrappers.R in a previous PR.

This reverts all `=` to `<-`.

It would othewise be tedious to use styler to fight the auto-generated styling from extendr wrappers.


eitsupi:
I think this is no longer needed.

PR #252, opened 2023-06-22
point to rpolars/extendr branch pr473_553_555_566

sorhawell:
This PR will adopt the current state of extendr PRs
extendr/extendr#473 - allow returning rust result to R
extendr/extendr#553 - allow renaming extendr struct on R side only
extendr/extendr#555 - fix memory leak
extendr/extendr#566 - escape `%` to avoid trigger formatting with `Rprintf`. #258

r-polars already adopted a previous state of PR473


eitsupi (approved):
Thank you for working to improve extendr!

PR #253, opened 2023-06-22
try use extendr PR553 to rename classes

sorhawell:
RLazyFrame on rust side
PolarsLazyFrame on R side


R/dataframe__frame.R
```diff
@@ -572,15 +572,15 @@ DataFrameCompareToOtherDF = function(self, other, op) {
 
 
 
-#' Convert an existing DataFrame to a LazyFrame
+#' Convert an existing DataFrame to a PolarsLazyFrame
```
etiennebacher, line 575: I guess in docs it would be easier to read `LazyFrame` only

R/lazyframe__lazy.R
```diff
@@ -1,56 +1,56 @@
-#' @title Inner workings of the LazyFrame-class
+#' @title Inner workings of the PolarsLazyFrame-class
 #'
-#' @name LazyFrame_class
-#' @description The `LazyFrame`-class is simply two environments of respectively
+#' @name PolarsLazyFrame_class
```
etiennebacher, line 3: if you change the "name" arg here, then it will change the name of the Rd files and therefore it will also break the "docs" workflow. So either these shouldn't be changed or the script "make-docs.R" should be updated

R/lazyframe__lazy.R
```diff
@@ -1,56 +1,56 @@
...
+#' `DataFrame_object$lazy() -> PolarsLazyFrame_object` and `PolarsLazyFrame_object$collect() -> DataFrame_object`.
 #' This is quite similar to the lazy-collect syntax of the dplyrpackage to
 #' interact with database connections such as SQL variants. Most SQL databases
 #' would be able to perform the same otimizations as polars such Predicate Pushdown
 #' and Projection. However polars can intertact and optimize queries with both
 #' SQL DBs and other data sources such parquet files simultanously. (#TODO
 #' implement r-polars SQL ;).
 #'
-#' @details Check out the source code in R/LazyFrame__lazy.R how public methods
+#' @details Check out the source code in R/PolarsLazyFrame__lazy.R how public methods
```
etiennebacher, line 30: shouldn't be changed

etiennebacher (changes requested):
Thanks for doing this, it will make it much safer for me to make S3 methods for polars objects in `tidypolars`.

However I think there are a lot of changes that are not necessary here (I marked a few in review  comments). 

Is it also possible to do the same for `DataFrame`?

sorhawell:
@etiennebacher I think your suggestions are all correct. Though, I'm not really happy with that e.g. LazyFrame has so many different names
rust-side=RLazyFrame
R-side=PolarsLazyFrame
doc-side=LazyFrame

especially that we call it LazyFrame but it is PolarsLazyFrame ... becomes messy.

I will also try to hide all classes behind a superclass "polars" such that the class would be c("LazyFrame", "polars") and only exporting the polars methods

sorhawell:
I did manage to make prototype with a super class "polars"

e.g. class of the polars DataType named "RPolarsDataType" can be changed to c("DataType", "polars")

then all S3 methods for DataType are made internal to not conflict with arrow

e.g. for `==.` this method is exported
```R
#' @export
"==.polars"  = function(e1, e2) {
  ns_pl[[paste0("==.",head(class(e1),1))]](e1,e2) #"ns_pl" is the polars internal namespace
}
```

it looks up the internal method and finds `==.DataType`. Arrow objects does not conflict because they do not have the `"polars"` superclass.

to really go through with this extendr needs to support not exporting `$.` and `[[.` for all classes and tag all return instances with superclass `"polars"`

all this is probably too experimental/crazy to move forward with :)

etiennebacher:
I don't have enough knowledge of `extendr` for this, but I don't understand why you also need a different name for the Rust side. I think the ideal setting would be to call it a `LazyFrame` in the docs (the readers already know it is from `polars`) and `PolarsLazyFrame` internally in R and Rust to prevent any conflicts with other R/Rust objects. Is this feasible? 

The solution you have in your last post seems quite hacky, I think that the first way to go was better (but again, I don't know how `extendr` limits you in what you want to do)

sorhawell:
I think you're right.

sorhawell:
@Sicheng-Pan uses rust only to make the type on rust side to make the alias `RDF` for rdataframe::DataFrame. Don't know why I did not think about that before. Then it would be possible to rename r-polars extendr types to what ever on R side directly via plain extendr, and then give a type alias on rust side.

This means the renaming for tidy-polars would not need PR553 in extendr I guess

sorhawell:
this PR has become stale, new attempt will be made from fresh PR

PR #90, opened 2023-03-24
DO NOT MERGE: tracking hello_r_universe branch

sorhawell:

Issue #80, opened 2023-03-18
CRAN Release?


eitsupi:
The Rust version of Debian testing will soon be 1.65, which may allow us to build rpolars tuned to build in the R-universe.
(Note that the macOS builder's Rust version may be older than Debian's. eitsupi/prqlr#94)
https://tracker.debian.org/pkg/rustc
https://qa.debian.org/excuses.php?experimental=1&package=rustc

@sorhawell Could you potentially do a CRAN release?
Probably need to modify build scripts (like #29) and documentation (Of course I would like to contribute to that.).

sorhawell:
Cool :)

For mac [this issue](https://github.com/pola-rs/r-polars/issues/35) will trigger a warning or error with CRAN check. I likely have to bundle some shared macOS tool or include as a requirement in DESCRIPTION.

Is it allowed on CRAN, not to release on all OS, e.g. skip mac in first round?

Regarding #29
Using CARGO_HOME is very useful to run CHECK in development without recompiling for 30 min. There is something about an environment variable called something like IS_CRAN right? I could disable CARGO_HOME for cran builds then.
I failed to rediscover the exact envvar name in docs.

eitsupi:
> For mac [this issue](https://github.com/pola-rs/r-polars/issues/35) will trigger a warning or error with CRAN check. I likely have to bundle some shared macOS tool or include as a requirement in DESCRIPTION.

I'm sorry I don't understand the whole issue, but are you saying that due to the issue, rpolars can's update polars?
(Currently, the command check is succeeding on the main branch, right?)

> There is something about an environment variable called something like IS_CRAN right?

Yes, this must be `NOT_CRAN`. (See also extendr/rextendr#233)
Note that this is an environment variable name initiated independently by the `devtools` package, not defined by CRAN.

sorhawell:
r-lib/rcmdchek does not support to filter errors. Many projects get the same warning on large binary size, which CRAN is not enforcing anyways.

Best advice I could find in issue track is to ignore any warnings in R CMD Check to avoid builds flagged as failed.

I found the practice sub-optimal, as I would not notice my own newly introduced warnings.
[...]

PR #262, opened 2023-06-27
feat!: add `simd` feature and remove `polars/simd` from the default feature

eitsupi:
Related to #80 and #90

Enable/disable `simd` by environment variable at build time.
`simd` is disabled by default for CRAN releases.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@ for demonstration purposes (#240).
 - Polars is internally moving away from string errors to a new error-type called `RPolarsErr` both on rust- and R-side. Final error messages should look very similar (#233).
 - LazyFrame_columns, _schema, _dtypes implemented. Improvements to internal `RPolarsErr`. Also `RPolarsErr` will now print each context of the error on a separate line (#250).
 - Add helpful reference landing page at `polars.github.io/reference_home` (#223).
+- rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
+  `RPOLARS_ALL_FEATURES` to `true` when build r-polars (#262).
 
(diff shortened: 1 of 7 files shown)


eitsupi:
After this merged, I want to switch the R-universe source to the main branch.

NEWS.md
```diff
@@ -8,6 +8,8 @@ for demonstration purposes (#240).
 - Polars is internally moving away from string errors to a new error-type called `RPolarsErr` both on rust- and R-side. Final error messages should look very similar (#233).
 - LazyFrame_columns, _schema, _dtypes implemented. Improvements to internal `RPolarsErr`. Also `RPolarsErr` will now print each context of the error on a separate line (#250).
 - Add helpful reference landing page at `polars.github.io/reference_home` (#223, #264).
+- rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
+  `RPOLARS_ALL_FEATURES` to `true` when build r-polars (#262).
```
etiennebacher, line 12: It's a bit unclear to me how this will work in practice. If I'm a classic user, do I have to do anything when I install `r-polars` binaries? Same question for a developer

eitsupi: This is an update on the Rust crate build. If we are installing binary R package, this is irrelevant.

I will update the documentation on this in a follow-up PR as discussed here https://github.com/pola-rs/r-polars/issues/80#issuecomment-1610339106.

eitsupi:
@sorhawell Could you take a look at this?

sorhawell (approved):
awesome thanks :)

eitsupi:
Thank you for your review!

PR #268, opened 2023-06-27
DataFrame.describe + auto-convert String-Err to RPolarsErr

sorhawell:
## Aim
use describe as occasion to rework `RPolarsErr` conversion behavior.

Before string errors and `RPolarsErr` where treated "equal". Now string errors will be upgraded to `RPolarsErr`. String errors can arise from legacy code with naive errors and from any base R function emitting errors.

Prior if an error was String it would propagate as that to the final error to the user.
Now R side `result()` and rust side `unpack_r_result_list`  immediately upgrade errors to `RPolarsErr`.

`RPolarsErr` builds context in a more rigorous way and produces better traces of the error.

## Summary of changes:
 
 -  `>DataFrame>$describe()`
-  internal S3 method `upgrade_err` which can convert R errors and any legacy error into RPolarsErr.
 - now when `result()` captures errors they will be upgraded with upgrade_err()
 - .. also when R result_list unpacked any error will be upgraded to robj_to!
 - replace protoProxyArray with robj_to! in select
 -  rework inner rust `robj_to_rexpr` being called by robj_to!
 -  no long support $select(list_1, list_2) on multiple lists args. Either input is one single arg which is one Expression a list of expressions. OR input is multiple args which are expressions. (this matches py-polars behavior)


tests/testthat/test-dataframe.R
```diff
@@ -927,3 +927,23 @@ test_that("rename", {
   b = df$rename(list(miles_per_gallon = "mpg", horsepower = "hp"))$columns
   expect_identical(a, b)
 })
+
+
+
+test_that("describe", {
+  expect_snapshot(pl$DataFrame(mtcars)$describe()$to_list(),variant = "deparse")
```
etiennebacher, line 934: why do you use `to_list()`? I think the snapshot is easier to read with the default output:
```r
> pl$DataFrame(mtcars)$describe()
shape: (9, 12)
┌────────────┬───────────┬──────────┬────────────┬───┬──────────┬──────────┬──────────┬────────┐
│ describe   ┆ mpg       ┆ cyl      ┆ disp       ┆ … ┆ vs       ┆ am       ┆ gear     ┆ carb   │
│ ---        ┆ ---       ┆ ---      ┆ ---        ┆   ┆ ---      ┆ ---      ┆ ---      ┆ ---    │
│ str        ┆ f64       ┆ f64      ┆ f64        ┆   ┆ f64      ┆ f64      ┆ f64      ┆ f64    │
╞════════════╪═══════════╪══════════╪════════════╪═══╪══════════╪══════════╪══════════╪════════╡
│ count      ┆ 32.0      ┆ 32.0     ┆ 32.0       ┆ … ┆ 32.0     ┆ 32.0     ┆ 32.0     ┆ 32.0   │
│ null_count ┆ 0.0       ┆ 0.0      ┆ 0.0        ┆ … ┆ 0.0      ┆ 0.0      ┆ 0.0      ┆ 0.0    │
│ mean       ┆ 20.090625 ┆ 6.1875   ┆ 230.721875 ┆ … ┆ 0.4375   ┆ 0.40625  ┆ 3.6875   ┆ 2.8125 │
│ std        ┆ 6.026948  ┆ 1.785922 ┆ 123.938694 ┆ … ┆ 0.504016 ┆ 0.498991 ┆ 0.737804 ┆ 1.6152 │
│ min        ┆ 10.4      ┆ 4.0      ┆ 71.1       ┆ … ┆ 0.0      ┆ 0.0      ┆ 3.0      ┆ 1.0    │
│ max        ┆ 33.9      ┆ 8.0      ┆ 472.0      ┆ … ┆ 1.0      ┆ 1.0      ┆ 5.0      ┆ 8.0    │
│ median     ┆ 19.2      ┆ 6.0      ┆ 196.3      ┆ … ┆ 0.0      ┆ 0.0      ┆ 4.0      ┆ 2.0    │
│ 25pct      ┆ 15.5      ┆ 4.0      ┆ 121.0      ┆ … ┆ 0.0      ┆ 0.0      ┆ 3.0      ┆ 2.0    │
│ 75pct      ┆ 22.8      ┆ 8.0      ┆ 350.0      ┆ … ┆ 1.0      ┆ 1.0      ┆ 4.0      ┆ 4.0    │
└────────────┴───────────┴──────────┴────────────┴───┴──────────┴──────────┴──────────┴────────┘
```

sorhawell: I got this error with that the "roundtrip" conversion failed. I guess it was because I used `expect_snapshot_value` which does support polars objects. I will take your advice and use expect_snapshot and capture the printed form instead :)

eitsupi (commented):
Please auto formatting (`make fmt GIT_DIF_TARGET=main`)

sorhawell:
> make fmt GIT_DIF_TARGET=main

thx I misunderstood how to use make fmt

eitsupi:
> thx I misunderstood how to use make fmt

Yes, currently by default only uncommitted files are formatted, so you need to set `GIT_DIF_TARGET=main` to format after a commit.......

sorhawell:
hmm maybe bad merge, seems like I'm reverting reverse -> descending PR

eitsupi:
> hmm maybe bad merge, seems like I'm reverting reverse -> descending PR

I think this is due to https://github.com/pola-rs/r-polars/pull/291#issuecomment-1617974362

I will open a follow up PR...

eitsupi (approved):
Thanks!

ddotta:
Awesome !

PR #271, opened 2023-06-28
feat: add `polars_info` function

### LINKED ISSUE #269: `polars_info()` function
After #262 merged, it would be useful to show info about features etc.

The arrow package has `arrow_info` function.
https://arrow.apache.org/docs/r/reference/arrow_info.html
https://github.com/apache/arrow/blob/6f3bd2524c2abe3a4a278fc1c62fc5c49b56cab3/r/R/arrow-info.R#L31-L74

eitsupi:
Based on #262, close #269

Implement a function like `arrow::arrow_info()`

```r
> polars_info()
R Polars package version: 0.6.1.9000

Features:          
simd FALSE
```

It provides fairly limited information for now, but other information can be added in the future.


sorhawell:
good idea

the py-polars equivalent

```py
>>> pl.show_versions()
--------Version info---------
Polars:      0.18.4
Index type:  UInt32
Platform:    macOS-12.6.6-x86_64-i386-64bit 
Python:      3.10.9 (main, Dec 15 2022, 18:18:30) [Clang 14.0.0 (clang-1400.0.29.202)]

----Optional dependencies---- 
numpy:       1.24.3
pandas:      <not installed>
pyarrow:     12.0.0
connectorx:  <not installed>
deltalake:   <not installed>
fsspec:      <not installed>
matplotlib:  <not installed>
xlsx2csv:    <not installed>
xlsxwriter:  <not installed>
>>> 
```

R/info.R
```diff
@@ -0,0 +1,32 @@
...
+  print_key_values = function(title, vals, ...) {
+    df = data.frame(vals, ...)
+    names(df) = ""
+
+    cat(title, ":", sep = "")
+    print(df)
+    cat("\n")
+  }
+
+  cat("R Polars package version: ", format(x$version), "\n\n", sep = "")
```
etiennebacher, line 30: is it possible to have the rust-polars version as well? (and does it make sense to display it?)

etiennebacher (commented):
LGTM, can you add a snapshot test?

R/info.R
eitsupi: Good point.
But we uses unreleased version of rust-polars here, so I think showing rust-polars' version make little sense.

eitsupi:
> can you add a snapshot test?

I was hesitant to add a snapshot test, given that results can vary depending on build conditions.
But perhaps by forcing the values to be overwritten, only print results can be tested...

R/info.R
eitsupi: Upon further investigation, it seems that there is no constant in rust-polars to display the version.
So I think we need to update rust-polars first to do that.

NEWS.md
```diff
@@ -11,6 +11,7 @@ for demonstration purposes (#240).
 - Add helpful reference landing page at `polars.github.io/reference_home` (#223, #264).
 - rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
   `RPOLARS_ALL_FEATURES` to `true` when build r-polars (#262).
+  A new function `polars_info()` will tell which features enabled (#271).
```
etiennebacher, line 14:
```suggestion
- A new function `polars_info()` will tell which features enabled (#271).
```

tests/testthat/test-info.R
```diff
@@ -0,0 +1,17 @@
+patrick::with_parameters_test_that("polars_info()features are logical",
```
etiennebacher, line 1:
```suggestion
patrick::with_parameters_test_that("polars_info() features are logical",
```

etiennebacher (approved):
Thank you @eitsupi

eitsupi:
Sorry, I may have made a mistake on something.......
On the website after this PR is merged, `simd` shows `FALSE` as follows.
This should be `TRUE`.......

```r
polars_info()
#> R Polars package version: 0.6.1.9000
#> 
#> Features:          
#> simd FALSE
#> 
```

eitsupi:
Now I understand why. I had misunderstood the use of target_feature.
https://doc.rust-lang.org/reference/attributes/codegen.html#the-target_feature-attribute

I should use `feature`, not `target_feature`...

R/info.R
eitsupi: I have opened an issue about that. pola-rs/polars#9659

PR #274, opened 2023-06-29
ci: R CMD check without `simd` feature with release virsion of Rust

### LINKED ISSUE #263: Workflows without the `simd` feature
After #262 is merged, R CMD check, MSRV check, etc. should be done without the `simd` feature.

eitsupi:
Close #263

It also update the Makefile to always build with `--all-features`.


.github/workflows/check.yaml
```diff
@@ -12,38 +12,39 @@ concurrency:
...
           - { os: windows-latest, r: "devel", rtools-version: "43" }
           - { os: windows-latest, r: "release", rtools-version: "" }
           - { os: windows-latest, r: "oldrel-1", rtools-version: "42" }
           - { os: ubuntu-latest, r: "devel", http-user-agent: "release" }
           - { os: ubuntu-latest, r: "release" }
           - { os: ubuntu-latest, r: "oldrel-1" }
+        include:
+          - config: { os: macos-latest, r: "release", all-features: true }
+          - config: { os: windows-latest, r: "release", all-features: true }
+          - config: { os: ubuntu-latest, r: "release", all-features: true }
```
eitsupi, line 33: @sorhawell I would like to remove R-oldrel on Windows and R-devel on Windows because the matrix is getting huge.
What do you think?

eitsupi:
Currently the build is failing, but I think we need to merge it as is and fix the error later.

.github/workflows/check.yaml
sorhawell: can we keep, Windows R-devel for early warning? [I turned it of previously and regretted that decision  :)](https://github.com/pola-rs/r-polars/blob/5fe587b2492ceb54bb14355040c1a33de106b0b1/src/Makevars#L25)

I don't mind turning off oldrel

eitsupi: Makes sense. I have removed R-oldrel on Windows.

eitsupi:
I think we can merge this right now.
It is unfortunate that the larger matrix will reduce the number of tests that can be run concurrently, but I think it will get us a lot closer to a CRAN release.

eitsupi:
@sorhawell @etiennebacher Could you merge this so we can move forward on #280?

etiennebacher:
FYI I can't really review PRs that are mostly about Rust (either code or infrastructure) because I don't know enough to see if there are mistakes or not (for ex in #271)

etiennebacher (approved):

eitsupi:
Thanks for merging!

Issue #276, opened 2023-06-29
Can't build with Rust stable toolchain


eitsupi:
The changes made by #233 prevent builds except for nightly-toolchain.
https://github.com/pola-rs/r-polars/actions/runs/5411224946/jobs/9833719506?pr=274

Can we revert this change for now? Or Could you fix that error?

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/issues/80#issuecomment-1612903299_

sorhawell:
@Sicheng-Pan @eitsupi 
Sorry that is a blunder from my side, I overlooked PR used an unstable feature.

I think there is tag to forbid the compiler to use unstable features.

I will make a dumb hotfix for now.

@Sicheng-Pan  maybe you can find another way to bubble the various errors.

Sicheng-Pan:
@sorhawell Yeah I should have avoided using nightly features. I checked your PR and it looks good to me.

eitsupi:
Thanks for taking a look at this right away!
I think it is definitely necessary to test the build on CI with the stable toolchains. (#274)

PR #278, opened 2023-06-29
trait specialization replace with two plain funcs

sorhawell:
hotfix #276


sorhawell:
@eitsupi  if no errors this hotfix should solve #276

Sicheng-Pan (approved):
So we are using some custom functions to convert Extendr and Polars errors into ours. I believe this approach is good enough for now as currently there are not too many places where we need to do the manual conversion.

eitsupi:
Thanks both!

Issue #208, opened 2023-05-10
R-universe r-polars build fails because rust-polars is no longer backwards compatible with rustc 1.65.0 stable


sorhawell:
R-universe build fails
https://github.com/r-universe/rpolars/actions/runs/4884403053

~~I tried to downgrade to rust 1.65 stable locally, but I got no errors so far.~~
~~Something with simd is mentioned in some of the compiler errrors, but not sure what it is about.~~

EDIT: In a clean build I was able to reproduce the error by downgrading to rustc 1.65 stable

sorhawell:
the issue is upstream rust-polars has [adopted crate argminmax 0.6.1](https://github.com/pola-rs/polars/pull/8074) which shows to no bet backwards compatible with rustc 1.65.0 stable but onlye 1.66 stable. The minimal previous minimal rustc version was 1.62 for polars.

argminmax 0.6.1 currently needs rustc >= 1.66.0 to no  give the [following compiler error](https://github.com/r-universe/rpolars/actions/runs/4938598307/jobs/8828571811)


possible solutions:
 - ask R universe to upgrade to rustc 1.66.0  (problematic because a CRAN release would then be several more years away)
 - kindly ask polars to feature gate argminmax crate
 - kindly ask argminmax to become backwards compatible with 1.65 stable
 - for rust-polars where argminmax was never adopted (could work in the short run, would probably to too difficult to maintain over time)

sorhawell:
[error reproduced in hello_r_universe](https://github.com/pola-rs/r-polars/actions/runs/4941854934/jobs/8835133766)

eitsupi:
Thanks for taking a look at this.

Perhaps Debian testing will skip Rust 1.64 and 1.65, so this problem will be solved when we can build this package in CRAN machine anyway.
https://tracker.debian.org/pkg/rustc

> we ourselves should probably add an "R-universe-like" build to our regular PR checks to discover such backwards compatibility issues earlier.

I too thought that the handling of SIMD disabled should be tested more.

If someone wants to work on it, one way is to rewrite Cargo.toml by detecting the available Rust state by the `configure` script like the `gifski` package, I think.
https://github.com/r-rust/gifski/blob/68911d927fb34c43dede24ce1baab73346e8bf04/configure#L11-L22

sorhawell:
> Perhaps Debian testing will skip Rust 1.64 and 1.65, so this problem will be solved when we can build this package in CRAN machine anyway.

[...]

PR #279, opened 2023-06-30
ci: add MSRV check workflow (should support Rust 1.66 now)

### LINKED ISSUE #275: MSRV check on CI
Related to #263

eitsupi:
Close #275

### DIFF
--- a/.github/workflows/msrv.yaml
+++ b/.github/workflows/msrv.yaml
@@ -1,0 +1,31 @@
+name: Check MSRV
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - src/rust/**
+  pull_request:
+    branches:
+      - main
+    paths:
+      - src/rust/**
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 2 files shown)


sorhawell (approved):

sorhawell:
so adding `rust-version = 1.66` in top of yaml means MSRV, and if we fic #208 we lower it to 1.65?

sorhawell:
@eitsupi  if this test PR works we can set a profile release-optimized and connect it to the `RPOLARS_ALL_FEATURES` envvar. 

new behaviour would be:
`RPOLARS_ALL_FEATURES`  = `true`:  will compile also argminmax crate with opt-level = 3 and with "polars/simd".
`RPOLARS_ALL_FEATURES`  != `true`:  will compile also argminmax crate with opt-level = 1 and without "polars/simd".

eitsupi:
Thanks for taking a look at this. I will merge this and make it easy to see how upcoming changes will affect MSRV.

PR #280, opened 2023-06-30
fix: support Rust 1.65

### LINKED ISSUE #208: R-universe r-polars build fails because rust-polars is no longer backwards compatible with rustc 1.65.0 stable

sorhawell:
Fix #208


sorhawell:
@eitsupi  not sure whylinker cant find rpolars.a, but it does compile on 1.65.0 on my machine if `RPOLARS_ALL_FEATURES != 'true`

eitsupi:
Nice! It seems passing the MSRV check.

sorhawell:
> Nice! It seems passing the MSRV check.

changing cargo-toml from 1.66 to 1.65 now it should be more impressive :)

eitsupi:
> changing cargo-toml from 1.66 to 1.65 now it should be more impressive :)

Thanks, it seems passing!

eitsupi:
> The output for each profile will be placed in a directory of the same name as the profile in the [target directory](https://doc.rust-lang.org/cargo/guide/build-cache.html). As in the example above, the output would go into the target/release-lto directory.

https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles

eitsupi:
Given the need to make significant changes to Makevars if adding `profile.release-optimized`, I suspect that simply setting `profile.release.package.argminmax` is sufficient.

eitsupi (approved):
I think it can be merged now.

eitsupi:
@sorhawell Can I merge this and go ahead?

sorhawell:
Ohh profile changes output dir, that was what I was missing . Thanks @eitsupi  , looks good!

Issue #258, opened 2023-06-23
prefix + `%` crashes R


sorhawell:
passing `%` as char to prefix will crash the rsession

```R
library(polars)
df = pl$DataFrame(mtcars)
df3 = do.call(df$select,list(pl$all()$quantile(.5)$prefix("hi%")))
print(df3)
```

sorhawell:
maybe somewhere between rust and extendr the `%` is cursed 

add debug print to prefix
```rust
    fn prefix(&self, prefix: Robj) -> RResult<Self> {
        let s = robj_to!(String, prefix)?;
        rprintln!("bytes{:?} \n", s.as_bytes());
        s.chars().into_iter().for_each(|c| {
            //hey
            rprintln!(" c[{:?}-{}]", c, c);
        });
        Ok(self.0.clone().prefix(s.as_str()).into())
    }

```

it seems rust dbg fmt cannot display % properly however other exotic chars works fine

```R
> pl$col()$prefix("%")
bytes[37] 

 c['%]
$ok
polars Expr: RENAME_ALIAS col("")

$err
NULL

attr(,"class")
[1] "extendr_result"
> pl$col()$prefix("◊")
bytes[226, 151, 138] 

 c['◊'-◊]
$ok
polars Expr: RENAME_ALIAS col("")

$err
NULL

attr(,"class")
[1] "extendr_result"
```

sorhawell:
fixed when #252  merges

sorhawell:
the final fix is found [here in extendr](https://github.com/extendr/extendr/pull/566)

etiennebacher:
All linked PRs are merged and the behavior now looks correct so I'm closing this:
```r
> library(polars)
>   df = pl$DataFrame(mtcars)
>   df3 = do.call(df$select,list(pl$all()$quantile(.5)$prefix("hi%")))
>   df3
shape: (1, 11)
┌────────┬────────┬─────────┬───────┬───┬───────┬───────┬─────────┬─────────┐
│ hi%mpg ┆ hi%cyl ┆ hi%disp ┆ hi%hp ┆ … ┆ hi%vs ┆ hi%am ┆ hi%gear ┆ hi%carb │
│ ---    ┆ ---    ┆ ---     ┆ ---   ┆   ┆ ---   ┆ ---   ┆ ---     ┆ ---     │
[...]

PR #284, opened 2023-06-30
test extendr rprintln supports `%`-char

sorhawell:
Related to #258

### DIFF
--- a/R/extendr-wrappers.R
+++ b/R/extendr-wrappers.R
@@ -57,6 +57,8 @@ test_robj_to_i64 <- function(robj) .Call(wrap__test_robj_to_i64, robj)
 
 test_robj_to_u32 <- function(robj) .Call(wrap__test_robj_to_u32, robj)
 
+test_print_string <- function(s) invisible(.Call(wrap__test_print_string, s))
+
 RPolarsErr <- new.env(parent = emptyenv())
 
 RPolarsErr$new <- function() .Call(wrap__RPolarsErr__new)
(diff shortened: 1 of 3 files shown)


eitsupi:
Thanks, but wouldn't it be enough to do the following test on the R side without changing the Rust code?
https://github.com/pola-rs/r-polars/issues/258#issuecomment-1614522160

```r
> library(polars)
>   df = pl$DataFrame(mtcars)
>   df3 = do.call(df$select,list(pl$all()$quantile(.5)$prefix("hi%")))
>   df3
shape: (1, 11)
┌────────┬────────┬─────────┬───────┬───┬───────┬───────┬─────────┬─────────┐
│ hi%mpg ┆ hi%cyl ┆ hi%disp ┆ hi%hp ┆ … ┆ hi%vs ┆ hi%am ┆ hi%gear ┆ hi%carb │
│ ---    ┆ ---    ┆ ---     ┆ ---   ┆   ┆ ---   ┆ ---   ┆ ---     ┆ ---     │
│ f64    ┆ f64    ┆ f64     ┆ f64   ┆   ┆ f64   ┆ f64   ┆ f64     ┆ f64     │
╞════════╪════════╪═════════╪═══════╪═══╪═══════╪═══════╪═════════╪═════════╡
│ 19.2   ┆ 6.0    ┆ 225.0   ┆ 123.0 ┆ … ┆ 0.0   ┆ 0.0   ┆ 4.0     ┆ 2.0     │
└────────┴────────┴─────────┴───────┴───┴───────┴───────┴─────────┴─────────┘
```

sorhawell:
That one will lead to segfault if it does not work. Either way we will learn :)

eitsupi:
> That one will lead to segfault if it does not work. Either way we will learn :)

Oh, you mean you want to add a test that does not cause segmentation fault?
That would certainly be worthwhile. (Not sure if that's common, though)

eitsupi (approved):

PR #285, opened 2023-06-30
fix: `polars_info()` shows features correctly

eitsupi:
https://github.com/pola-rs/r-polars/pull/271#issuecomment-1614721193

We should use `feature` here, not `target_feature`.

Now it works fine.

```r
> polars_info()
R Polars package version: 0.6.1.9000

Features:         
simd TRUE
```


etiennebacher (approved):

etiennebacher:
Thanks!

Issue #230, opened 2023-06-02
rename subnamespace `Expr.arr` to `.list`


sorhawell:
https://github.com/pola-rs/polars/pull/8999

etiennebacher:
This has to be done at some point but it would bloat the code to keep both `arr` and `list` namespaces. I suppose we can replace all `arr` by `list` at once, which will be a massive breaking change. That could be worth a new release (maybe `0.8.1`?). 

What do you think @sorhawell @eitsupi ?

sorhawell:
The subnamespaces can be moved around relatively easy. I can start a  community PR with the namespace rerouting

Maybe the complete change could be done in two PRs

Patch update to 0.8.1 or so
 -  reroute arr subnamespace from Expr$arr to also Expr$list (~5 lines of r code)
 -  update docs to recommend list over arr + deprecation warning on arr (many search replace, but hopefully fast)
 
Minor update to 0.9 maybe
 - close Expr$arr access
 -  rename internally to remove mentions of arr

etiennebacher:
Sounds good to me. I think we could release 0.8.1 right after your "rerouting" PR and then take more time until 0.9.0 so that the deprecation notice is visible to the users at least for a few weeks

etiennebacher:
`arr` is now deprecated and will be removed in 0.9.0

PR #289, opened 2023-07-02
feat!: bump rust-polars to 0.30.0

eitsupi:
https://github.com/pola-rs/polars/releases/tag/rs-0.29.0
https://github.com/pola-rs/polars/releases/tag/rs-0.30.0


eitsupi:
The errors in Rust have been fixed but have not yet passed the R tests.
Feel free to edit this branch as I may not have enough time to work on this.

etiennebacher:
I think this should close #230

eitsupi:
I fixed as many of the test failures as I could.

The remaining failures are a problem where `date_range_lazy` now returns a list, so when converted to R, it becomes a list instead of a vector. (pola-rs/polars#8513)

sorhawell:
Thanks. I will take a look at it.

eitsupi:
I was just looking at this myself and Python's `date_range` behaves quite strangely.
I mean, when `eager=True` it returns an un-nested Series, but when `eager=False` it returns a list type, which I think is more appropriate here given the meaning of range, but anyway it returns something completely different for `True` and `Flase`.
Some tests must be removed if this is to be mimicked.

eitsupi:
> I found polars did a behavior change on date_range such that it produces Eager = False a list of DateRanges.
> It is being discussed here (pola-rs/polars#9019)

Oh, thank you for finding that.

The change definitely makes sense, but we cannot incorporate that change here because it has not yet been made in the main branch.
Perhaps we can change the `lazy=FALSE` behavior in this repository using `implode`? (This is different from the behavior of py-polars, but since it will certainly be changed in py-polars in the future, it does not seem to be a problem to change it ahead of time.)

sorhawell:
One timezone error remaining and the we have restore previous behavior.

We can update the date_range(s) behavior in a new PR after this has merged

sorhawell:
tiny detail for reference "[fix subtle typo bug](https://github.com/pola-rs/r-polars/pull/289/commits/28c1b7e46c87420e562b840c3d94d5b2ff469e97)" is an incorrect commit msg. There was a unrelated subtle bug (high/low) which did not show, and then I disabled a test on timezone until a future refactor.

eitsupi:
All tests now pass; I would like to defer the renaming from `list` to `arr` on the R side to a follow-up fix as it would be an additional major task.

NEWS.md
```diff
@@ -4,6 +4,12 @@
 
 - Replace the argument `reverse` by `descending` in all sorting functions. This
   is for consistency with the upstream Polars (#291, #293).
+- Bump rust-polars from 2023-04-20 unreleased version to version 0.30.0 released in 2023-05-30 (#289).
+    - Rename `concat_lst` to `concat_list`.
+    <!-- TODO: - Rename `arr` to `list`. -->
+    - Rename `$str$explode` to `$str$str_explode`.
+    - Remove `tz_aware` and `utc` arguments from `str_parse`.
+    - `$date_range`'s the `lazy` argument is default to `TRUE`.
```
etiennebacher, line 12:
```suggestion
    - in `$date_range`'s the `lazy` argument is now `TRUE` by default.
```

R/expr__string.R
```diff
@@ -706,20 +700,20 @@ ExprStr_slice = function(offset, length = NULL) {
     unwrap("in str$slice:")
 }
 
-#' explode
-#' @name ExprStr_explode
+#' str_explode
```
etiennebacher, line 703: Would be better to have a proper title here (not just for this function)

etiennebacher (approved):
Looks great, thanks for updating this. It's fine for me to do all the renaming from `arr` to `list` in a separate PR, are you also ok with this @sorhawell ?

eitsupi:
@etiennebacher Thanks for your review.

Since this PR contains updates to the toolchain and everything needs to be recompiled every time we switch branches, I think it's better to merge it in sooner rather than later, so I'm merging it.
Additional modifications can be made later.......

PR #292, opened 2023-07-03
fix(installation): set `CARGO_HOME` to a temp dir if not `NOT_CRAN=true`

### LINKED ISSUE #29: Set CARGO_HOME envvar in Makevars
This is mandatory if you want to release this package on CRAN eventually. If not, you don't need to care.

https://github.com/r-rust/faq#how-to-avoid-writing-in-home

eitsupi:
Close #29

If not `NOT_CRAN=true`, set `CARGO_HOME` to a temporary directory to suppress writing to the home directory.

ref: extendr/rextendr#233

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -12,12 +12,22 @@ all: C_clean
 
 $(SHLIB): $(STATLIB)
 
+CARGOTMP = $(CURDIR)/.cargo
+
 $(STATLIB):
+	if [ "$(NOT_CRAN)" != "true" ]; then \
+		export CARGO_HOME=$(CARGOTMP); \
+	fi && \
+	export PATH="$(PATH):$(HOME)/.cargo/bin" && \
 	if [ "$(RPOLARS_ALL_FEATURES)" == "true" ]; then \
 		$(BUILD_COMMAND) --all-features; \
 	else \
 		$(BUILD_COMMAND); \
-	fi;
+	fi
+	if [ "$(NOT_CRAN)" != "true" ]; then \
+		rm -Rf $(CARGOTMP) && \
+		rm -Rf $(LIBDIR)/build; \
+	fi
 	if [ -f "$(STATLIB)" ]; then \
 		echo "file is there: "; \
 	else \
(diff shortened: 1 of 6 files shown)


src/Makevars
```diff
@@ -5,19 +5,29 @@ LIBDIR = ./rust/target/$(RPOLARS_PROFILE)
...
 
 all: C_clean
 
 $(SHLIB): $(STATLIB)
 
+CARGOTMP = $(CURDIR)/.cargo
+
 $(STATLIB):
+	if [ "$(NOT_CRAN)" != "true" ]; then \
+		export CARGO_HOME=$(CARGOTMP); \
```
etiennebacher, line 19: > If `NOT_CRAN=true`, set CARGO_HOME to a temporary directory to suppress writing to the home directory.

Shouldn't it be `"$(NOT_CRAN)" == "true"` above (and same below)? sorry if this is trivial, I'm not used to makefiles

etiennebacher: by the way I find it harder to read when we have double negative, like `NOT_CRAN = false`. Is it possible to rename `NOT_CRAN` to `CRAN`?

eitsupi: Thanks for taking a look at this, the PR description was incorrect! (I have fixed)

The intent of this is to set `CARGO_HOME` on CRAN, so if not `NOT_CRAN=ture`, it is considered it is on CRAN.

etiennebacher: thanks for the explanation, re my second comment, `"$(NOT_CRAN)" != "true"` -> `"$(CRAN)" == "false"` would be much clearer

sorhawell (approved):
nice, looks good to me

src/Makevars
eitsupi: `NOT_CRAN` is an environment variable set by packages such as `devtools` and used by the community to indicate that it is outside of CRAN.
In other words, on CRAN, no environment variable is set to indicate that they are on CRAN.

eitsupi: Here is a note about the origin of `NOT_CRAN`.
https://github.com/extendr/rextendr/pull/233#issuecomment-1475259856

PR #291, opened 2023-07-02
Replace `reverse` by `descending` in sorting functions

etiennebacher:
Close #51

PR #293, opened 2023-07-03
fix: reverse -> descending

eitsupi:
A follow up for #291

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@ for demonstration purposes (#240).
 - `<DataFrame>$glimpse()` is a fast `str()`-like view of a `DataFrame` (#277).
 - `$over()` now accepts a vector of column names (#287).
 - Replace the argument `reverse` by `descending` in all sorting functions. This
-  is for consistency with the upstream Polars (#290, #293).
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -17,7 +17,7 @@ for demonstration purposes (#240).
 - `<DataFrame>$glimpse()` is a fast `str()`-like view of a `DataFrame` (#277).
 - `$over()` now accepts a vector of column names (#287).
 - Replace the argument `reverse` by `descending` in all sorting functions. This
-is for consistency with the upstream Polars (#290).
+  is for consistency with the upstream Polars (#290, #293).
```
etiennebacher, line 20:
```suggestion
  is for consistency with the upstream Polars (#291, #293).
```

etiennebacher:
Thanks, I don't understand how #291 passed the tests with these changes missing...

eitsupi:
> Thanks, I don't understand how #291 passed the tests with these changes missing...

I think this is because it does not actually use the names of the arguments on the Rust side, but only the order of the arguments.

etiennebacher (approved):
thanks!

PR #294, opened 2023-07-03
docs(README): update installation options

### LINKED ISSUE #145: long_arms64 branch?
The README says the following, is this still correct today?

The `long_arms64` branch has not been updated for a long time. (last commit was 2 months ago 15501904b49aaa52eadfe349fc42df08af206e4d)

````md
One exception worth noting is MacOS (arm64), i.e. systems based on the
new M1/M2 “Silicon” chips. To install **polars** on one of these
machines, we need to build the package from source and this requires
[Xcode](https://developer.apple.com/xcode/) (`xcode-select --install`).
Once Xcode is installed, you can run the below code chunk to build
**polars**. The corresponding `Makevars` script will download a \~200MB
cross-compiled object file, while your machine links and builds the
final R package.

``` r
# install.packages("remotes")

remotes::install_github(
  "https://github.com/pola-rs/r-polars",
  ref = "long_arms64", force =TRUE
)
```
````

### LINKED ISSUE #286: Docs about build options
- #262 starts using feature flags and controls features with environment variable `RPOLARS_ALL_FEATURES`.
- #280 allows changing profiles via the environment variable `RPOLARS_PROFILE`.

These should be noted in the documentation for the installation.

### PR DESCRIPTION
**Pull request by eitsupi**
Close #145, close #286

Updated information on installation.

### CODE REVIEW

File: README.Rmd [hunk 1]
```diff
@@ -51,20 +51,26 @@ options for a variety of operating systems:
...
+
+Binary packages on R-universe are compiled by stable Rust, with nightly features disabled.
 
 ```r
 install.packages("polars", repos = "https://rpolars.r-universe.dev")
 ```
 
+```r
+# For Ubuntu binary installation
+install.packages("polars", repos = "https://rpolars.r-universe.dev/bin/linux/jammy/4.3")
```
etiennebacher, line 65: is the ubuntu binary also available for 4.2? if so, we should mention it

README.Rmd
```diff
@@ -74,49 +80,37 @@ install the latest release of **polars** on Linux (x86_64) one would use:
...
-### r2u
-Speeding up your workflow? On Ubuntu, install polars + arrow from binaries and resolve
-system dependencies reliably and quickly with r2u ([see link for configuration](https://eddelbuettel.github.io/r2u/)).
-```r
-rp = c("https://rpolars.r-universe.dev/bin/linux/jammy/4.3", "https://cloud.r-project.org")
-install.packages(c("polars", "arrow"), repos = rp)
+```{r, include = FALSE}
+rust_toolchain_version = brio::read_file("Makefile") |>
+  stringr::str_extract(r"((?<=RUST_TOOLCHAIN_VERSION).*nightly.*(?=\n))") |>
+  stringr::str_extract(r"(nightly.*)")
```
etiennebacher, line 106: this returns `NA` for me but it seems to work here so let's keep it like that

etiennebacher (approved):
thanks!

README.Rmd
eitsupi: Thank you for taking a look at this. It is odd......
It is a simple process so I omitted the process to check if the value is valid, but we might want to add it.

eitsupi: @etiennebacher Is it likely that you have not disabled the automatic conversion between CRLF and LF by Git for Windows?

etiennebacher: it's probably the cause, if I use `\r` instead of `\n` in the regex of the second line then it works

eitsupi: I have added .gitattributes to ensure `lf` for eol.

eitsupi: Thank you for looking into this, it is really annoying that the automatic conversion between CRLF and LF is still the default setting in Git for Windows.....

etiennebacher: TBH I very rarely have an issue with this so it's at the bottom of the list of Windows specificities that annoy me 😄

README.Rmd
```diff
@@ -74,49 +80,37 @@ install the latest release of **polars** on Linux (x86_64) one would use:
...
 Similarly for Windows
-([URL](https://github.com/pola-rs/r-polars/releases/latest/download/polars.zip)) 
+([URL](https://github.com/pola-rs/r-polars/releases/latest/download/polars.zip))
 and MacOS (x86_64,
 [URL](https://github.com/pola-rs/r-polars/releases/latest/download/polars__x86_64-apple-darwin20.tgz)).
 Just remember to invoke the `repos = NULL` argument if you are installing these
 binary builds directly from within R.
 
-One exception worth noting is MacOS (arm64), i.e. systems based on the new M1/M2
-"Silicon" chips. To install **polars** on one of these machines, we need to
```
etiennebacher, line 89: Since you removed this paragraph, does it mean that this PR close https://github.com/pola-rs/r-polars/issues/145?

eitsupi: Indeed.

Obviously that is not being maintained, so I think it would be better to have a source install.
Also, a successful CRAN release will host the arm64 macOS binary package on CRAN.

etiennebacher: ok, I updated the first comment of the PR

README.Rmd
```diff
@@ -74,49 +80,37 @@ install the latest release of **polars** on Linux (x86_64) one would use:
...
-Please [file an issue](https://github.com/pola-rs/r-polars/issues) if you
-require a different target or operating system build. Finally, see the bottom of
-this README for details on how to install rust to build from source (only
-relevant for developers, or users of unsupported operating systems).
+For source installation,
+the Rust toolchain (Rust `r RcppTOML::parseTOML("src/rust/Cargo.toml")$package$"rust-version"` or later) must be configured.
 
+Please check the <https://github.com/r-rust/hellorust> repository for about Rust code in R packages.
 
-### r2u
```
sorhawell, line 112: r2u was removed here, issues tracked in #378

PR #296, opened 2023-07-03
docs(NEWS): polish NEWS

eitsupi:
Includes a fix to undo extendr-wrappers changes that was incorrectly updated by #268.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,12 +1,13 @@
 # polars (development version)
+
 ## What's changed
 
 - New function `rename()` for `LazyFrame` and `DataFrame` (#239)
 - `<DataFrame>$unique()` and `<LazyFrame>$unique()` gain a `maintain_order` argument (#238).
-- New `pl$LazyFrame()` to quickly create a `LazyFrame`, mostly in examples or 
+- New `pl$LazyFrame()` to quickly create a `LazyFrame`, mostly in examples or
 for demonstration purposes (#240).
 - Polars is internally moving away from string errors to a new error-type called `RPolarsErr` both on rust- and R-side. Final error messages should look very similar (#233).
-- LazyFrame_columns, _schema, _dtypes implemented. Improvements to internal `RPolarsErr`. Also `RPolarsErr` will now print each context of the error on a separate line (#250).
+- LazyFrame_columns, _schema,_dtypes implemented. Improvements to internal `RPolarsErr`. Also `RPolarsErr` will now print each context of the error on a separate line (#250).
 - Fix memory leak on error bug. Fix printing of `%` bug. Prepare for renaming of polars classes (#252).
 - Add helpful reference landing page at `polars.github.io/reference_home` (#223, #264).
 - rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,40 +1,52 @@
...
 - Polars is internally moving away from string errors to a new error-type called `RPolarsErr` both on rust- and R-side. Final error messages should look very similar (#233).
-- LazyFrame_columns, _schema, _dtypes implemented. Improvements to internal `RPolarsErr`. Also `RPolarsErr` will now print each context of the error on a separate line (#250).
+- `$columns()`, `$schema()`, `$dtypes()` for `LazyFrame` implemented (#250).
+- Improvements to internal `RPolarsErr`. Also `RPolarsErr` will now print each context of the error on a separate line (#250).
 - Fix memory leak on error bug. Fix printing of `%` bug. Prepare for renaming of polars classes (#252).
 - Add helpful reference landing page at `polars.github.io/reference_home` (#223, #264).
-- rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
-  `RPOLARS_ALL_FEATURES` to `true` when build r-polars (#262).
+- Supports Rust 1.65 (#262, #280)
+  - rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
```
etiennebacher, line 20:
```suggestion
- rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable
```

etiennebacher (approved):
Thanks for cleaning this. I think that after this version is released, we should split the news in several categories like new functions or methods, bug fixes, docs, etc.

NEWS.md
eitsupi: Thanks for taking a look at this.

The reason we disabled `simd` by default was to support Rust 1.65, so I think it makes sense to make it a child item.
What do you think?

etiennebacher: ah ok, but since it's the only one I think it makes sense to put it in the same sentence, like 
```
- Supports Rust 1.65. This means that rust-polars' `simd` feature is now disabled by default. To enable it, set the environment variable... (#262, #280)
```

eitsupi: That makes sense.

But the addition of the new profile `release-optimized` is also related to this, so I will add a new child item.

NEWS.md
```diff
@@ -1,40 +1,55 @@
...
+    `RPOLARS_ALL_FEATURES` to `true` when build r-polars (#262).
+  - `opt-level` of `argminmax` is now set to `1` in the `release` profile for to support Rust < 1.66.
+    The profile can be changed by setting the environment variable `RPOLARS_PROFILE` and when set it to `release-optimized`,
+    `opt-level` of `argminmax` is set to `3`.
```
etiennebacher, line 24:
```suggestion
  - `opt-level` of `argminmax` is now set to `1` in the `release` profile to support Rust < 1.66.
    The profile can be changed by setting the environment variable `RPOLARS_PROFILE` (when set to `release-optimized`,
    `opt-level` of `argminmax` is set to `3`).
```

eitsupi:
Interestingly, MkDocs material does not seem to support the level of listing?

In https://rpolars.github.io/NEWS/

image

In https://github.com/pola-rs/r-polars/blob/main/NEWS.md

image

eitsupi:
> It might be needed to use a different symbol for each list level: [squidfunk.github.io/mkdocs-material/reference/lists/?h=list#usage](https://squidfunk.github.io/mkdocs-material/reference/lists/?h=list#usage)

I see, thanks for letting me know!

etiennebacher:
Note that I'm just guessing, I'll check locally

PR #298, opened 2023-07-03
Fix news indentation

etiennebacher:
Cf #296

@eitsupi nothing to do with symbols, just not enough indentation


eitsupi (approved):
Thanks!

eitsupi:
Perhaps `NEWS.md` is handled by pandoc via altdoc?
I noticed that the indentation size is not reflected and the `<` is escaped.

image

etiennebacher:
Both issues (indenting and escaping) are due to a bad reformatting by `altdoc:::reformat_md()`, I'll take a look

PR #301, opened 2023-07-04
Square bracket for LazyFrame

vincentarelbundock:
@eitsupi ,

I uncommented the tests and the bracket method. Unfortunately, I can't seem to build the package on my local machine right now, and I won't have any computer time in the next two weeks, so I won't be able to do anything else on this front.

Hopefully Github actions shows us something useful now...


vincentarelbundock:
Looks like we may still be missing a `width` method: 

```sh
[1] "/Users/runner/work/r-polars/r-polars"
      ▆
   1. ├─df[, c("mpg", "hp")] at test-s3_methods.R:205:2
   2. └─polars:::`[.LazyFrame`(df, , c("mpg", "hp")) at test-s3_methods.R:205:2
   3.   ├─x$width at polars/R/s3_methods.R:64:2
   4.   └─polars:::`$.LazyFrame`(x, "width") at polars/R/s3_methods.R:64:2
   5.     └─polars:::verify_method_call(LazyFrame, name) at check/polars.Rcheck/polars:2:2
```

eitsupi:
Thanks for taking a quick look at this.

Perhaps we can keep this in draft form for now.

etiennebacher:
Yes methods `$width()`, `$height()` and `$shape()` are missing for `LazyFrame`

sorhawell:
I suppose `width` can be derived from `schema` something like

`LazyFrame_width = function() length(self$schema)`

 But `height` and `shape` are not known before collecting.

vincentarelbundock:
I added a few s3 methods with NA as suggested.

I think the brackets should *not* allow subsetting of rows for LazyFrame, because a lot of our sanity checks rely on knowing the exact number of rows. So my latest commit returns an informative error in that case.

There are also some tests.

I won't be able to do any more changes to this branch, so I will ask for review now. Feel free to merge or make changes as you deem appropriate.

eitsupi:
It seems lack of test cases yet.

```r
> df = pl$DataFrame(mtcars)$lazy()

> df[, "hp"]
Error: syntax error: to_series is not a method/attribute of the class LazyFrame 
 when calling method:
 df[, "hp"]
```

etiennebacher (approved):
Looks great, thanks. Can you add a NEWS item? We'll need to wait for #308 I guess?

eitsupi:
@etiennebacher Thanks for your review!

I think this is a small but valuable feature addition, so it would be nice if it could be merged into the main branch and from there into #308.
Is it okay to add features when resubmitting? (I have very little experience submitting packages to CRAN)

If you determine that this is not a problem I would be happy to merge this in.

eitsupi (approved):
Thanks @vincentarelbundock!

etiennebacher:
I think it's fine to add this before resubmitting, but we'll have to make sure not to forget these files when addressing the comments in #308

PR #303, opened 2023-07-04
Use `altdoc` dev version

etiennebacher:
Cf #298 

@eitsupi is this enough to use the dev version of `altdoc`?


eitsupi (approved):
Thanks!
CI for docs seems woking fine, so it is ready to merge.

Issue #297, opened 2023-07-03
Collaborative list before submitting to CRAN


etiennebacher:
@eitsupi @sorhawell just in case you don't know this list: https://github.com/ThinkR-open/prepare-for-cran

(in particular there will be a lot of typos to check with `spelling::spell_check_package()`)

PR #304, opened 2023-07-04
docs: fix some misspelling

eitsupi:
Related to #297

I ran `spelling::spell_check_package()` and fixed some typos.


etiennebacher (approved):
@eitsupi I don't know if you want to fix all typos here or do it gradually. I'm fine if you want to merge

DESCRIPTION
```diff
@@ -28,6 +29,7 @@ Suggests:
     patrick,
     pillar,
     rmarkdown,
+    spelling,
```
eitsupi, line 32: Is this needed here?
I believe `Config/Needs/dev` is more appropriate as it will only be used during development.

eitsupi (commented):
Thank you for taking a look at this!
I did everything I could and am happy to merge this at this point.

DESCRIPTION
etiennebacher: indeed, that was an accidental commit

Issue #267, opened 2023-06-27
only read a few columns from a csv


tdhock:
Hi @sorhawell 
I see you can use `n_rows` argument of  `read_csv_` to limit number of rows, but is there some way to limit which columns are read? I would like to do something like the following,
```r
>      my_file = tempfile()
> write.csv(iris, my_file, row.names=FALSE)
> data.table::fread(my_file, nrows = 2, select=1:2)
   Sepal.Length Sepal.Width
1:          5.1         3.5
2:          4.9         3.0
```
I looked at the man page, ?polars::csv_reader, and the closest mention I found was the following, but I don't see a dtype/string corresponding to "ignore this column" so I am posting an issue.
```
overwrite_dtype: (NULL is disable) named list of dtypes OR dtype-names,
          where name points to a column. Can overwrite dtypes during
          inference. Supported types so far are: name | alias | polars
          side dtype "Boolean" | "logical" => DataType::Boolean,
          "Float32" | "double" => DataType::Float32, "Float64" |
          "float64" => DataType::Float64, "Int32" | "integer" =>
          DataType::Int32, "Int64" | "integer64" => DataType::Int64,
          "Utf8" | "character" => DataType::Utf8,
```

sorhawell:
hi @tdhock  :)

fast answer, it is possible to scan a data source into a LazyFrame and e.g. `select()` and/or `filter()` when calling `collect()`the optimized plan will push down projections and predicates to the internal csv_reader.

the `csv_reader`- function was some of the first I wrote in this projection, it is likely due to be updated.

```R
> my_file = tempfile()
> write.csv(iris, my_file, row.names=FALSE)
> 
> lf = pl$lazy_csv_reader(
+   my_file
+ )$filter(
+   pl$col("Species") == "setosa"  
+ )$select(
+   "Sepal.Length", "Sepal.Width"  
+ )
> 
> lf$describe_optimized_plan()
FAST_PROJECT: [Sepal.Length, Sepal.Width]

    CSV SCAN /var/folders/v1/b2c26lpn2yjd997jg_gn4fgc0000gn/T//Rtmp6fCkkc/file527052b4a418
    PROJECT 3/5 COLUMNS
    SELECTION: [(col("Species")) == (Utf8(setosa))]
> 
> lf$collect()
shape: (50, 2)
┌──────────────┬─────────────┐
│ Sepal.Length ┆ Sepal.Width │
│ ---          ┆ ---         │
│ f64          ┆ f64         │
[...]

PR #305, opened 2023-07-04
Rename functions to read CSV

etiennebacher:
Mentioned in https://github.com/pola-rs/r-polars/issues/267#issuecomment-1611369883

This PR:

* `lazy_csv_reader` -> `scan_csv`
* `csv_reader` -> `read_csv`

For now, they only can be accessed with `pl$` because I removed `read_csv_`, but if we still want to have this kind of high-level helper, we should rename it (maybe `read_csv_pl()`?)

Feel free to push directly here


eitsupi:
Thank you for working on this!

I thought we should do this too, but when I looked at the py-polars implementation I realized that Python's `polars.scan_csv` is completely different from the current R's `lazy_csv_reader`.

That is, Python's `scan_csv` seems to map directly to Rust's `scan_csv`.
https://github.com/pola-rs/polars/blob/d40e403c81620f659244af20ce36470af8afe419/polars/polars-lazy/polars-plan/src/logical_plan/builder.rs#L206

sorhawell:
[actually py-polars uses also LazyCsvReader](https://github.com/pola-rs/polars/blob/1d06aa7f2f1fa7dc4d01f4aa403bc70653c42a3b/py-polars/src/lazyframe.rs#L153) via [here](https://github.com/pola-rs/polars/blob/1d06aa7f2f1fa7dc4d01f4aa403bc70653c42a3b/py-polars/polars/lazyframe/frame.py#L302) and via [here](https://github.com/pola-rs/polars/blob/1d06aa7f2f1fa7dc4d01f4aa403bc70653c42a3b/py-polars/polars/io/csv/functions.py#L680)

I have no idea what the difference between `scan_csv` and `LazyCsvReader`

R/info.R
```diff
@@ -1,10 +1,10 @@
 #' Report information of the package
 #'
 #' @return A list with information of the package
+#' @name polars_info
 #' @examples
 #' polars_info()
-#' @export
-polars_info = function() {
+pl$polars_info = function() {
```
eitsupi, line 7: Could you also update the examples?

eitsupi (approved):
Thank you for the updates! I think it is a good change for consistency.

Perhaps we can merge this since it seems to have passed the test?

eitsupi:
> [actually py-polars uses also LazyCsvReader](https://github.com/pola-rs/polars/blob/1d06aa7f2f1fa7dc4d01f4aa403bc70653c42a3b/py-polars/src/lazyframe.rs#L153) via [here](https://github.com/pola-rs/polars/blob/1d06aa7f2f1fa7dc4d01f4aa403bc70653c42a3b/py-polars/polars/lazyframe/frame.py#L302) and via [here](https://github.com/pola-rs/polars/blob/1d06aa7f2f1fa7dc4d01f4aa403bc70653c42a3b/py-polars/polars/io/csv/functions.py#L680)

Oh, sorry. I must have misread the code.......

sorhawell (approved):

PR #309, opened 2023-07-04
docs: add LICENSE.note about Rust dependencies

eitsupi:
Sorry, I forgot that we needed to add this before submitting to CRAN.

Since CRAN policy instructs us to put the license regarding the source code included in the binary in the LICENSE.note, so adds that via the `rextendr`'s function and `cargo-license`.

See extendr/rextendr#236 for details about LICENSE.note.


sorhawell (approved):
many thanks for remembering this!

PR #310, opened 2023-07-04
Enable `how = "cross"` in `$join()`

etiennebacher:
For some reason, `how = "cross"` was manually blocked but it seems to work fine now.


sorhawell (approved):
looks good to me :)

PR #311, opened 2023-07-05
Background execution and R process pool

Sicheng-Pan:
This PR aims to implement an interface to create asynchronous jobs that can be executed in background threads.

Specifically, this will provide an alternative option to run the `Lazy` query in background without blocking the R console.

In order to run `map` and `apply` on the `LazyFrame` without using the existing R interpreter, new R processes have to be created in background to handle the evaluation tasks.

The user will get thread handles for the background jobs, and can check if they are done (non-blocking) or wait for them to finish (blocking).

Example:
```R
> handle <- pl$LazyFrame(mtcars)$with_column(pl$col("mpg")$map_in_background(function(x) x * 0.43)$alias("kml"))$collect_in_background()
> handle$is_finished() |> unwrap()
[1] TRUE
> handle$join() |> unwrap()
shape: (32, 12)
┌──────┬─────┬───────┬───────┬───┬─────┬──────┬──────┬───────┐
│ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ … ┆ am  ┆ gear ┆ carb ┆ kml   │
│ ---  ┆ --- ┆ ---   ┆ ---   ┆   ┆ --- ┆ ---  ┆ ---  ┆ ---   │
│ f64  ┆ f64 ┆ f64   ┆ f64   ┆   ┆ f64 ┆ f64  ┆ f64  ┆ f64   │
╞══════╪═════╪═══════╪═══════╪═══╪═════╪══════╪══════╪═══════╡
│ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ … ┆ 1.0 ┆ 4.0  ┆ 4.0  ┆ 9.03  │
│ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ … ┆ 1.0 ┆ 4.0  ┆ 4.0  ┆ 9.03  │
│ 22.8 ┆ 4.0 ┆ 108.0 ┆ 93.0  ┆ … ┆ 1.0 ┆ 4.0  ┆ 1.0  ┆ 9.804 │
│ 21.4 ┆ 6.0 ┆ 258.0 ┆ 110.0 ┆ … ┆ 0.0 ┆ 3.0  ┆ 1.0  ┆ 9.202 │
│ …    ┆ …   ┆ …     ┆ …     ┆ … ┆ …   ┆ …    ┆ …    ┆ …     │
│ 15.8 ┆ 8.0 ┆ 351.0 ┆ 264.0 ┆ … ┆ 1.0 ┆ 5.0  ┆ 4.0  ┆ 6.794 │
│ 19.7 ┆ 6.0 ┆ 145.0 ┆ 175.0 ┆ … ┆ 1.0 ┆ 5.0  ┆ 6.0  ┆ 8.471 │
│ 15.0 ┆ 8.0 ┆ 301.0 ┆ 335.0 ┆ … ┆ 1.0 ┆ 5.0  ┆ 8.0  ┆ 6.45  │
│ 21.4 ┆ 4.0 ┆ 121.0 ┆ 109.0 ┆ … ┆ 1.0 ┆ 4.0  ┆ 2.0  ┆ 9.202 │
└──────┴─────┴───────┴───────┴───┴─────┴──────┴──────┴───────┘
```


sorhawell:
hi @Sicheng-Pan this PR is very interesting ! :)

I'd like to hear your thoughts on the final syntax, would do you think of below?

I think `$collect_in_background()` is a meaningful variation of `$collect()` as it does not halt the Rsession at the expense of at least one more R process is launched if any R code need to be "mapped" (`$map()`,`$apply()`, ...). Also `$collect_in_background()` returns a handle, not a `<DataFrame>`.

however for `$map_in_background()` I suggest to use regular `$map()` and resolve if map is pushed to another R process via if `$collect()` or `$collect_in_background()` was used.

also it could be advantageous with an option like  `pl$set_polars_options(r_extra_process_pool = 5)`, where default is `0` . If user sets this process are spun up and recycled for more jobs.

eitsupi:
Looks amazing!
Thanks for working on this!

Could you please resolve merge conflicts?

Sicheng-Pan:
@sorhawell Since currently we do not have a global options store on the rust side, I think it is better to leave `map` and `map_in_background` as separate expressions and directly leave the choice to the users. Besides, `map` and `map_in_background` do have some differences: `map` can have side-effects in the main process, while `map_in_background` cannot have side-effects and can only support pure R functions for now.

sorhawell:
I propose 3 last thing before merging:
 - unit tests. Show all features work. Does not have to be large data sizes.
 - benchmark script in `./inst/misc/` one example with little data and perhaps 1000 maps. And one with bigger data and perhaps only a single map. From results write in docs, when user is recommended to use parallel R processes.
 - apply, add support for apply also. Actually very similar to map. Used for mapping over groups.
 
Doesn't that make sense ?

If you like I can add one of these.

sorhawell:
in this PR `LazyFrame` has collect() + collect_in_background()  + collect_background() . That is confusing.
I think there should only be collect() and at max one extra.

In best scenario the user should just call $collect() and everything is settled via options, args and defaults.

sorhawell:
the `RThreadHandle` print.RThreadHandle and is_finished() does not work after handle is exhausted. I have added some expectations to unit-test but not resolved them

sorhawell:
I have added two scenarios "low io - low cpu" and "low io - high cpu".

 in "low io - high cpu"  scenario, R background processes mapping is double as fast. However it does not matter what `capacity` is set to and it appears only two processes are spawned.

```R
> f_all_cols(lf)$collect() |> system.time()
   user  system elapsed 
  7.023   0.148   7.136 
> pl$set_global_rpool_cap(1)
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()  #burn-in start processes
   user  system elapsed 
  0.013   0.010   3.935 
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.014   0.012   3.940 
> pl$set_global_rpool_cap(4)
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time() #burn-in start processes
   user  system elapsed 
  0.014   0.016   3.955 
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.013   0.011   3.642 
> pl$get_global_rpool_cap()
$available
[1] 2

$capacity
[1] 4
```


In "low io - low cpu" where `in_background = FALSE` is much faster as expected. Sometimes when running with `in_background = TRUE` many time in row the performance just becomes 10x faster and match single process. I cannot explain that. Maybe it skips running in background process.
 

```R
> pl$set_global_rpool_cap(4)
> f_sum_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.006   0.006   0.330 
> f_sum_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.005   0.007   0.022 
> f_sum_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.007   0.007   0.025 
 ```

**Comment by Sicheng-Pan** [comment]
@sorhawell It seems that the R-CMD-check is failing on CI. The tests can pass locally on my side. May I ask if you know any possible reasons behind this?

**Comment by sorhawell** [comment]
@Sicheng-Pan 
```R
Error: Error in eval(ei, envir) : checking tests ... [18s/18s] ERROR
  Running ‘testthat.R’ [18s/18s]
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  <RPolarsErr_error/error/condition>
     0: In R: in $collect():
     0: During function call [test_check("polars")]
     1: When calling $collect() on LazyFrame
     2: ComputeError(ErrString("Execution halted with the following contexts\n   1: When trying to rent a R process from the global R process pool\n   2: When trying to spawn a background R process\n   3: Os { code: 8, kind: Uncategorized, message: \"Exec format error\" }\n"))
  
```
It seems there is a OS specific issue with the system call to start a process on ubuntu/linux. We do not see it on a mac.
The rust crate you use to make system calls have likely produced this error `Os { code: 8, kind: Uncategorized, message: \"Exec format error\" }\n"))
  `
  You may find clues there for how to read it. Maybe some calls are not supported on all platforms. Maybe others are.
  
  If you find no clues you I can try debug it in a ubuntu docker container (rocker docker). I have one ready installed for another project.
  
  
  In the latest build you see an error from printing  RThreadHandle
  
  ```R
    Running 'testthat.R' [25s]
Running the tests in 'tests/testthat.R' failed.
Last 13 lines of output:
     0: During function call [test_check("polars")]
  
  Backtrace:
      ▆
   1. ├─base::print(handle) at test-rbackground.R:43:2
   2. └─polars:::print.RThreadHandle(handle)
   3.   ├─base::cat(as.character(x), "\n") at polars/R/rbackground.R:19:22
   4.   ├─base::as.character(x) at polars/R/rbackground.R:19:22
   5.   └─polars:::as.character.RThreadHandle(x) at polars/R/rbackground.R:19:22
   6.     └─polars:::unwrap(.pr$RThreadHandle$thread_description(x)) at polars/R/rbackground.R:14:29
```

I caused that by adding the unit test expectation. I could fix the error but I left it undone to mess to much with your often fine design choices :)

**Comment by Sicheng-Pan** [comment]
@sorhawell I examined the error messages but I still don't have any clue why this could fail. It seems that `Command::spawn` failed on Ubuntu and `thread::spawn` panicked on Windows, but I could not understand why, as these commands should have cross-platform support. If you know what's going on feel free to fix it (or point it out in the comments so that I can fix it)!

**Comment by sorhawell** [comment]
I will try to test the commits between last pass and first error to pin point the issue. Maybe it is share mem.

**Comment by sorhawell** [comment]
[maybe something with quotes ](https://doc.rust-lang.org/std/process/struct.Command.html#method.arg)

**Comment by Sicheng-Pan** [comment]
> @Sicheng-Pan I got a little closer. When trying to run only `R` it fails when running `ls` it does not fail. But likely hangs as there is no handler.
> 
> [29e146f](https://github.com/pola-rs/r-polars/commit/29e146ff48e3b9dba8a565f3a017a1fcc9986ed3)
> 
> I would take step back and start a new branch from main where you unit-test starting processes with R. Maybe another way of staring processes is needed to work on all platforms.
> 
> ... or just stay in this branch but disable all current process unit tests and then slowly add on more and more complicated process calls and see the green check for each commit.

@sorhawell Are you sure about whether `R` command is present in the CI environment? Could you please try `which R` and see the output?

**Comment by sorhawell** [comment]
@Sicheng-Pan  you might be right which points to `r_check_bin/R` which [is a develish little shell script](https://github.com/mlammens/spThin/blob/master/spThin.Rcheck/R_check_bin/R)

echo "'R' should not be used without a path -- see par. 1.6 of the manual"
exit 1

we need to find some cross platform way for R to locate it self. Maybe there is some variable in R it self specifing this. Then we need to start a command with path to the right binary in rust.

**Comment by sorhawell** [comment]
Simple solution would be to have an environment variable where is R binary. If rust cannot start a process, the error suggests the user to specify path to R binary.

**Comment by Sicheng-Pan** [comment]
@sorhawell I've made some small modifications to the code. Now all checks can pass on ubuntu/macos, but it seems that there are new errors on the windows platforms.

**Comment by sorhawell** [comment]
@Sicheng-Pan  to fix the windows issues you could revert to previous way for just windows, something like...

```rust
if R!("R.version$os"=="Windows").expect("in any R session").as_bool().expect("== produceds a bool") == true {
  command::new("R")
} else {
 ....
}.spawn()
```

Sicheng-Pan:
> @Sicheng-Pan  to fix the windows issues you could revert to previous way for just windows, something like...
> 
> 
> 
> ```rust
> 
> if R!("R.version$os"=="Windows").expect("in any R session").as_bool().expect("== producess a bool") == true {
> 
>   command::new("R")
> 
> } else {
> 
>  ....
> 
> }.spawn()
> 
> ```

@sorhawell We probably couldn't use anything from the R side because this function may not be called in the main thread

R/expr__expr.R
```diff
@@ -681,6 +687,9 @@ Expr_map = function(f, output_type = NULL, agg_list = FALSE, in_background = FAL
 #'  investigate.
 #' @name Expr_apply_in_background
 Expr_apply_in_background = function(f, output_type = NULL) {
+  if (Sys.getenv("RPOLARS_R_BINARY_PATH") == "") {
```
sorhawell, line 690: [envvars are not thread safe](https://stackoverflow.com/questions/30476600/how-can-getenv-be-thread-safe). We would probably be fine in most cases though.

I have read py-polars have similar envvar where it is discouraged to modify it after start-up (.onLoad()). If we really needed a thread safe abstraction we could use the `Storage`-crate as with the global [like here in `lib.rs`](https://github.com/pola-rs/r-polars/blob/5d4a59bd6e05f59ffc1ad9759a0c022caeffcdb9/src/rust/src/lib.rs#L36).

I think we might be a bit better of moving this Default setting to [.onLoad](https://github.com/pola-rs/r-polars/blob/5d4a59bd6e05f59ffc1ad9759a0c022caeffcdb9/R/zzz.R#L119)

We get mod this

Sicheng-Pan: Why does this needs to be thread safe? From my understanding the only scenario where this function could be called is when users are constructing expressions in the single-threaded R session.

sorhawell:
It seems `in_background=TRUE` is equally 2x faster than (`in_background`=FALSE) in "low io - high cpu" for 0,1 and 2 processes. Having 8 processes is faster than 2. Could there be some edge condition?

last benchmark is updated [97b2fb9](https://github.com/pola-rs/r-polars/pull/311/commits/97b2fb958502b94e2b215dda183ad2d6c3a230e6)
```R
### 3a -----------  Use R processes in parallel, low io, high cpu
lf <- pl$LazyFrame(lapply(1:12,\(i) rep(i,5)))
f_all_cols <-  \(lf,...) lf$select(pl$all()$map(\(x) {
  for(i in 1:10000) y = sum(rnorm(1000))
  sum(x)
},...))

> pl$set_global_rpool_cap(1)
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time() #burn-in start processes
   user  system elapsed 
  0.007   0.015   4.124 
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.006   0.014   3.871 
> f_all_cols(lf, in_background = FALSE)$collect() |> system.time()
   user  system elapsed 
  8.209   0.350   8.362 
> 
> ##Appears to be a factor overhead when comparing single process
> pl$set_global_rpool_cap(0)
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time() #burn-in start processes
   user  system elapsed 
  0.007   0.015   4.314 
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.007   0.015   4.491 
> f_all_cols(lf, in_background = FALSE)$collect() |> system.time()
   user  system elapsed 
  8.815   0.956   9.725 
> 
> pl$set_global_rpool_cap(2)
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time() #burn-in start processes
   user  system elapsed 
  0.006   0.015   3.989 
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.023   0.038   3.803 
> f_all_cols(lf, in_background = FALSE)$collect() |> system.time()
   user  system elapsed 
  8.943   1.151   9.886 
> 
> pl$set_global_rpool_cap(8)
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time() #burn-in start processes
   user  system elapsed 
  0.005   0.010   3.048 
> f_all_cols(lf, in_background = TRUE)$collect() |> system.time()
   user  system elapsed 
  0.006   0.015   2.425 
> f_all_cols(lf, in_background = FALSE)$collect() |> system.time()
   user  system elapsed 
  8.261   0.317   8.436 
> pl$get_global_rpool_cap() #only 2 processes appears to be spawned
$available
[1] 8

$capacity
[1] 8
```

R/expr__expr.R
sorhawell: Depends on how easy it is to copy the `Static Lazy` pattern. It just in principle cleanest to have as few "invariants" as possible.Some point in the future some careless dev may refactor your code and then it is good to have as few "not causing havoc right now"-elements as possible.

Currently the thread "message container" struct `ParRobj` only works with plain R vectors, not R functions across. There is some type checking, but maybe some altRep, might make it blow in the future. The function  `map_alias()` do contain invariant it should be run from the main thread. I just missed some PyO3 features to implement this cleanly thread-safe. I had an experimental warning for a long time in docs.

So if not worth I also suggest to move the code pieces  to `.onLoad.` I imagine this is a R session related setting, not a query specific setting. In the case a user serializes the query to disk and loads it from another machine it would not work. Maybe you have other scenarios in mind?

R/lazyframe__lazy.R
```diff
@@ -270,6 +270,27 @@ LazyFrame_collect_background = function() {
...
+#' @keywords LazyFrame DataFrame_new
+#' @return a thread handle for the task
+#' @examples
+#' prexpr <- pl$col("mpg")$map(\(x) {Sys.sleep(1.5);x * 0.43}, in_background = TRUE)$alias("kml")
+#' handle = pl$LazyFrame(mtcars)$with_column(prexpr)$collect_in_background()
+#' if(!handle$is_finished()) print("not done yet")
+#' df = handle$join() #get result
+#' df
+#' handle$is_finished()
+#' handle$join()
```
sorhawell, line 289: it seems the informative text from thiserror does not come through to the user
```R
0: During function call [handle$join()]
1: Handled
```
It should state `Joined on a used thread handler`. A casual R user may not understand what just `Handled` means.

R/lazyframe__lazy.R
```diff
@@ -270,6 +270,27 @@ LazyFrame_collect_background = function() {
   .pr$LazyFrame$collect_background(self)
 }
 
+#' @title Collect_in_background
+#' @description collect DataFrame by lazy query in a background thread.
+#' @details
+#' Useful to not block the R session while query executes. Any use of `$map()` or `$apply()` is
+#' diverted to other R sessions. If R functions rely on global variables this can change the
```
sorhawell, line 277: I just added this to docs "Any use of $map() and $apply() will be diverted to other R session" That was not entirely correct. Error messages when trying to use map and apply with `collect_in_background` are very poor.

it seems map(in_background=TRUE) does not work with `collect_inbackground`

```R
pl$set_global_rpool_cap(1)

lf = pl$LazyFrame(iris)

#no informative error, improve
h = lf$groupby("Species")$agg(
  pl$col("Sepal.Length")$apply(\(s) s$to_r() |> length())
)$collect_in_background()

#poor error message, improve.
h$join()

#this works
h = lf$groupby("Species")$agg(
  pl$col("Sepal.Length")$apply_in_background(\(s) s$to_r() |> length())
)$collect_in_background()
h$join()

#this does not work, why not ?
h = pl$LazyFrame(iris)$select(pl$col("Species")$map(\(x) length(x)),in_background =TRUE)$collect_in_background()

```

sorhawell (changes requested):

R/expr__expr.R
```diff
@@ -663,11 +664,34 @@ construct_ProtoExprArray = function(...) {
...
+    if (Sys.getenv("RPOLARS_R_BINARY_PATH") == "") {
+      Sys.setenv(RPOLARS_R_BINARY_PATH = file.path(R.home("bin"), "R"))
+    }
+    map_fn = .pr$Expr$map_in_background
+  }
+  map_fn(self, f, output_type, agg_list)
 }
 
-
+#' Expr_apply_in_background
```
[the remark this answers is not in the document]
Sicheng-Pan: I referred `map_in_background` when I implemented `apply_in_background`, and I'm unsure if `apply` and `apply_in_background` have the same behavior now. Currently `apply_in_background` do not share the same parameters as `apply` in R.

R/lazyframe__lazy.R
Sicheng-Pan: Maybe we should remove `collect_inbackgrounnd` to avoid confusion

README.Rmd
```diff
@@ -279,16 +279,17 @@ cd r-polars
...
 
   ```r
-  devtools::check(
-    env_vars = list(RPOLARS_RUST_SOURCE="/YOUR/OWN/ABSOLUTE/PATH/r-polars/src/rust"),
-    check_dir = "./check/"
-    )
-  source("./inst/misc/filter_rcmdcheck.R")
-  Sys.sleep(5)
-  unlink("check",recursive = TRUE, force =TRUE)
+  source("inst/misc/load_polars.R")
```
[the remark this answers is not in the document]
Sicheng-Pan: Updated

README.Rmd
```diff
@@ -279,16 +279,17 @@ cd r-polars
...
+  
+  #to rextendr:document() + not_cran + load packages + all_features
+  load_polars()
+  
+  #to check package + reuses previous compilation in check, protects against deletion
+  check_polars("/YOUR/OWN/ABSOLUTE/PATH/r-polars/src/rust")
```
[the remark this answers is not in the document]
Sicheng-Pan: Updated

README.md
```diff
@@ -343,16 +343,17 @@ repo.
...
 
 ``` r
-devtools::check(
-  env_vars = list(RPOLARS_RUST_SOURCE="/YOUR/OWN/ABSOLUTE/PATH/r-polars/src/rust"),
-  check_dir = "./check/"
-  )
-source("./inst/misc/filter_rcmdcheck.R")
-Sys.sleep(5)
-unlink("check",recursive = TRUE, force =TRUE)
+source("inst/misc/load_polars.R")
```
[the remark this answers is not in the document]
Sicheng-Pan: Updated

R/lazyframe__lazy.R
sorhawell: I don't think this has been resolved. Using collect_in_background gives to me surprising behavior and not clear error messages (see above)

R/lazyframe__lazy.R
sorhawell: @Sicheng-Pan I have resolved this now, use of Rctx::handled().into() was problematic because Rctx is casted to dyn Error and then back to Rctx.plain().

R/expr__expr.R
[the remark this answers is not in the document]
sorhawell: wouldn't it be possible to just reconcile the args with a few lines ?

sorhawell (changes requested):

Sicheng-Pan:
> It seems `in_background=TRUE` is equally 2x faster than (`in_background`=FALSE) in "low io - high cpu" for 0,1 and 2 processes. Having 8 processes is faster than 2. Could there be some edge condition?
> 
> last benchmark is updated [97b2fb9](https://github.com/pola-rs/r-polars/pull/311/commits/97b2fb958502b94e2b215dda183ad2d6c3a230e6)

@sorhawell I have not examined the `polars` code base closely but I would assume that they are running the `map`s in parallel threads (for the benchmark case). Without using the `R` background process pool, all these treads have to share the main `R` thread from the user, which could become the bottleneck of computation. With the `R` background thread pool each thread can get its own `R` process for evaluation. A process pool of size `8` might be faster than a process pool of size `2` because with larger process it is more likely for a `polars` thread to inherit a used `R` process (after the completion of calculation in another thread), rather than creating a new one, but this may only lead to marginal benefits since the overhead of creating a new process might be small with respect to the computation under the `low io - high cpu` scenario.

R/lazyframe__lazy.R
Sicheng-Pan: The final statement is broken because you have wrong parenthesis. The `in_background` argument is not passed to `map`.

Sicheng-Pan: I've improved the error handling in background:
``` r
> th <- pl$LazyFrame(mtcars)$select(pl$col("mpg")$map(\(x) y, in_background = TRUE))$collect_in_background()
> th$join()
   0: During function call [th$join()]
   1: When trying to map a polars series with R function in the background R process
   2: EvalError(lang!(function (x) y, ExternalPtr.set_class(["Series"]))
```

Sicheng-Pan:
@sorhawell It seems that the docs check CI keeps hanging. Do you have any ideas why this is the case?

sorhawell:
> @sorhawell It seems that the docs check CI keeps hanging. Do you have any ideas why this is the case?

I have a theory it is likely because all the other workflows will build the package whereas [docs workflow will only load it in the session.](https://github.com/pola-rs/r-polars/blob/c24eb71adceef431521c6f6af9acd0cbc6901bc3/.github/workflows/docs.yaml#L65)

I will try to modify the workflow.

Sicheng-Pan:
@sorhawell Thanks! Anything else I should address for this PR?

src/rust/src/rbackground.rs
```diff
@@ -0,0 +1,476 @@
...
+impl RBackgroundPool {
+    pub fn new(cap: usize) -> Self {
+        RBackgroundPool {
+            pool: Arc::new(Mutex::new(VecDeque::new())),
+            cap: Arc::new(Mutex::new(cap)),
+        }
+    }
+
+    pub fn lease(&self) -> RResult<RBackgroundHandler> {
+        let mut sleep_time_us: u64 = 200;
```
sorhawell, line 284: @Sicheng-Pan  new tentative lease method

Sicheng-Pan:
> @Sicheng-Pan I have to run for today I just flush all changes and continue tmr.
> 
> I have found a bug in `RBackgroundPool` that explains the strange benchmark. It is good to dive deep into benchmarks of multi stuff. Sometimes there is gold to be found. `RBackgroundPool::lease` would always spawn a new process if non was available in pool, disregarding the cap. Processes would be discarded if capacity was exceeded still. This explains poor performance in low io - low cpu tests, because most time was spend starting and stopping R processes.
> 
> I made a poorman sleep mechnism where thread will only spawn new process if capacity is not exceeded. Otherwise they will perform some adaptive sleeping between 0.2-16.3ms to wait for R process to become available. A better mechanism where threads are parked and qued until an R process idle in pool.
> 
> Sometime in benchmark test 3a I even get performances which 100 times faster. I think there is one more bug to be found and the the overhead in low io - low cpu will be near zero.

@sorhawell Thanks for the investigation! Actually this is not a bug. I implemented it in this way on purpose, such that the `RBackgroundPool` behaves like a buffer (which can overflow) instead of a container (which can't). Imagine the case where you have heavy `R` computations that can be done in parallel in the background but the user only specified that the pool capacity is `1`. If no new R processes can be spawn then this will become a bottleneck. Meanwhile the overhead of creating a new process is acceptable from my perspective. If users want to eliminate the overhead they may try to set the capacity to around the number of CPU cores (with hyper-threading), assuming that `polars` will not spawn more threads than that. Also I thought that the low CPU case is not so relevant because in this scenario `in_background = FALSE` could be a better option.

src/rust/src/rbackground.rs
sorhawell: I have a smarter suggestion comming up :)

sorhawell:
I will write a counter suggestion and we will compare on benchmark. It might be ok with an unbounded pool.   However maybe some user would like to bound their mem usage with pool size. Also, polars uses 2x thread of physical cores, e.g. 8 thread on my machine. I don't think it is always beneficial to have that many R sessions open at once. If 8 threads and cap=2. This explains the slowness because about 6 threads must create new processes and destroy them again continuously. Maybe you could patch the unbounded pool to not continuously open and close processes but it seems difficult to me.

Sicheng-Pan:
> I will write a counter suggestion and we will compare on benchmark. It might be ok with an unbounded pool. However maybe some user would like to bound their mem usage with pool size. Also, polars uses 2x thread of physical cores, e.g. 8 thread on my machine. I don't think it is always beneficial to have that many R sessions open at once. If 8 threads and cap=2. This explains the slowness because about 6 threads must create new processes and destroy them again continuously. Maybe you could patch the unbounded pool to not continuously open and close processes but it seems difficult to me.

I can also make an option for the user to configure whether the upper limit is a soft limit of hard limit.

src/rust/src/rbackground.rs
Sicheng-Pan: Also for the current implementation I suppose it should always create new process when the pool is empty, just like before?

src/rust/src/rbackground.rs
```diff
@@ -281,10 +281,28 @@ impl RBackgroundPool {
...
+                break Ok(handle);
+            } else {
+                if *cap < 1 {
+                    rerr()
+                        .plain("cannot run backround R process with zero capacity")
+                        .hint("try pl$set_global_rpool_cap(4)")?;
+                }
+                if pool.len() < *cap {
+                    break RBackgroundHandler::new();
+                } else {
```
Sicheng-Pan, line 298: May be the pool is guaranteed to be empty when line 296 is reached, as `pop_front` fails to get an element?

sorhawell:
before [spawn processes in paralell](https://github.com/pola-rs/r-polars/pull/311/commits/4fea62d4c2e1329b2b1403a28297619c05cedafc)
```R
[1] "test 3a"
usr:   51ms sys:   15ms elp:   85ms ||  - 3a +io %bitrate %cpu foreground 
usr:   65ms sys:   86ms elp: 2054ms ||  - 3a +io %bitrate %cpu pool=8 background burn-in  
usr:   63ms sys:   87ms elp:   76ms ||  - 3a +io %bitrate %cpu pool=8 background 
usr:   58ms sys:   68ms elp:  186ms ||  - 3a +io %bitrate %cpu pool=4 background 
usr:   60ms sys:   69ms elp:  182ms ||  - 3a +io %bitrate %cpu pool=2 background 
usr:   60ms sys:   58ms elp:  325ms ||  - 3a +io %bitrate %cpu pool=1 background 
[1] "test 3b"
usr:10888ms sys:  452ms elp:11047ms ||  - 3b %io %bitrate +cpu foreground 
usr:    4ms sys:   12ms elp: 4415ms ||  - 3b %io %bitrate +cpu pool=8 background burn-in  
usr:    3ms sys:    7ms elp: 2891ms ||  - 3b %io %bitrate +cpu pool=8 background 
usr:    4ms sys:    8ms elp: 3273ms ||  - 3b %io %bitrate +cpu pool=6 background 
usr:    4ms sys:    7ms elp: 3709ms ||  - 3b %io %bitrate +cpu pool=4 background 
usr:    4ms sys:    6ms elp: 5648ms ||  - 3b %io %bitrate +cpu pool=2 background 
usr:    4ms sys:    8ms elp:11798ms ||  - 3b %io %bitrate +cpu pool=1 background 
```

after has lower burn-in times 
```R
[1] "test 3a - parallel"
usr:   55ms sys:   16ms elp:   96ms ||  - 3a +io %bitrate %cpu foreground 
usr:   65ms sys:   74ms elp:  713ms ||  - 3a +io %bitrate %cpu pool=8 background burn-in  
usr:   62ms sys:   85ms elp:   99ms ||  - 3a +io %bitrate %cpu pool=8 background 
usr:   70ms sys:  105ms elp:  142ms ||  - 3a +io %bitrate %cpu pool=4 background 
usr:   59ms sys:   67ms elp:  174ms ||  - 3a +io %bitrate %cpu pool=2 background 
usr:   54ms sys:   56ms elp:  321ms ||  - 3a +io %bitrate %cpu pool=1 background 
[1] "test 3b - parallel"
usr:10941ms sys:  459ms elp:11152ms ||  - 3b %io %bitrate +cpu foreground 
usr:    5ms sys:   13ms elp: 3474ms ||  - 3b %io %bitrate +cpu pool=8 background burn-in  
usr:    3ms sys:    9ms elp: 2656ms ||  - 3b %io %bitrate +cpu pool=8 background 
usr:    4ms sys:    7ms elp: 2993ms ||  - 3b %io %bitrate +cpu pool=6 background 
usr:    4ms sys:    7ms elp: 3562ms ||  - 3b %io %bitrate +cpu pool=4 background 
usr:    4ms sys:    7ms elp: 5637ms ||  - 3b %io %bitrate +cpu pool=2 background 
usr:    3ms sys:    6ms elp:11131ms ||  - 3b %io %bitrate +cpu pool=1 background 
```

sorhawell:
[previous pool impl ](https://github.com/pola-rs/r-polars/pull/311/commits/13bd3c733a0c27985d7f50f9db5bef61ce66aa0e)
```R
[1] "test 1a - sequential"
usr:  236ms sys:   29ms elp:  265ms ||  - 1a +io %bitrate %cpu foreground 
usr:  126ms sys:   49ms elp:  522ms ||  - 1a +io %bitrate %cpu background1 
usr:  127ms sys:   48ms elp:  533ms ||  - 1a +io %bitrate %cpu background2 
[1] "test 1b - sequential"
usr:  196ms sys:  114ms elp:  310ms ||  - 1b -io +bitrate %cpu foreground 
usr:  286ms sys:  356ms elp: 1641ms ||  - 1b -io +bitrate %cpu background1 
usr:  277ms sys:  323ms elp: 1623ms ||  - 1b -io +bitrate %cpu background2 
[1] "test 2a - sequential"
usr: 2445ms sys:  544ms elp: 3060ms ||  - 2a %io ~bitrate +cpu foreground 
usr:    1ms sys:    5ms elp: 2857ms ||  - 2a %io ~bitrate +cpu background 
usr:    1ms sys:    7ms elp: 2818ms ||  - 2a %io ~bitrate +cpu background2 
[1] "test 3a - parallel"
usr:   49ms sys:   16ms elp:   86ms ||  - 3a +io %bitrate %cpu foreground 
usr:   67ms sys:   90ms elp: 2117ms ||  - 3a +io %bitrate %cpu pool=8 background burn-in  
usr:   65ms sys:   89ms elp:   93ms ||  - 3a +io %bitrate %cpu pool=8 background 
usr:   79ms sys:  135ms elp:15036ms ||  - 3a +io %bitrate %cpu pool=4 background 
usr:  144ms sys:  389ms elp:108357ms ||  - 3a +io %bitrate %cpu pool=2 background 
usr:  169ms sys:  520ms elp:150738ms ||  - 3a +io %bitrate %cpu pool=1 background 
[1] "test 3b - parallel"
usr:10960ms sys:  485ms elp:11163ms ||  - 3b %io %bitrate +cpu foreground 
usr:    4ms sys:   13ms elp: 4528ms ||  - 3b %io %bitrate +cpu pool=8 background burn-in  
usr:    4ms sys:   10ms elp: 2853ms ||  - 3b %io %bitrate +cpu pool=8 background 
usr:    4ms sys:   11ms elp: 3158ms ||  - 3b %io %bitrate +cpu pool=6 background 
usr:    4ms sys:   14ms elp: 3814ms ||  - 3b %io %bitrate +cpu pool=4 background 
usr:    5ms sys:   18ms elp: 4360ms ||  - 3b %io %bitrate +cpu pool=2 background 
usr:    6ms sys:   19ms elp: 5158ms ||  - 3b %io %bitrate +cpu pool=1 background 
[1] "test 3c - parallel"
usr:11055ms sys:  736ms elp:11688ms ||  - 3c %io +bitrate +cpu foreground  
usr:  166ms sys:  136ms elp: 4587ms ||  - 3c %io +bitrate +cpu pool=8 background burn-in  
usr:  238ms sys:  125ms elp: 3137ms ||  - 3c %io +bitrate +cpu pool=8 background 
usr:  159ms sys:  105ms elp: 3375ms ||  - 3c %io +bitrate +cpu pool=6 background 
usr:  173ms sys:  103ms elp: 3829ms ||  - 3c %io +bitrate +cpu pool=4 background 
usr:  133ms sys:   97ms elp: 4426ms ||  - 3c %io +bitrate +cpu pool=2 background 
usr:  117ms sys:   95ms elp: 5509ms ||  - 3c %io +bitrate +cpu pool=1 background 
```

[new queued pool impl](https://github.com/pola-rs/r-polars/pull/311/commits/4fea62d4c2e1329b2b1403a28297619c05cedafc)
```R
[1] "test 1a - sequential"
usr:  242ms sys:   33ms elp:  276ms ||  - 1a +io %bitrate %cpu foreground 
usr:  130ms sys:   53ms elp:  548ms ||  - 1a +io %bitrate %cpu background1 
usr:  128ms sys:   51ms elp:  541ms ||  - 1a +io %bitrate %cpu background2 
[1] "test 1b - sequential"
usr:  214ms sys:  127ms elp:  345ms ||  - 1b -io +bitrate %cpu foreground 
usr:  292ms sys:  425ms elp: 1991ms ||  - 1b -io +bitrate %cpu background1 
usr:  288ms sys:  371ms elp: 1966ms ||  - 1b -io +bitrate %cpu background2 
[1] "test 2a - sequential"
usr: 2542ms sys:  643ms elp: 3206ms ||  - 2a %io ~bitrate +cpu foreground 
usr:    2ms sys:    6ms elp: 3205ms ||  - 2a %io ~bitrate +cpu background 
usr:    2ms sys:    8ms elp: 3142ms ||  - 2a %io ~bitrate +cpu background2 
[1] "test 3a - parallel"
usr:   58ms sys:   20ms elp:  100ms ||  - 3a +io %bitrate %cpu foreground 
usr:   76ms sys:   86ms elp:  938ms ||  - 3a +io %bitrate %cpu pool=8 background burn-in  
usr:   68ms sys:   93ms elp:   94ms ||  - 3a +io %bitrate %cpu pool=8 background 
usr:   76ms sys:  116ms elp:  151ms ||  - 3a +io %bitrate %cpu pool=4 background 
usr:   66ms sys:   80ms elp:  208ms ||  - 3a +io %bitrate %cpu pool=2 background 
usr:   61ms sys:   66ms elp:  357ms ||  - 3a +io %bitrate %cpu pool=1 background 
[1] "test 3b - parallel"
usr:11721ms sys:  736ms elp:12294ms ||  - 3b %io %bitrate +cpu foreground 
usr:    5ms sys:   19ms elp: 3812ms ||  - 3b %io %bitrate +cpu pool=8 background burn-in  
usr:    4ms sys:    9ms elp: 3092ms ||  - 3b %io %bitrate +cpu pool=8 background 
usr:    4ms sys:   10ms elp: 3283ms ||  - 3b %io %bitrate +cpu pool=6 background 
usr:    4ms sys:    7ms elp: 3835ms ||  - 3b %io %bitrate +cpu pool=4 background 
usr:    4ms sys:    8ms elp: 5859ms ||  - 3b %io %bitrate +cpu pool=2 background 
usr:    4ms sys:    7ms elp:11201ms ||  - 3b %io %bitrate +cpu pool=1 background 
[1] "test 3c - parallel"
usr:11056ms sys:  798ms elp:11492ms ||  - 3c %io +bitrate +cpu foreground  
usr:  165ms sys:  175ms elp: 3674ms ||  - 3c %io +bitrate +cpu pool=8 background burn-in  
usr:  178ms sys:   95ms elp: 3082ms ||  - 3c %io +bitrate +cpu pool=8 background 
usr:  247ms sys:  119ms elp: 3330ms ||  - 3c %io +bitrate +cpu pool=6 background 
usr:  140ms sys:   82ms elp: 3875ms ||  - 3c %io +bitrate +cpu pool=4 background 
usr:   92ms sys:   57ms elp: 5935ms ||  - 3c %io +bitrate +cpu pool=2 background 
usr:   86ms sys:   49ms elp:11403ms ||  - 3c %io +bitrate +cpu pool=1 background 
```

sorhawell:
my conclusion: 
Previous pool impl is equally if number of processes as equal to polars workers or for any sequantial processesing. See "3x pool=8 not burn-in" and "1x+2x"

Previous pool impl was slower in parallel  especially for io bounded tasks for lower number of pool size, because some or most processes would be created and destroyed for each iteration. For cpu bound test 3c, previous pool impl is actually faster, but that is because it was never limited on number of processes.

In general the io overhead of backgroun is about 2x over foreground
The bitrate overhead (inter process communnication) is about 3-15x. But it is tough to compete with arrow zero-cost.
If using r-polars conversion to R vectors which CPU bound that bitrate is not especially limiting see test "3d".

The downside of the new impl it is +200 lines of code. The upside it will allow to use background R processes in more scenarios which are not extreme CPU bound, e.g. for aggragting over groups. User can actually set the actual max number of R processes via `pl$set_global_rpool()`which may be useful to reduce memory use.

sorhawell:
> > I will write a counter suggestion and we will compare on benchmark. It might be ok with an unbounded pool. However maybe some user would like to bound their mem usage with pool size. Also, polars uses 2x thread of physical cores, e.g. 8 thread on my machine. I don't think it is always beneficial to have that many R sessions open at once. If 8 threads and cap=2. This explains the slowness because about 6 threads must create new processes and destroy them again continuously. Maybe you could patch the unbounded pool to not continuously open and close processes but it seems difficult to me.
> 
> I can also make an option for the user to configure whether the upper limit is a soft limit of hard limit.

I could not see an easy way to add a hard limit without some global accounting of total leased/pool process-handlers. How do you find this updated pool impl?

Sicheng-Pan:
> I could not see an easy way to add a hard limit without some global accounting of total leased/pool process-handlers. How do you find this updated pool impl?

@sorhawell I was thinking about this and I came to the conclusion that a counter for leased processes is necessary. Also we would need either a `Condvar` or a queue of `mspc::Sender` so that threads can wait for the in-use `R` processes, and I planned to make this an `Option<_>` so that we could tell if we have a hard limit (`Some(condvar/queue)`) or soft limit (`None`). But I thought this will change a lot to the old implementation, especially because I inevitably need to write a inner struct for the thread pool, so I did not proceed to do this. It seems that you implemented all these except for the `Option<_>` part, and the code looks pretty good to me. Thanks for the help!

sorhawell:
@eitsupi  I have changed the Makefile `make install` to no handle dependencies. I guess `make requirements` could do that. I have added install to `make docs` because some examples require polars being build as a package to be loaded in background R processes.

sorhawell:
@Sicheng-Pan  As far as I see I think the huge PR is ready to go. Many thanks for your patience!

eitsupi (commented):
Thanks for this great work!

A few questions.

src/Makevars
```diff
@@ -20,7 +20,7 @@ $(STATLIB):
 	fi && \
 	export PATH="$(PATH):$(HOME)/.cargo/bin" && \
 	if [ "$(RPOLARS_ALL_FEATURES)" == "true" ]; then \
-		$(BUILD_COMMAND) --all-features; \
+		$(BUILD_COMMAND) --features "simd"; \
```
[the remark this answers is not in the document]
sorhawell: To debug the new polars IPC handling, I needed threads do a lot of std.out printing. I revisiting the new pool handler it will become much easier to debug.  I have hidden the debug printing behind the feature rpolars_debug_print.

I will try to make it prettier

inst/misc/benchmark_rbackground.R
```diff
@@ -0,0 +1,201 @@
+time_print = \(expr, name) {
+    expr |> system.time() |> round(5) |> (get("*"))(1000) |>  formatC(width=5, format = "d") |>
+    paste0("ms") |> (\(time) paste(c("usr", "sys", "elp"),time[1:3],sep=":"))() |>
+    cat()
+    cat(" || ",name,"\n")
```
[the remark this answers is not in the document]
sorhawell: adding fmt https://github.com/pola-rs/r-polars/pull/311/commits/46058dbdcf72de11894a9888f4a255dde216893e

src/Makevars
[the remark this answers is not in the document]
eitsupi: Ok, thank for the update.

sorhawell:
@Sicheng-Pan  many thanks !!

PR #312, opened 2023-07-06
build: add devcontainer.json for to support Dev Containers

eitsupi:
Same as pola-rs/polars#9690, pola-rs/polars#9697

This definitely works well on amd64 machines (I use it all the time), but be aware that if we use it on an arm64 machine, it will cause source installation of the `arrow` package and take a long time to build.


eitsupi:
TODO: install cmake

eitsupi:
Closing in favor of #633

PR #317, opened 2023-07-10
fix: use libR-sys 0.4.0

### LINKED ISSUE #315: Should not use libR-sys 0.5.0 on CRAN
See mlverse/tok#4

One of the flavors on CRAN seems to fail to build libR-sys 0.5.0.
We are currently using libR-sys 0.5.0 here.

https://github.com/pola-rs/r-polars/blob/1c1fc73256f1ebd385ffd5d1e5caaf42e85f0276/src/rust/Cargo.lock#L968-L972

@sorhawell I believe we are currently using a fork of extendr, can we revert libR-sys back to 0.4?
It would be nice to be able to do this prior to submission to CRAN if possible. (If we submit to CRAN first, you should get an immediate warning and be asked to resubmit within a month.)

eitsupi:
Fix #315

ref extendr/extendr#581

### DIFF
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -497,7 +497,7 @@ checksum = "0198b9d0078e0f30dedc7acbb21c974e838fc8fae3ee170128658a98cb2c1c04"
 [[package]]
 name = "extendr-api"
 version = "0.4.0"
-source = "git+https://github.com/rpolars/extendr?branch=pr473_553_555_566#e8c03bae5568cbc24fefa1a2eec9c35f5ff299f4"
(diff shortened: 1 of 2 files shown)


sorhawell:
@eitsupi  sorry I missed this PR in the story this morning.

extendr #581 PR did not pass extendr unit test but it seems to be graphical device errors which does not involve polars.

I changed this PR to point directly on branch "rpolars/extendr/pl0.7.0rc" which we own.

sorhawell (approved):

eitsupi:
Sorry I am not able to see this. I am really busy cleaning up my room right now.......

Does it look like you forgot to update the lock file?
In other words, the updates you have made here are not actually reflected.

PR #322, opened 2023-07-11
Not cran dev builds

sorhawell:
*Tip:* To speed up the local rextendr::document() or R CMD check, use of the following:

`develop_polars.R` is stand-alone and can be used in other branches

I made these as I struggled with deleting cargo cache again every time `NOT_CRAN` was set incorrectly or one of the other build flag.

  ```r
  source("inst/misc/develop_polars.R")
  
  # to rextendr:document() + not_cran + load packages + all_features
  load_polars()
  
  # to check package + reuses previous compilation in check, protects against deletion
  check_polars()
  
  # to build package
  build_polars()
  
  #  to submit to cran
  submit_polars()
  
  ```


sorhawell:
@eitsupi the new cran friendly makevars got me to delete target alot of times. I wrote these `load_polars()` and `check_polars(YOUR_TARGET_PATH)` for faster development.


The reason `make build` is not always ideal for development is it requires to all run rextendr::document() or load_all() again in the session where developing.

sorhawell:
[now available in main via 0.7.0rc ](https://github.com/pola-rs/r-polars/blob/main/inst/misc/develop_polars.R)

Issue #261, opened 2023-06-27
implement `<LazyFrame>.profile()`


sorhawell:
nice to have, also for tutorials on polars.
[https://pola-rs.github.io/polars/py-polars/html/reference/lazyframe/api/polars.LazyFrame.profile.html#polars.LazyFrame.profile](https://pola-rs.github.io/polars/py-polars/html/reference/lazyframe/api/polars.LazyFrame.profile.html#polars.LazyFrame.profile)

Sicheng-Pan:
Completed with #323

PR #323, opened 2023-07-12
Implement lazyframe profiling and optimization toggles

Sicheng-Pan:
#261


sorhawell:
we are missing a unit test. py-polars do not always unit test this either. We need to extract the curret settings from a LazyFrame / logicalplan. I have previous found bugs in rust-polars by testing such optimization settings.

sorhawell:
@Sicheng-Pan  I'm ok with merging this PR now.

I think you mentioned you wanted to allow both the setting opt-toggles via collect as in py-polars, but that it also makes sense to se individually as in rust-polars. I'm agree with that, if you want to add that back in.

sorhawell:
This PR is needed for #343  which could use opt-toogles +  polars CSE feature + robj_to!(i32, robj)

I realized that `profile` and `fetch` do collect like `collect` each of this functions need individual support for R_function. If using the plain rust `.profile()` method then R fn `.map()` is not supported.
Refactored and added to internal functions `collect_with_r_func_support` and `collect_with_r_func_support`. It should be easy to later also add `fetch_with_r_func_support`

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
...
+  streaming = FALSE,
+  collect_in_background = FALSE
+) {
+
+  if (isTRUE(no_optimization)) {
+    predicate_pushdown = FALSE
+    projection_pushdown = FALSE
+    slice_pushdown = FALSE
+    common_subplan_elimination = FALSE
+  }
```
etiennebacher, line 304: Shouldn't you put `type_coercion = FALSE` and `simplify_expression = FALSE` as well?

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
...
+
+  if (isTRUE(streaming)) {
+    common_subplan_elimination = FALSE
+  }
+
+  collect_f = if( isTRUE(collect_in_background)) {
```
etiennebacher, line 310:
```suggestion
  collect_f = if(isTRUE(collect_in_background)) {
```

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
...
+      streaming
+    ) |>
+    and_then(collect_f) |>
+    unwrap("in $collect():")
 }
 
 #' @title New DataFrame from LazyFrame_object$collect()
-#' @description collect DataFrame by lazy query
+#' @description collect DataFrame by lazy query (SOFT DEPRECATED)
+#' @details This function is soft deprecated. Use $collect(collect_in_background = TRUE) instead
```
etiennebacher, line 332: I don't understand this doc, how is this different from `$collect()`? I know it gets deprecated but until it's removed it will appear in the docs so it might be good to clarify. Also the example uses `collect()` instead of `collect_in_background()`

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
 
 #' @title New DataFrame from LazyFrame_object$collect()
 #' @description collect DataFrame by lazy query
+#' @param type_coercion Boolean. Do type coercion optimization.
+#' @param predicate_pushdown  Boolean. Do predicate pushdown optimization.
+#' @param projection_pushdown  Boolean. Do projection pushdown optimization.
+#' @param simplify_expression  Boolean. Run simplify expressions optimization.
+#' @param no_optimization  Boolean. Turn off (certain) optimizations.
+#' @param slice_pushdown  Boolean. Slice pushdown optimization.
```
etiennebacher, line 273: Can you improve the description of the parameters? For example, change "Do type coercion optimization." to "Coerce types such that operations succeed and run on minimal required memory."

There are good descriptions of each param here: https://pola-rs.github.io/polars-book/user-guide/lazy/optimizations/

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
...
+#' @param common_subplan_elimination  Boolean. Will try to cache branching subplans that occur on
+#' self-joins or unions.
+#' @param streaming  Boolean. Run parts of the query in a streaming fashion
+#' (this is in an alpha state)
+#' @param collect_in_background Boolean. Detach this query from R session. Computation will start
+#' in background. Get a handle which later can be converted into the resulting DataFrame. Useful
+#' in interactive mode to not lock R session.
+#' @details
+#' Note: use `$fetch()` if you want to run your query on the first `n` rows only.
```
etiennebacher, line 282:
```suggestion
#' Note: use `$fetch(n)` if you want to run your query on the first `n` rows only.
```

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
...
+#' This can be a huge time saver in debugging queries.
 #' @keywords LazyFrame DataFrame_new
-#' @return collected `DataFrame`
+#' @return collected `DataFrame` or if colkect
```
etiennebacher, line 285:
```suggestion
#' @return A `DataFrame`
```
(no idea what you meant after the "or")

R/lazyframe__lazy.R
```diff
@@ -265,15 +265,71 @@ LazyFrame_filter = "use_extendr_wrapper"
...
 #' @examples pl$DataFrame(iris)$lazy()$filter(pl$col("Species") == "setosa")$collect()
```
etiennebacher, line 286:
```suggestion
#' @examples pl$LazyFrame(iris)$filter(pl$col("Species") == "setosa")$collect()
```

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
     unwrap("in $dtypes()")
 })
 
+#' @title Collect and profile a lazy query.
+#' @description This will run the query and return a tuple containing the materialized DataFrame and
```
etiennebacher, line 986:
```suggestion
#' @description This will run the query and return a list containing the materialized DataFrame and
```
I guess an R list is the equivalent of Python tuple?

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
...
+#'  a DataFrame that contains profiling information of each node that is executed.
+#' @details The units of the timings are microseconds.
+#'
+#' @keywords LazyFrame
+#' @return List of two DataFrames, (collected result, profile stats)
```
etiennebacher, line 991:
```suggestion
#' @return List of two `DataFrame`s: one with the collected result, the other with the timings of each step.
```

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
...
+#' @examples
+#'
+#' #Use $profile() to compare two queries
```
etiennebacher, line 994:
```suggestion
#' # Use $profile() to compare two queries
```

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
...
+#'
+#' # print one '.', take a series convert to r vector, take first value, add 5
+#' r_func = \(s) {cat(".");s$to_r()[1] + 5}
```
etiennebacher, line 997: example hard to grasp, isn't there a most simple / realistic example?

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
...
+#'
+#' # map each Species-group of each numeric column with an R function, takes ~7000us slow !
```
etiennebacher, line 999:
```suggestion
#' # map each Species-group of each numeric column with an R function, takes ~7000us (slow!)
```

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
...
+#' pl$LazyFrame(iris)$
+#'   sort("Sepal.Length")$  #for no specific reason
```
etiennebacher, line 1001:
```suggestion
#'   sort("Sepal.Length")$  
```

R/lazyframe__lazy.R
```diff
@@ -926,6 +982,37 @@ LazyFrame_dtypes = method_as_property(function() {
...
+#'   groupby("Species", maintain_order = TRUE)$
+#'   agg(pl$col(pl$Float64)$apply(r_func))$
+#'   profile()
+#'
+#' # map each Species-group with native polars, takes ~120us better
```
etiennebacher, line 1006:
```suggestion
#' # map each Species-group with native polars, takes ~120us only
```

tests/testthat/test-lazy_profile.R
```diff
@@ -0,0 +1,41 @@
+test_that("can modify lazy profile settings", {
+  #some way to check if .pr$LazyFrame$optimization_toggle works
```
etiennebacher, line 2:
```suggestion
  # TODO: some way to check if .pr$LazyFrame$optimization_toggle works
```

etiennebacher (changes requested):
I could only review the R part, need someone else for the Rust part

R/lazyframe__lazy.R
sorhawell: [I just match py-polars definition of no_optimization](https://github.com/pola-rs/polars/blob/d9ff23c5316ad489de9e4411c5e3dc275a214e71/py-polars/polars/lazyframe/frame.py#L1580)

with rust-polars 0.31.0  a few more option will be added

src/rust/src/utils/mod.rs
```diff
@@ -501,28 +528,79 @@ pub fn robj_to_usize(robj: extendr_api::Robj) -> RResult<usize> {
     robj_to_u64(robj).and_then(try_u64_into_usize)
 }
 
+fn err_no_nan<T>() -> RResult<T> {
+    rerr().plain("any NA value is not allowed here".to_string())
```
Sicheng-Pan, line 532: `.to_string()` could be omitted since `Into<String>` is implemented for `&str`.

R/lazyframe__lazy.R
sorhawell: This documentation and behavior will be updated alot in #311. I suggest to just leave it as is here.

src/rust/src/utils/mod.rs
```diff
@@ -501,28 +528,79 @@ pub fn robj_to_usize(robj: extendr_api::Robj) -> RResult<usize> {
...
+            Some(rerr().plain("does not support this R type for this conversion".to_string()))
+        }
     }
-    .ok_or(RPolarsErr::new())
+    .unwrap_or_else(err_no_nan)
     .bad_robj(&robj)
-    .mistyped(tn::<i64>());
+    .mistyped(tn::<i32>())
+    .when("converting into type");
+}
```
Sicheng-Pan, line 582: Maybe consider about reusing the codes for `robj_to_i64` with additional bound checks?

Sicheng-Pan: Maybe something like `Ok(i32::try_from(robj_to_i64(robj)?)?)`

R/lazyframe__lazy.R
sorhawell: adding example `pl$LazyFrame()$select(pl$lit(2) + 2)$profile()`. Ordering other examples by complexity.

R/lazyframe__lazy.R
sorhawell: add in other commit

tests/testthat/test-lazy_profile.R
sorhawell: will be possible from rust-polars 0.31.0

src/rust/src/utils/mod.rs
sorhawell: that is certainly also an option and I did so for usize. Errors can look weird though for the user if the error is about i64 but the function needed an i32. I tried to boil the conversion so it would not be to lengthy to do for each type, at least for the regulars f64, i64 i32.

src/rust/src/utils/mod.rs
sorhawell: done

Sicheng-Pan: Actually there are multiple occurrences of `.to_string()` that you may want to remove

src/rust/src/utils/mod.rs
sorhawell: I must admit though, I do think I have gotten myself into to many rabbit holes lately and not progressed enough. Sry bout that :)

etiennebacher (approved):
I just did minor doc changes. @sorhawell since you and @Sicheng-Pan have worked on the Rust part I guess it's already reviewed.

I'm just waiting the CI to pass before merging, thank you both!

PR #327, opened 2023-07-18
DO NOT MERGE: check if shared memory commit worked on all OS

sorhawell:


sorhawell:
oups later commits did pass

PR #328, opened 2023-07-19
DO NOT MERGE: Try fix proc linux

sorhawell:


Sicheng-Pan:
No longer needed since this is fixed in #311

PR #340, opened 2023-07-31
build: always NOT_CRAN=true

### LINKED ISSUE #339: NOT_CRAN -> IS_CRAN?
@eitsupi  would it be reasonable to move away from the NOT_CRAN  environment variable approach?

I know me and @Sicheng-Pan  are killing our cache by accident several times. It takes my computer 20 min per time to recompile. I think this will certainly be a hurdle for any new contributor.

Now when CRAN is not on the table, we could invert the settings to be a IS_CRAN setting? If CRAN ever comes on the table again, maybe it easier to have a dedicated release script, which modifies the makevars file or any other file.

eitsupi:
close #339, reopen #29

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -15,7 +15,7 @@ $(SHLIB): $(STATLIB)
 CARGOTMP = $(CURDIR)/.cargo
 
 $(STATLIB):
-	if [ "$(NOT_CRAN)" != "true" ]; then \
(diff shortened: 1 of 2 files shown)


sorhawell:
thanks is a very clean switch to switch, which could be switched again if venturing back into CRAN-world :)

sorhawell (approved):

PR #342, opened 2023-08-01
Fix describe colon

sorhawell:
internally $describe() uses `:` as separtor between ´agg.fun.name` and `column.name`. Column names with `:` would get split and only last part would remain. This is fixed by using ´$splitn(":",2)` internally.

add a unit test

chore:
revert nanoarrow.Rd
rextendr 0.3.1 -> 0.3.1.9000 or should that be 0.3.1? I'm fine with either.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -97,5 +97,5 @@ Collate:
     'translation.R'
     'vctrs.R'
     'zzz.R'
-Config/rextendr/version: 0.3.1
+Config/rextendr/version: 0.3.1.9000
 VignetteBuilder: knitr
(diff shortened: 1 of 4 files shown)


eitsupi:
> rextendr 0.3.1 -> 0.3.1.9000 or should that be 0.3.1? I'm fine with either.

If I understand correctly, there is currently no difference between 0.3.1 and the development version in terms of useful features here. So I recommend adopting 0.3.1.

eitsupi (changes requested):
Could you update the PR title?

eitsupi (approved):
Thanks!

Issue #344, opened 2023-08-03
Multiplying with `*` symbol only works in one direction


etiennebacher:
I can't use  `*` with e.g `pl$max()`:

``` r
library(polars)
packageVersion("polars")
#> [1] '0.7.0.9000'

df = pl$DataFrame(x = c(0, 0, 1), y = c(1, 2, 3))

# fine
df$filter(pl$col("y") > pl$max("y")$mul(0.5))
#> shape: (2, 2)
#> ┌─────┬─────┐
#> │ x   ┆ y   │
#> │ --- ┆ --- │
#> │ f64 ┆ f64 │
#> ╞═════╪═════╡
#> │ 0.0 ┆ 2.0 │
#> │ 1.0 ┆ 3.0 │
#> └─────┴─────┘

# error
df$filter(pl$col("y") > 0.5 * pl$max("y"))
#> Error in e1$mul: $ operator is invalid for atomic vectors
```

But it works in `py-polars`:
```python
>>> import polars as pl
>>> df = pl.DataFrame(dict(
...     x=[0, 0, 1, 1], 
...     y=[1, 2, 3, 3]) 
... )
>>>
>>> df.filter((pl.col("y") > 0.5 * pl.max("y")))
shape: (3, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 0   ┆ 2   │
│ 1   ┆ 3   │
│ 1   ┆ 3   │
└─────┴─────┘
```

etiennebacher:
Due to `*.Expr` only working when the first term is an `Expr`. Simplified reprex:

``` r
library(polars)

pl$max("y") * 0.5
#> polars Expr: [(col("y").max()) * (0.5)]

0.5 * pl$max("y")
#> Error in e1$mul: $ operator is invalid for atomic vectors
```

etiennebacher:
Closed with #346

PR #346, opened 2023-08-03
Rework Expr operators and errors 

sorhawell:
to solve #344

@etiennebacher  oups I also tried to solve that issue :)

I think we can combine #346  and #345


etiennebacher:
I'd say let's take your fix and add my tests here? ;)

sorhawell:
derived `wrap_e` from `robj_to!`  + update all binary operator implementation to use new RPolarsErr handling

R/expr__expr.R
```diff
@@ -309,15 +336,15 @@ Expr_gt = function(other) {
...
+  .pr$Expr$eq(self, other) |> unwrap("in $eq()")
 }
 #' @export
 #' @details
 #' See Inf,NaN,NULL,Null/NA translations here \code{\link[polars]{docs_translations}}
 #' @param e1 lhs Expr
 #' @param e2 rhs Expr or anything which can become a literal Expression
 #' @rdname Expr_eq
-"==.Expr" = function(e1, e2) e1$eq(e2)
+"==.Expr" = function(e1, e2) result(wrap_e(e1)$eq(e2)) |> unwrap("using the '='-operator")
```
etiennebacher, line 347:
```suggestion
"==.Expr" = function(e1, e2) result(wrap_e(e1)$eq(e2)) |> unwrap("using the '=='-operator")
```

R/expr__expr.R
```diff
@@ -1001,7 +1034,7 @@ Expr_rpow = function(base) {
 
 #' @rdname Expr_rpow
 #' @export
-"%**%.Expr" = function(e1, e2) e1$rpow(e2)
+"%**%.Expr" = function(e1, e2) result(wrap_e(e1)$rpow(e2)) |> unwrap("using the '**'-operator")
```
etiennebacher, line 1037: What is the equivalent of this `rpow` in R?  

`**` is the same as `^` so putting `**` in the error message doesn't seem correct

tests/testthat/test-expr.R
```diff
@@ -52,6 +52,94 @@ test_that("expression Arithmetics", {
...
+      FALSE | TRUE
+    )$to_data_frame(),
+    data.frame(
+      oneexp_onecol = dat_df$x | TRUE,
+      onelit_oneexp = FALSE | dat_df$y,
+      x = dat_df$x | dat_df$y,
+      literal = FALSE | TRUE
+    )
+  )
+})
```
etiennebacher, line 141:
```suggestion
# & and | require another test dataset, it can't be the one above
test_that("logical ops symbol work with expressions", {
  dat = pl$DataFrame(
    x = c(TRUE, FALSE, TRUE, FALSE),
    y = c(TRUE, TRUE, FALSE, FALSE)
  )
  dat_df = dat$to_data_frame()
  expect_equal(
    dat$select(
      (pl$col("x") & TRUE)$alias("oneexp_onelit"),
      (FALSE & pl$col("y"))$alias("onelit_oneexp"),
      pl$col("x") & pl$col("y"),
      FALSE & TRUE
    )$to_data_frame(),
    data.frame(
      oneexp_onelit = dat_df$x & TRUE,
      onelit_oneexp = FALSE & dat_df$y,
      x = dat_df$x & dat_df$y,
      literal = FALSE & TRUE
    )
  )
  expect_equal(
    dat$select(
      (pl$col("x") | TRUE)$alias("oneexp_onelit"),
      (FALSE | pl$col("y"))$alias("onelit_oneexp"),
      pl$col("x") | pl$col("y"),
      FALSE | TRUE
    )$to_data_frame(),
    data.frame(
      oneexp_onelit = dat_df$x | TRUE,
      onelit_oneexp = FALSE | dat_df$y,
      x = dat_df$x | dat_df$y,
      literal = FALSE | TRUE
    )
  )
})
```

etiennebacher (changes requested):
Same as before, I can't review the Rust part. Minor things to change

R/expr__expr.R
sorhawell: [oh I see py-polars does not document dunder method __pow__](https://github.com/pola-rs/polars/blob/d9ff23c5316ad489de9e4411c5e3dc275a214e71/py-polars/polars/expr/expr.py#L211)

I should remove it. It is only needed if rhs inherits `Expr` but not left. R does that automatically

etiennebacher: can you also remove the related TODO I added in the tests?

sorhawell: rpow is now removed completely

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # polars (development version)
 
+## BREAKING CHANGES
+- `$rpow()` is removed. It should never have been translated. Use `^` and `$pov()` instead (#346).
```
etiennebacher, line 4:
```suggestion
- `$rpow()` is removed. It should never have been translated. Use `^` and `$pow()` instead (#346).
```

R/expr__expr.R
```diff
@@ -2441,10 +2407,13 @@ Expr_limit = function(n = 10) {
...
 
 
-#' Exponentiation `^` or `**`
+#' Exponentiation `^` and `**`.
 #' @description Raise expression to the power of exponent.
 #' @keywords Expr
 #' @param exponent exponent
+#' @details in py-polars python `^` is the xor operator. The R interpreter will replace the
+#' `**` with `^`, such that `**` means `^`. Except e.g. in strings like "**". Use of `**` will
+#' not work in some cases, because it is not a real separate operator in R.
```
etiennebacher, line 2416:
```suggestion
#' @details In python-polars, `^` is the xor operator. The R interpreter will replace the
#' `^` with `**`, which means that you can use both `**` and `^` for the exponentiation. Use of `**` will
#' not work in some cases, because it is not a real separate operator in R.
```
Are my corrections what you meant?

Also, you say "Use of `**` will not work in some cases, because it is not a real separate operator in R." 
-> Not clear, when does it fail? Can we show examples of this?

etiennebacher (changes requested):

R/expr__expr.R
sorhawell: It is strange not well documented quirk of R 

I will add something like this to documentation.

``` r
#in a new session
2^1
#> [1] 2
2**1 #this works because ** is converted to the ^-operator  by R interpreter
#> [1] 2

get("^")(2,1) #this works because there exists a function called "^"
#> [1] 2
get("**")(2,1)  #the R interpreter will not convert "**" to "^"  and there is no function named "**"
#> Error in get("**"): object '**' not found
```

<sup>Created on 2023-08-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

looking up `?"**"`
** is translated in the parser to ^, but this was undocumented for many years. It appears as an index entry in Becker et al (1988), pointing to the help for Deprecated but is not actually mentioned on that page. Even though it had been deprecated in S for 20 years, it was still accepted in R in 2008.

etiennebacher: I don't think you should add the examples above in the docs, they are more related to base R than to `polars`. I think mentioning that `^` and `**` mean the same thing in `polars` expressions should be enough. If you have some use cases where using `**` doesn't work then they should be shown in examples.

etiennebacher (approved):

Issue #204, opened 2023-05-10
missing lazy functions


sorhawell:
the following functions  `py-polars/polars/functions/lazy.py` are not yet translated in `./R/functions__lazy.R` (file renamed in #196).

tag any merged translated, or write your name after to reserve it for a PR

- [x] pl$map
- [x] pl$apply
- [x] pl$fold (#403)
- [x] pl$reduce (#403)
- [ ] pl$cum_fold
- [ ] pl$cum_reduce
- [x] pl$exclude
- [ ] pl$groups (maybe wait for https://github.com/pola-rs/polars/issues/13274)
- [ ] pl$format
- [x] pl$cumsum
- [x] pl$any
- [x] pl$all cover bitwise also
- [x] pl$quantile
- [x] pl$arange
- [x] pl$collect_all => @etiennebacher: I don't think we need to implement this, it's basically `lapply(<list of LF>, \(x) x$collect())`
- [ ] pl$repeat
- [x] pl$arg_where
- [x] pl$spearman_rank_corr --- Deprecated in pypolars
- [x] pl$pearson_corr --- Deprecated in pypolars
- [x] pl$corr (#351)
- [x] pl$cov (#351)
- [x] pl$arg_sort_by
- [x] pl$duration
- [x] pl$datetime_
- [x] pl$date_
- [x] pl$concat_str_ (#349)
- [x] pl$from_epoch (#708)
- [x] pl$rolling_cov (#351)
- [x] pl$rolling_corr (#351)

eitsupi:
I close it because it is outdated.

PR #351, opened 2023-08-07
Implement `cov`, `corr`, `rolling_cov`, `rolling_corr`

Sicheng-Pan:
#204


sorhawell:
Thank you for implementing these. We should avoid using extendr-internal conversion via `&Expr` as it will bypass our own conversions, not throw an RPolarsErr and will except fewer valid inputs. [See also here.](https://github.com/pola-rs/r-polars/pull/349#issuecomment-1666196827)

`robj_to!` is currently the way. `extendr` does not support custom internal conversions. Alternatives are generic trait wrappers or a macro like robj_to!. [See some examples here](https://github.com/sorhawell/does_it_leak/blob/wrap/src/rust/src/lib.rs)

R/functions__lazy.R
```diff
@@ -844,3 +844,95 @@ pl$struct = function(
...
+#' @param window_size int The length of the window
+#' @param min_periods NULL or int The number of values in the window that should be non-null before computing a result.
+#' If NULL, it will be set equal to window size.
+#' @param ddof integer Delta Degrees of Freedom: the divisor used in the calculation is N - ddof, where N represents the number of elements. By default ddof is 1.
+#' @return Expr for the computed rolling covariance
+#' @examples
+#' lf <- pl$LazyFrame(data.frame(a = c(1, 8, 3), b = c(4, 5, 2)))
+#' lf$select(pl$rolling_cov("a", "b", window_size = 2))$collect()
+pl$rolling_cov = function(a, b, window_size, min_periods = NULL, ddof = 1) {
+  if (is_string(a)) {
```
sorhawell, line 877: use `robj_to!(PLExprCol, a)?` on rust side

R/functions__lazy.R
```diff
@@ -844,3 +844,95 @@ pl$struct = function(
...
+#' @param ddof integer Delta Degrees of Freedom: the divisor used in the calculation is N - ddof, where N represents the number of elements. By default ddof is 1.
+#' @param propagate_nans bool Used only when calculating the spearman rank correlation.
+#' If `True` any `NaN` encountered will lead to `NaN` in the output.
+#' Defaults to `False` where `NaN` are regarded as larger than any finite number and thus lead to the highest rank.
+#' @return Expr for the computed correlation
+#' @examples
+#' lf <- pl$LazyFrame(data.frame(a = c(1, 8, 3), b = c(4, 5, 2)))
+#' lf$select(pl$corr("a", "b", method = "spearman"))$collect()
+pl$corr = function(a, b, method = "pearson", ddof = 1, propagate_nans = FALSE) {
+  if (is_string(a)) {
```
sorhawell, line 904: as above

R/functions__lazy.R
```diff
@@ -844,3 +844,95 @@ pl$struct = function(
...
+#' @param window_size int The length of the window
+#' @param min_periods NULL or int The number of values in the window that should be non-null before computing a result.
+#' If NULL, it will be set equal to window size.
+#' @param ddof integer Delta Degrees of Freedom: the divisor used in the calculation is N - ddof, where N represents the number of elements. By default ddof is 1.
+#' @return Expr for the computed rolling correlation
+#' @examples
+#' lf <- pl$LazyFrame(data.frame(a = c(1, 8, 3), b = c(4, 5, 2)))
+#' lf$select(pl$rolling_corr("a", "b", window_size = 2))$collect()
+pl$rolling_corr = function(a, b, window_size, min_periods = NULL, ddof = 1) {
+  if (is_string(a)) {
```
sorhawell, line 927: as above

R/functions__lazy.R
sorhawell: add unwrap here

src/rust/src/lazy/dsl.rs
```diff
@@ -2210,6 +2205,61 @@ impl Expr {
...
+        Ok(pl::rolling_cov(
+            a.0.clone(),
+            b.0.clone(),
+            new_rolling_cov_options(window_size, min_periods, ddof)?,
+        )
+        .into())
+    }
+
+    pub fn corr(
+        a: &Expr,
```
sorhawell, line 2229: avoid use of `&Expr` (see comment) or any other arg type than `Robj` it is still around in code base. But every PR should chip away at old conversions.

sorhawell (changes requested):
add unit test which compares to equivalents in R  see e.g. "pl$std pl$var" -test . You may use something from `data.table` to test expected rolling behavior. It is not uncommon there is a difference in behavior. If it is a bug in rust-polars (it happens) we can write a PR to them, or otherwise explain in docs when/why polars is behaving different than plain R.

R/functions__lazy.R
Sicheng-Pan: updated

R/functions__lazy.R
Sicheng-Pan: updated

R/functions__lazy.R
Sicheng-Pan: updated

src/rust/src/lazy/dsl.rs
Sicheng-Pan: updated

sorhawell (approved):
Looks very good to me. Thanks a lot!

sorhawell (approved):
Looks very good to me. Many thanks :)

PR #353, opened 2023-08-07
Moderate ** claims in docs + rework title

sorhawell:
@etiennebacher TY for your persistence. I previously had issues rendering ** examples with ::reprex and in docs examples. I cannot reproduce that anymore. I likely have have done something as silly as using a dev load instead of fully building the package.

[BTW it seems titles do not support quotes. Is it better to just avoid quotes in titles ?](https://rpolars.github.io/reference/Expr_pow/)

image


R/expr__expr.R
```diff
@@ -2411,11 +2411,9 @@ Expr_limit = function(n = 10) {
 #' @description Raise expression to the power of exponent.
 #' @keywords Expr
 #' @param exponent exponent
-#' @details Use of `**` is not recommended in r-polars. The R interpreter will replace the `**` with
-#' `^`, such that `**` means `^`. Except e.g. in strings like "**". Use of `**` will not work for
-#' some internal function-lookups, because it is not a real separate operator in R. See example
-#' below or read further at `?"**"`. In py-polars python `^` is the XOR operator and `**` is the
-#' exponentiation operator.
+#' @details The R interpreter will replace the `**` with `^`, such that `**` means `^`. Except e.g.
+#' in strings like "**". Read further at `?"**"`. In py-polars python `^` is the XOR operator and
```
etiennebacher, line 2415:
```suggestion
#' @details The R interpreter will replace the `**` with `^`, such that `**` means `^` (except in
#' strings e.g. "**"). Read further at `?"**"`. In py-polars python `^` is the XOR operator and
```

etiennebacher:
Thanks, I think we can either drop the code formatting in title or in this case just remove the symbols (so the title would just be "Exponentiation")

etiennebacher (approved):

Issue #354, opened 2023-08-08
`sort()` doesn't work anymore if I don't explicitly load `polars`


etiennebacher:
There's an error when I run `$sort()` on a `DataFrame` created without explicitly loading `polars`. I think this used to work until recently so one the latest commits must have changed this:
``` r
### Doesn't work if I don't load polars
polars::pl$DataFrame(mtcars)$sort("mpg")
#> Error: in sort(): the arg [...] or Execution halted with the following contexts
#>    1: The argument [by] caused an error
#>    2: When converting element 1 into an Expr
#>    3:  not convertible into Expr because:
#> 
#>    4: The argument [robj] caused an error
#>    5: cannot be converted into an Expr
#>    6: Got value [Rvalue: ["mpg"], Rsexp: Strings, Rclass: ["character"]]
#>    7: object 'pl' not found
#>  [ERROR CUT]

library(polars)

polars::pl$DataFrame(mtcars)$sort("mpg")
#> shape: (32, 11)
#> ┌──────┬─────┬───────┬───────┬───┬─────┬─────┬──────┬──────┐
#> │ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ … ┆ vs  ┆ am  ┆ gear ┆ carb │
#> │ ---  ┆ --- ┆ ---   ┆ ---   ┆   ┆ --- ┆ --- ┆ ---  ┆ ---  │
#> │ f64  ┆ f64 ┆ f64   ┆ f64   ┆   ┆ f64 ┆ f64 ┆ f64  ┆ f64  │
#> ╞══════╪═════╪═══════╪═══════╪═══╪═════╪═════╪══════╪══════╡
#> │ 10.4 ┆ 8.0 ┆ 472.0 ┆ 205.0 ┆ … ┆ 0.0 ┆ 0.0 ┆ 3.0  ┆ 4.0  │
#> │ 10.4 ┆ 8.0 ┆ 460.0 ┆ 215.0 ┆ … ┆ 0.0 ┆ 0.0 ┆ 3.0  ┆ 4.0  │
#> │ 13.3 ┆ 8.0 ┆ 350.0 ┆ 245.0 ┆ … ┆ 0.0 ┆ 0.0 ┆ 3.0  ┆ 4.0  │
#> │ 14.3 ┆ 8.0 ┆ 360.0 ┆ 245.0 ┆ … ┆ 0.0 ┆ 0.0 ┆ 3.0  ┆ 4.0  │
[...]

PR #355, opened 2023-08-08
fix use polars without library(polars)

sorhawell:
solves #354

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,6 +24,7 @@ Suggests:
     arrow,
     bench,
     bit64,
+    callr,
     data.table,
     knitr,
     lubridate,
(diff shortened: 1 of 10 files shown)


DESCRIPTION
```diff
@@ -49,7 +49,8 @@ Config/Needs/dev:
     robinlovelace/styler.equals,
     spelling,
     stringr,
-    styler
+    styler,
+    callr
```
etiennebacher, line 53: can you move it to keep it the alphabetical order of deps?

sorhawell:
Maybe I have to upgrade `callr` to `suggests` to use it in unit-test.

etiennebacher (approved):

etiennebacher:
I can confirm this works, thanks the quick fix @sorhawell

PR #359, opened 2023-08-10
fix collect(collect_in_background=TRUE)

sorhawell:
`collect(collect_in_background = TRUE)`  referred to old removed implementation.

fix + unit test + style some `<-`

no news because bug was introduced in this 9000 version and has not been released

### DIFF
--- a/R/lazyframe__lazy.R
+++ b/R/lazyframe__lazy.R
@@ -315,7 +315,7 @@ LazyFrame_collect = function(
   }
 
   collect_f = if (isTRUE(collect_in_background)) {
-    .pr$LazyFrame$collect_background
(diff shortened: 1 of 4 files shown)


Sicheng-Pan (approved):
Looks good to me. The unit tests for the background execution weren't broken so I thought everything's ok, but apparently that's not the case. Should we merge `collect_in_background` and `collect`?

sorhawell:
> Looks good to me. The unit tests for the background execution weren't broken so I thought everything's ok, but apparently that's not the case. Should we merge `collect_in_background` and `collect`?

It was a bit random I noticed it. Yeah, I would not mind fusing the methods. Let's look at that in a new PR.

Issue #288, opened 2023-07-02
Show rust-polars version via `polars_info()`


eitsupi:
Needs latest polars (> 0.30) pola-rs/polars#9660

sorhawell:
#363 is an R side implementation. Suggest to leave this PR open until rust impl is available.

PR #363, opened 2023-08-11
Add rust polars version info

sorhawell:
this is an R side implementation. A rust side implementation will likely be available from rust-polars 0.31.0
https://github.com/pola-rs/polars/pull/9660

This partially resolves: https://github.com/pola-rs/r-polars/issues/288

### DIFF
--- a/R/info.R
+++ b/R/info.R
@@ -5,7 +5,9 @@ RUST_POLARS_VERSION = (\() {
   Cargo.lock = readLines("./src/rust/Cargo.lock")
   polars.idx = which(Cargo.lock == r"{name = "polars"}")[1]
   this_line = Cargo.lock[polars.idx + 1]
-  if(isTRUE(substr(this_line,1,7) == "version")) return(substr(this_line,12,nchar(this_line)-1))
+  if (isTRUE(substr(this_line, 1, 7) == "version")) {
+    return(substr(this_line, 12, nchar(this_line) - 1))
+  }
   warning("failed to find RUST_POLARS_VERSION version")
   "unknown"
 })()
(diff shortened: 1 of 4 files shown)


R/info.R
```diff
@@ -1,3 +1,18 @@
+# get version of rust-polars version from Cargo.lock at package build time
+# from polars 0.31.1 this can be migrated to rust side see
+# https://github.com/pola-rs/polars/pull/9660
+RUST_POLARS_VERSION = (\() {
+  Cargo.lock = readLines("./src/rust/Cargo.lock")
```
eitsupi, line 5: I don't think this will be working user side.

eitsupi (changes requested):
Thanks, but this is needed and will be working?

R/info.R
sorhawell: I think it would work. In what scenario? If all files are deleted after compilation?

sorhawell: It happens at package build time.

etiennebacher: just tried and I can confirm it works:

``` r
library(polars)
list.files(".")
#> [1] "ok-coqui_reprex.R"        "ok-coqui_reprex.spin.R"  
#> [3] "ok-coqui_reprex.spin.Rmd"
pl$polars_info()
#> r-polars package version : 0.7.0.9000
#> rust-polars crate version: 0.30.0
#> 
#> Features:          
#> simd FALSE
```
(also works in another project directory)

eitsupi: Ok, thanks.

eitsupi (approved):

PR #375, opened 2023-09-06
Deprecate `arr` to use `list` instead

sorhawell:
first of two PRs related to #230

PR #377, opened 2023-09-08
Add arm64 cross compilation github release

sorhawell:
Close #4
Close #371

Add source packages containing arm64 Linux or arm64 macOS Rust binary to the GitHub release.

PR #379, opened 2023-09-09
Improve deprecation messages of `arr`

etiennebacher:
Minor improvement following #375. @sorhawell do you want to release 0.8.1 after this one?


sorhawell:
> Minor improvement following #375. @sorhawell do you want to release 0.8.1 after this one?

Probably also get in #377 and an extra #377_2 to split up chores and cross compilation. But then we release 8.1 :)

sorhawell (approved):

PR #382, opened 2023-09-09
chore: arr-list news + fmt

sorhawell:
This is the chore part of #377. After merging this PR, PR #377 will only have changes related to mac cross compilation.


eitsupi (approved):
Thanks!

PR #369, opened 2023-08-30
WIP: refactor lit, col, DataFrame

sorhawell:
rewrite new DataFrame to be derived from pl$select() but where strings are literals.
refactor pl$col and pl$lit to rust side and thereby remove alot of calls from rust side to R

PR #385, opened 2023-09-12
tidy + optimize pl$DataFrame, pl$Series

### LINKED ISSUE #336: How to specify dtypes when I create a DataFrame from scratch ?
It would be great to specify dtypes by passing an `schema` argument at the creation of a DataFrame. 
It exists in py-polars implementation [here](https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/index.html#dataframe).

Here's an example:

```r
mydf <- pl$DataFrame(
  col1 = 1L:3L,
  col2 = c(1.023,2.543,3.768)
)
mydf
shape: (3, 2)
┌──────┬───────┐
│ col1 ┆ col2  │
│ ---  ┆ ---   │
│ i32  ┆ f64   │
╞══════╪═══════╡
│ 1    ┆ 1.023 │
│ 2    ┆ 2.543 │
│ 3    ┆ 3.768 │
└──────┴───────┘
```
What I'd like to do:
```r
mydf <- pl$DataFrame(
  col1 = 1L:3L,
  col2 = c("a","b","c"),
  schema = list(
    "col1" = pl$Int8,
    "col2" = pl$Float32
  )
)
```
Maybe I missed something about how to do that. If so, sorry in advance.

sorhawell:
summary of refactoring:

 - remove internal is_DataFrame_data_input() guard from pl$DataFrame. Reason: not needed, error msg further deep in conversions are better. The guard could also prevent some valid conversion some R class which is not a list or a vector.
 
 - remove pl$DataFrame dead args parallel and via_select. These are not documented and does not do anything, and should have been removed with a previous PR. I would argue this should not count as breaking change to remove some exotic dead args.
 
 - add result_minimal, as_polars_series uses an error to signal  internally no impl was found for a given R class or it does not work. `result_minimal` is light weight version of result which skips all the error upgrading/conversion to signal this.
 
 - raise any other error from as_polars_series which was not due to missing implementation.
 
 - pl$Series drop not needed guard `if (inherits(x, "Series")) {...` internal conversion will do this also anyways. Remove `convert_to_fewer_types()` preprocessing which is now covered by `as_polars_series.POSIXlt`.
 

This is a follow up to PR #369 that brought optional support for vctrs or any other classed R object via the s3 method as_polars_seri.

When running with debug I see the implementation calls methods left and right and take 1ms! per column. Really only an issue for very wide tables. However, I propose a cleanup of the implementation which also is a bit faster .25ms per small column or so. I have also removed some old code not needed anymore in pl$DataFrame
```R
 > pl$options$debug_polars(TRUE)
 > df = do.call(cbind,lapply(1:100, \(i_) mtcars))
 > dim(df)
 [1]   32 1100
 > system.time(pl$DataFrame(df))
[TIME? ms]
pl$DataFrame() -> [0.797ms]
pl$lit() -> [4.627ms]
  RPolarsErr$plain() -> [0.823ms]
   .pr$RPolarsErr$new() -> [0.159ms]
RPolarsErr$plain() -> [0.7918ms]
.
. {~1100 repeats...}
.
   .pr$RPolarsErr$new() -> [0.149ms]
RPolarsErr$plain() -> [1.21ms]
   .pr$RPolarsErr$new() -> [0.262ms]
RPolarsErr$plain() -> [0.2382ms]
pl$select() -> [0.7279ms]
   .pr$DataFrame$select() ->    user  system elapsed 
  1.205   0.040   1.546 
 
 #with out debug print
> pl$options() -> > pl$options$debug_polars(FALSE)
[1.562ms]
pl$options() -> [0.3068ms]
pl$set_polars_options() -> 
> system.time(pl$DataFrame(df))
   user  system elapsed 
  0.808   0.013   0.825   
```
and after
```R
> system.time(pl$DataFrame(df))
[9284ms]
pl$DataFrame() -> [9.998ms]
pl$lit() -> [122.3ms]
pl$select() -> [0.5779ms]
   .pr$DataFrame$select() ->    user  system elapsed 
  0.148   0.005   0.149 
```

Close #336


R/dataframe__frame.R
```diff
@@ -174,27 +154,27 @@ pl$DataFrame = function(..., make_names_unique = TRUE, parallel = FALSE, via_sel
...
-      keys = make.unique(keys, sep = "_")
-    } else {
-      stopf(
-        paste(
-          "conflicting column names not allowed:",
-          paste(unique(keys[duplicated(keys)]), collapse = ", ")
+
+  result({
+    # check for conflicting names, to avoid silent overwrite
+    if (any(duplicated(keys))) {
```
etiennebacher, line 160: you can use `anyDuplicated(keys) > 0` instead (that's what `lintr` recommands)

etiennebacher (commented):
Cool, always nice to see the internals become less complicated 😄

etiennebacher:
While you're refactoring `pl$DataFrame()`, do you want to address #336 as well?

sorhawell:
> it could  be some syntactic sugar for `$select(pl$col("col1")$cast(new_dtype_1), ... ,... )` that would be ok simple.

If to support direct conversion from any R type to any polars type. That would be a big rework. Since casting is faster than R-polars conversion, it is probably not worth it.

etiennebacher:
@sorhawell I took a stab at implementing the arg `schema` in `pl$DataFrame()`, I hope it's fine to do it here. If you agree with these changes, I can implement it in `LazyFrame` as well

etiennebacher (approved):
Actually, I forgot `LazyFrame()` was just a simple wrapper. I added a few tests and just waiting for CI to pass, thanks @sorhawell!

PR #391, opened 2023-09-23
Refactor `options` handling (2)

etiennebacher:
Same as #384 but this needs to be merged after 0.8.1 is released so that it integrates 0.9.0 because it's a breaking change


eitsupi (approved):
Thanks!

PR #399, opened 2023-09-26
Implement `$sample()` for `DataFrame`

etiennebacher:


sorhawell:
Thanks @etiennebacher !

This PR as is I think was fine. However I have some minor details.

In the past rust-polars errors where often just converted into a String, but we try to avoid that in any new or refactored code.  For complicated trait-rule-reasons we cannot currently just use the `?`-operator to convert rust-polars  to rpolars errors. We have to use the polars_to_rpolars_err() function, but it will ensure the canonical conversion is used.

avoid stop() outside a result() as it will raise an unstyled error with no context, which can be difficult for user to interpret.

result catches any error and upgrades it, so inside a result clause stop() is fine. Otherwise for Err_plain() is effectively the same outcome with fewer internal steps.
```R
> result(stop("oups")) |> unwrap("in some example_function():")
   0: In R: in some example_function():
   0: During function call [unwrap(result(stop("oups")), "in some example_function():")]
   1: oups
   
> Err_plain("oups") |> unwrap("example_function():")
   0: In R: example_function():
   0: During function call [unwrap(Err_plain("oups"), "example_function():")]
   1: oups
```

all the pcase stuff is not mandatory

sorhawell (approved):

eitsupi:
Thanks!

etiennebacher:
Thanks for finishing this PR @sorhawell 

> In the past rust-polars errors where often just converted into a String, but we try to avoid that in any new or refactored code. For complicated trait-rule-reasons we cannot currently just use the ?-operator to convert rust-polars PolarsError to our r-polars RPolarsError (but likely in future version of rust we can). We have to use the polars_to_rpolars_err() function for now, but it will ensure our canonical conversion is used.

Yes I feel like you explained this to me before but I didn't remember. We should probably refactor functions on the Rust side to use this (I can try to do this for a couple of them and then ask you to check)

PR #402, opened 2023-09-28
Remove deprecated `$with_column()` and subnamespace `$arr`

etiennebacher:
@sorhawell apparently `Expr_list` is defined twice and the first definition is a deprecated function (no idea since when it's deprecated):

https://github.com/pola-rs/r-polars/blob/33a56d998d5b99fff74106fc2483263cfe6c3846/R/expr__expr.R#L4224-L4230

Do you want to remove this now?


sorhawell:
oh yes the expression `<Expr>$list` was deprecated/renamed in py-polars to `<Expr>$implode()` [we followed suit already in r-polars 0.6.0 4 months ago](https://github.com/pola-rs/r-polars/blob/1650738fce443f99b460da432ecaf98b629a8d01/DESCRIPTION)

We accidentially hard removed it from 0.8.1 when renaming arr to list but that was about to some no harm is done.

We can delete Expr_list, yes.

sorhawell (approved):

sorhawell:
Looks good to me :)

Issue #400, opened 2023-09-27
How do I write custom functions? Like for datatable?


jalilsharif:
Hello,

How do I write custom functions for r-polars something like this for data.table?
```R

# 1. Load the data.table package
library(data.table)

# 2. Create a dummy data.table
test <- data.table(a = rnorm(10),
                     b = rnorm(10),
                     c = rnorm(10))

# 3. Compute the standard deviation by row
columns <- colnames(test)

# Using lapply and sapply for more efficiency
test[, SD := sapply(1:.N, function(i) sd(unlist(lapply(.SD, function(col) col[i])))), .SDcols = columns]
test[, Median:= apply(test[,columns, with=FALSE], FUN=median, MAR=1)] # median by row
test[, Mean:= apply(test[,columns, with=FALSE], FUN=mean, MAR=1)] # mean by row
cov <- function(x){
  cov = sd(x)/mean(x)
  return(cov)
}
test[, COV:= apply(test[,columns, with=FALSE], FUN=c_of_var, MAR=1)] # cov by row
```

etiennebacher:
Are you computing `SD` by row? I don't really understand what you're computing there

jalilsharif:
Yes, indeed, the aim is to undertake custom functions by row. 

this can also include custom written functions like:
```
cov <- function(x){
  cov = sd(x)/mean(x)
  return(cov)
}
test[, COV:= apply(test[,columns, with=FALSE], FUN=c_of_var, MAR=1)] # cov by row
```

etiennebacher:
So normally you should be able to create custom functions by using the polars `Expr` syntax:

``` r
library(polars)
foo <- function(x) {
  x$sum() / x$mean()
}
```
If you pass a polars column to the function above, it returns a polars `Expr`:
```r
foo(pl$col("x"))
#> polars Expr: [(col("x").sum()) // (col("x").mean())]
```

Then regarding rowwise operations, you need to use `concat_list()` and then the `list` subnamespace. In the example below I make `FOO` using the custom function and `FOO2` "by hand" so that we can check they correspond:

``` r
library(polars)

dat <- pl$DataFrame(
  a = rnorm(10),
  b = rnorm(10),
  c = rnorm(10)
)

columns <- colnames(dat)

foo <- function(x) {
  x$sum() / x$mean()
}

dat$with_columns(
  Mean = pl$concat_list(pl$col(columns))$list$mean(),
[...]

PR #403, opened 2023-09-28
[WIP] Start implementing `reduce()` and `fold()`

etiennebacher:
I started to look into how to implement those two functions but for now I run into a panic in a sub-thread:

```r
df = pl$DataFrame(mtcars)

df$with_columns(
  pl$fold(
     acc = pl$lit(1), lambda = \(acc, x) acc, exprs = pl$col("mpg", "drat")
  )
)
```
```r
0: In R: in $with_columns()
0: During function call [df$with_columns(pl$reduce(lambda = function(acc, x) acc, exprs = pl$col("mpg",
"drat")))]
1: A polars sub-thread panicked. See panic msg, which is likely more informative than this error: Any { .. }
```
and I can't find a way to get the panic message. Also, this error comes from `concurrent_handler()` in Rust but I don't understand when this function is actually called.

---

@sorhawell this is mostly me dabbling with Rust so this might be completely off. Let me know if you want to rewrite that from scratch, otherwise I'd greatly appreciate some pointers on this if you have the time 😄

### DIFF
--- a/R/extendr-wrappers.R
+++ b/R/extendr-wrappers.R
@@ -35,6 +35,10 @@ concat_list <- function(exprs) .Call(wrap__concat_list, exprs)
 
 concat_str <- function(dotdotdot, separator) .Call(wrap__concat_str, dotdotdot, separator)
 
+fold <- function(acc, lambda, exprs) .Call(wrap__fold, acc, lambda, exprs)
+
+reduce <- function(lambda, exprs) .Call(wrap__reduce, lambda, exprs)
+
 r_date_range_lazy <- function(start, end, every, closed, time_unit, time_zone, explode) .Call(wrap__r_date_range_lazy, start, end, every, closed, time_unit, time_zone, explode)
 
 as_struct <- function(exprs) .Call(wrap__as_struct, exprs)
--- a/R/functions__lazy.R
+++ b/R/functions__lazy.R
@@ -937,7 +937,6 @@ pl$rolling_corr = function(a, b, window_size, min_periods = NULL, ddof = 1) {
 pl$fold = function(acc, lambda, exprs) {
   l_expr = lapply(as.list(exprs), wrap_e)
   pra = do.call(construct_ProtoExprArray, l_expr)
-  browser()
   unwrap(fold(acc, lambda, pra))
 }
 
--- a/R/functions__lazy.R
+++ b/R/functions__lazy.R
@@ -907,3 +907,53 @@ pl$rolling_corr = function(a, b, window_size, min_periods = NULL, ddof = 1) {
   }
   .pr$Expr$rolling_corr(a, b, window_size, min_periods, ddof) |> unwrap("in pl$rolling_corr()")
 }
+
+
+
+#' Accumulate over multiple columns horizontally / rowwise with a left fold
+#'
+#' @description
+#' `pl$fold()` and `pl$reduce()` allows one to do rowwise operations. The only
+#' difference between them is that `pl$fold()` has an additional argument (`acc`)
+#' that contains the value that will be initialized when the fold starts.
+#'
+#' @name pl_reduce
+#'
+#' @param lambda Function to apply over the accumulator and the value.
+#' @param exprs Expressions to aggregate over. May also be a wildcard expression.
+#'
+#' @return An expression that will be applied rowwise
+#'
+#' @examples
+#' df = pl$DataFrame(mtcars)
+#'
+#' # Make the rowwise sum of all columns and add 1 to it
+#' df$with_columns(
+#'   pl$fold(
+#'     acc = pl$lit(1), lambda = \(acc, x) acc + x, exprs = pl$col("mpg", "drat")
+#'   )
+#' )
+
+pl$fold = function(acc, lambda, exprs) {
+  l_expr = lapply(as.list(exprs), wrap_e)
+  pra = do.call(construct_ProtoExprArray, l_expr)
+  browser()
+  unwrap(fold(acc, lambda, pra))
+}
+
+
+#' @examples
+#' df = pl$DataFrame(mtcars)
+#'
+#' # Make the rowwise sum of all columns and add 1 to it
+#' df$with_columns(
+#'   pl$reduce(
+#'      lambda = \(acc, x) acc + x, exprs = pl$col("mpg", "drat")
+#'   )
+#' )
+
+pl$reduce = function(lambda, exprs) {
+  l_expr = lapply(as.list(exprs), wrap_e)
+  pra = do.call(construct_ProtoExprArray, l_expr)
+  unwrap(reduce(lambda, pra))
+}
--- a/src/rust/src/rlib.rs
+++ b/src/rust/src/rlib.rs
@@ -283,8 +283,8 @@ fn polars_features() -> List {
 fn fold(acc: &Expr, lambda: Robj, exprs: &ProtoExprArray) -> RResult<Expr> {
     use crate::utils::extendr_concurrent::{InitCell, ThreadCom};
     use polars::prelude::Series;
- 
-    // `fold_exprs()` takes two inputs ("acc" and "exprs"). I must run an R process that 
+
+    // `fold_exprs()` takes two inputs ("acc" and "exprs"). I must run an R process that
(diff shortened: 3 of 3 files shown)


sorhawell:
this polars fold is more dynamic than the macro-like fold i proposed in #400 as this fold here folds over the Series at query time, whereas in #400 we folded over the dead expressions. This fold is more related to map
The difference is this map-like fold, over a macro-like fold, is here it can access the actual values via the Series, but it will also a bit slower if a LOT of exprs to fold over, and likely less optimization is possible because polars cannot see the entire query at once and polars cannot reason about what happens in the R/python code and side effects are possible.

polars and R lambdas / map is pretty hairy stuff. I will try to write it as hint as how I think it should be done


This seems to be the py-polars call stack for fold
https://github.com/pola-rs/polars/blob/py-0.19.5/py-polars/polars/functions/lazy.py#L999-L1104
https://github.com/pola-rs/polars/blob/7c2098cd77bcd25aae1931fc9e3337059d070082/py-polars/src/functions/lazy.rs#L330
https://github.com/pola-rs/polars/blob/7c2098cd77bcd25aae1931fc9e3337059d070082/py-polars/src/map/lazy.rs#L71

py-polars can benefit from pyo3 to interface lambdas, where we our own extendr_concurrent.rs . It looks like you knew that :)

our lambda-interface have 4 "sections"

1 - the core polars-R interface module in extendr_concurrent.rs
2 - [the initialization of ThreadCom](https://github.com/pola-rs/r-polars/blob/33a56d998d5b99fff74106fc2483263cfe6c3846/src/rust/src/lib.rs#L39)
3 - [the more specific derived "host"-side functions for collect/profile/fetch in utils/concurrent.rs](https://github.com/pola-rs/r-polars/blob/33a56d998d5b99fff74106fc2483263cfe6c3846/src/rust/src/concurrent.rs#L37)
4 - [the function "client"-side functions like $map](https://github.com/pola-rs/r-polars/blob/33a56d998d5b99fff74106fc2483263cfe6c3846/src/rust/src/lazy/dsl.rs#L1707)
[and $map calls this ](https://github.com/pola-rs/r-polars/blob/33a56d998d5b99fff74106fc2483263cfe6c3846/src/rust/src/lazy/dsl.rs#L1819
) 

It seems you have used snippets from 2,3,4. You should likely only need something like in 4. Perhaps you may be able to derive $fold from $map() and just piggy-bag all the hairy stuff.

this is how rpolars map looks like
https://github.com/pola-rs/r-polars/blob/33a56d998d5b99fff74106fc2483263cfe6c3846/R/expr__expr.R#L738

The signature of binary_lambda is (Series,Series) -> Series, whereas it is (Series)->(Series) for map. I hope we can fairly cheap reuse the map implementation by using a struct to stitch acc Series together with x Series for input.

etiennebacher:
Thanks for all the details!

> Perhaps you may be able to derive $fold from $map() and just piggy-bag all the hairy stuff.
> ...
> The signature of binary_lambda is (Series,Series) -> Series, whereas it is (Series)->(Series) for map. I hope we can fairly cheap 
reuse the map implementation by using a struct to stitch acc Series together with x Series for input.

This is actually what I tried to do so far. The problem is that, as you say, the `move` call in `map()` has `(Series)->(Series)` while it should be `(Series, Series)->(Series)` in the case of `fold()`. Although I didn't try yet, I think stitching the two inputs `Series` together and passing them as a single arg will be a problem since the R function still expects two inputs.

Therefore, I thought the issue is that the `thread_com` object in `map()` can only take one `Series` because of how it is initialized with `&CONFIG`. This is why I defined my own `FOO` config in `fold()` that accepts two input `Series`. Given the error message in my initial post, it looks like something is wrong in the subthread, either when sending the request or when receiving the results.

sorhawell:
I persued the wrap struct approach. I will try also to look into what your error is exactly. My time is limited. So it might take some days

I have a panic in fold2 because the global threadcom is destroyed if folding over multiple exprs instead of just 1

otherwise it works, not sure why this error. Will try to look more into it

sorhawell:
@etiennebacher  The reason your approach gave errors are:

both host side (collect, fetch...) lambda [called serve_r ](https://github.com/pola-rs/r-polars/blob/0fd635bb0cb554ba9838d06f9a6330ab95428ca4/src/rust/src/concurrent.rs#L16)and client side lambda(map, fold, reduce) must have compatible signatures `fn(Series) -> Seiries`. Even though [concurrent_handler](https://github.com/pola-rs/r-polars/blob/0fd635bb0cb554ba9838d06f9a6330ab95428ca4/src/rust/src/utils/extendr_concurrent.rs#L179) is a generic function we must currently pick a single signature for the entire package. We could now or in future rework it such that the signature was an Enum of various signatures. @Sicheng-Pan used a similar pattern for sending various types IPC R jobs.

Also there can only be one global storage CONFIG where thread can acquire a ThreadCom and send R jobs to main thread otherwise.

sorhawell:
I found a subtle bug in the concurrent handling, where if a user writes a lambda that uses e.g. collect inside the concurrent handler, then a new concurrent handler will start and drop the old one. The result is no new job can be submitted and the query will stop. If all R jobs already were in queue, the error would not manifest. With $fold(), new R job are submitted sequentially, and therefore the error showed. Now the collectors:  collect(), fetch() & profile() [will check if there already is a concurrent handler running.](https://github.com/pola-rs/r-polars/blob/8179e684d505a5c479a525f8f5fc047c3a9929ba/src/rust/src/concurrent.rs#L47)

sorhawell:
@etiennebacher  we need to decide if we wanna go with Enum-signatures over struct-wrappers. Enum has near zero overhead, where struct just quite small. Enum would allow future flexibility. I don't know how clean the Enum-signtures can be implemented, or it would be mentally too abstract. I can give it a try.

sorhawell:
```R
pl$fold2(pl$lit(1:5),\(acc,x) acc + 2L*x, list(pl$lit(5:1), pl$lit(11:15)))$lit_to_s()
polars Series: shape: (5,)
Series: '' [i32]
[
	33
	34
	35
	36
	37
]
```

etiennebacher:
For info the current implementation doesn't work with the simple example from `py-polars`:

```r
df = pl$DataFrame(mtcars)

# Make the rowwise sum of two columns and add 1 to it
df$with_columns(
  pl$fold2(
     acc = pl$lit(1), lambda = \(acc, x) acc + x, exprs = pl$col("mpg", "drat")
  )
)
```
```
0: In R: in $with_columns()
0: During function call [df$with_columns(pl$fold2(acc = pl$lit(1), lambda = function(acc,
x) acc + x, exprs = pl$col("mpg", "drat")))]
lengths don't match: could not create a new dataframe: series "_a" has length 1 while series "mpg" has length 32
```

> we need to decide if we wanna go with Enum-signatures over struct-wrappers. Enum has near zero overhead, where struct just quite small in any practical sense. Enum would allow more future flexibility. I don't know how clean the Enum-signtures can be implemented, or it would be mentally too abstract.

Maybe I'm missing some info, but from what you describe it sounds like `Enum` only has advantages so I'd say we should use it.

> I found a subtle bug in the concurrent handling

Is this bug important outside of this PR? If so, it should probably be fixed in a separate smaller PR because this one might take a while to be merged.

sorhawell:
Makes sense I will separate out bugfix and enum-signature in one or two PRs.

sorhawell:
fold reduce seems to work well now

etiennebacher (commented):
Thanks for taking over this PR, @sorhawell. This looks good to me but I can't find a nice way to adapt the rowwise-median example of #400. Do you want to update your answer there and close it at the same time?

Also, I can't approve this PR since I created it

eitsupi (approved):

sorhawell:
> Thanks for taking over this PR, @sorhawell. This looks good to me but I can't find a nice way to adapt the rowwise-median example of #400. Do you want to update your answer there and close it at the same time?
> 
> Also, I can't approve this PR since I created it

I will assign the issue to me. The corresponding data.table example was not very efficient either because OP accessed a single row at the time likely without making use of data.table indexing.

Probably the best solution is to chunk the frame lazily and then transpose, compute medians on 100-1000 rows at the time and then concat back the result. It is likely fairly cache friendly. Any computation which can be performed vectorized can be done with some fold.

Issue #325, opened 2023-07-17
unit test optimization toggles


sorhawell:
follow up on #323 with unit tests when  #324 is merged. [Then `get_current_optimizations()` is available](https://github.com/pola-rs/polars/pull/9883/files).

PR #405, opened 2023-09-29
Unit test optimization toggles from LazyFrame

Sicheng-Pan:
#325

Update nix flake and write the unit test for LazyFrame optimization toggles.


Sicheng-Pan:
Currently the unit test is doing nothing because there is no way to extract optimization toggles after setting them in functions like `collect`

sorhawell:
Hi @Sicheng-Pan  thx for looking into this :) 

how about setting optstate on a lazyframe with internal functions and then check if actually set [with this new getter ](https://github.com/pola-rs/polars/pull/9883/files) ?

Sicheng-Pan:
> Hi @Sicheng-Pan thx for looking into this :)
> 
> how about setting optstate on a lazyframe with internal functions and then check if actually set [with this new getter ](https://github.com/pola-rs/polars/pull/9883/files) ?

@sorhawell Sounds good to me. I've already exported this getter function to R and I will write an internal setter as well. May be we could also make the setter public and allow the users to keep existing optimization settings in functions like `collect`?

sorhawell:
How about having some global option defining if optstate should be set via an external setter or vi collect etc? Then the option could be used in collect.

We should let default behavior match py-polars when there is no clear need otherwise.

Some edge casesnto consider:
What if opt state params is set both by external setter and via collect?

What if two lazyframes are joined, what shouldhappen to optstate?

How does collect know if a lazyframe settings should not be altered, because a user has already set them?

R/lazyframe__lazy.R
```diff
@@ -1185,8 +1238,10 @@ LazyFrame_fetch = function(
     comm_subplan_elim = FALSE
   }
 
-  self |>
-    .pr$LazyFrame$optimization_toggle(
+  if (isTRUE(inherit_optimization)) {
+    self
```
sorhawell, line 1242: this self is not being fetched

R/lazyframe__lazy.R
```diff
@@ -510,39 +558,42 @@ LazyFrame_sink_parquet = function(
...
     projection_pushdown = FALSE
     slice_pushdown = FALSE
   }
-
-  self |>
-    .pr$LazyFrame$optimization_toggle(
+  if (isTRUE(inherit_optimization)) {
+    self
+  } else {
+    self$set_optimization_toggle(
```
sorhawell, line 580: any error in set_optimization_toggle will be raised and escaped the final unwrap

sorhawell: consider using .pr$LazyFrame$set_opt... or wrap entire block in result() to catch any errors

sorhawell (changes requested):
there is a handful of bugs and error handling in flow in LazyFrame__lazy.R
there could be more unit tests to tease out these errors

R/lazyframe__lazy.R
Sicheng-Pan: Fixed.

R/lazyframe__lazy.R
Sicheng-Pan: Fixed.

tests/testthat/test-lazy.R
```diff
@@ -766,3 +766,18 @@ test_that("unnest", {
       to_data_frame()
   )
 })
+
+test_that("opt_toggles", {
+  lf = pl$LazyFrame(mtcars)$select(pl$col("mpg") * 0.42)
```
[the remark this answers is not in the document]
Sicheng-Pan: I've added a few more tests.

R/lazyframe__lazy.R
sorhawell: thx I will review it soon

sorhawell (approved):
Many thanks @Sicheng-Pan  !

PR #407, opened 2023-09-30
add support for pl$concat(<LazyFrame>, . . . ) + add `to_supertypes` auto casting

### LINKED ISSUE #386: `concat()` should work with LazyFrames
The following works with py-polars:
``` r
library(polars)

p1 <- pl$LazyFrame(a = 1, b = 2)
p2 <- pl$LazyFrame(a = 1, b = 2)

pl$concat(list(p1, p2))
#> Error: type of first list element: 'LazyFrame' is not supported
```

sorhawell:
closes #386 

### extras:
allow syntax `pl$concat(. . .)` and not only `pl$concat(list(. . .))`, this is matches behavoir for pl$DataFrame and $select().
to facility unit testing also implemented `<DataFrame>$n_chunks()` to count chunks 
add robj_to! DataFrame and LazyFrame
non-breaking upgrades to internal functions unpack_list() [skip unpack if these classes]  + get_err_ctx() [pick only this context type] + Plain_err() [allow multiline txt input]

### minor deviation from py-polars
This PR deviate slightly from [py-polars pl.concat which is a quite verbose implementation](https://github.com/pola-rs/polars/blob/py-0.19.6/py-polars/polars/functions/eager.py#L22-L221):
how = "align" is not implemented (it could be but it is just another task.)
how = "xyz_releaxed" are just accessed directly via arg `to_supertypes` which will cast differing types to any shared supertype.
r-polars do support `to_supertypes` also for concat Series, py-polars has not implemented that. Maybe that is why they don't just expose `to_supertypes`.

`to_supertypes = TRUE` is like R where `c(1, "b")` yields `[1] "1" "b"` and `to_supertypes = FALSE` will throw errors if not exactly the same type.


src/rust/src/rdataframe/mod.rs
```diff
@@ -78,14 +77,33 @@ impl From<pl::DataFrame> for DataFrame {
         DataFrame(item)
     }
 }
-
+use crate::rpolarserr::*;
```
sorhawell, line 80: I should move this to top

sorhawell: and missing NEWS

R/functions__eager.R
```diff
@@ -3,7 +3,8 @@
 #' @param l list of DataFrame, or Series, LazyFrame or Expr
```
etiennebacher, line 3: Need to redocument because `l` was replaced by `...`

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
 #' # diagonal
 #' pl$concat(l_hor, how = "diagonal")
 pl$concat = function(
```
etiennebacher, line 40: There should be some checks that inputs are of the same class. Currently, `pl$concat(pl$DataFrame(mtcars), pl$LazyFrame(mtcars))` works while it shouldn't. 

`py-polars` errors in this case:
```
pl.concat([pl.DataFrame({"x": 1}), pl.LazyFrame({"x": 1})])

AttributeError: 'LazyFrame' object has no attribute '_df'. Did you mean: '_ldf'?      
```

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
...
+    ..., # list of DataFrames or Series or lazyFrames or expr
     rechunk = TRUE,
-    how = c("vertical", "horizontal", "diagonal"),
-    parallel = TRUE # not used yet
-    ) {
+    how = c("vertical", "horizontal", "diagonal"), # , "vertical_relaxed","diangonal_relaxed"),
+    parallel = TRUE,
+    # eager = FALSE,
+    to_supertypes = FALSE) {
+  if (exists("do_browse", .GlobalEnv) && do_browse) browser()
```
etiennebacher, line 47:
```suggestion

```

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
...
+      }
+      diag_concat_lf(l, rechunk, parallel)
     },
-    inherits(first, "Series"),
+    how == "horizontal" && !eager,
     {
-      stopf("not implemented Series")
+      Err_plain(
+        "how=='horizontal' is not supported for lazy (first element is LazyFrame).",
+        "Try e.g. <LazyFrame>$join() to get Lazy join or pl$concat(lf1$collect(),lf2,lf3).",
```
etiennebacher, line 93:
```suggestion
        "Try e.g. <LazyFrame>$join() to get Lazy join or pl$concat(lf1$collect(), lf2, lf3).",
```

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
...
-    inherits(first, "Expr"),
+    how == "horizontal",
     {
-      stopf("not implemented Expr")
+      if (any(args_modified %in% c("parallel", "to_supertypes"))) {
+        warning(
+          "Args parallel, rechunk, eager and to_supertypes",
+          "takes no effect for how=='horizontal'",
+          call. = FALSE
+        )
```
etiennebacher, line 104:
```suggestion
        warning(
          "Arguments `parallel`, `rechunk`, `eager` and `to_supertypes` are not used when how=='horizontal'",
          call. = FALSE
        )
```

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
...
+      }
+      hor_concat_df(l)
     },
 
     # TODO implement Series, Expr, Lazy etc
-    or_else = stopf(paste0("type of first list element: '", class(first), "' is not supported"))
+    or_else = Err_plain("internal error:", how, "not handled")
```
etiennebacher, line 110: This is never reached because the `match.arg()` above ensures `how` takes appropriate values

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
...
   )
 
-  unwrap(result)
+  # convert back from lazy if eager
+  and_then(Result_out, \(x) {
+    pcase(
+      inherits(x, "DataFrame") && !eager, Err_plain("internal logical error in pl$concat()"),
+      inherits(x, "LazyFrame") && eager, Ok(x$collect()),
```
etiennebacher, line 117: This shouldn't be possible (related to py-polars error above)

R/functions__eager.R
```diff
@@ -37,41 +38,88 @@
...
 pl$concat = function(
-    l, # list of DataFrames or Series or lazyFrames or expr
+    ..., # list of DataFrames or Series or lazyFrames or expr
     rechunk = TRUE,
-    how = c("vertical", "horizontal", "diagonal"),
-    parallel = TRUE # not used yet
-    ) {
+    how = c("vertical", "horizontal", "diagonal"), # , "vertical_relaxed","diangonal_relaxed"),
+    parallel = TRUE,
+    # eager = FALSE,
```
etiennebacher, line 45: should be removed

etiennebacher (changes requested):
Thanks, the Rust side looks good but I have several comments on the R side because for now this doesn't match entirely the behavior of `py-polars`

R/functions__eager.R
sorhawell: I would argue that is a good thing, quite R like, and it does not break anything.

We use robj_to!() to decide if some Robj meaningfully can be coerced into another R/rust object, otherwise an appropriate error is returned/raised.

```R
# cbind a data.frame with a vector in R
> cbind(mtcars, zen_value = runif(32))
                     mpg cyl  disp  hp drat    wt  qsec vs am gear carb  zen_value
Mazda RX4           21.0   6 160.0 110 3.90 2.620 16.46  0  1    4    4 0.94224509
Mazda RX4 Wag       21.0   6 160.0 110 3.90 2.875 17.02  0  1    4    4 0.25997295
Datsun 710          22.8   4 108.0  93 3.85 2.320 18.61  1  1    4    1 0.70630125
Hornet 4 Drive      21.4   6 258.0 110 3.08 3.215 19.44  1  0    3    1 0.75788695
```


```R
# pl$concat DataFrame with Series horizontally
> pl$concat(pl$DataFrame(mtcars),pl$Series(runif(32),"zen-value"),how = "horizontal")
shape: (32, 12)
┌──────┬─────┬───────┬───────┬───┬─────┬──────┬──────┬───────────┐
│ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ … ┆ am  ┆ gear ┆ carb ┆ zen-value │
│ ---  ┆ --- ┆ ---   ┆ ---   ┆   ┆ --- ┆ ---  ┆ ---  ┆ ---       │
│ f64  ┆ f64 ┆ f64   ┆ f64   ┆   ┆ f64 ┆ f64  ┆ f64  ┆ f64       │
╞══════╪═════╪═══════╪═══════╪═══╪═════╪══════╪══════╪═══════════╡
│ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ … ┆ 1.0 ┆ 4.0  ┆ 4.0  ┆ 0.045261  │
```


```R
> pl$concat(pl$LazyFrame(mtcars),mtcars)
[1] "polars LazyFrame naive plan: (run ldf$describe_optimized_plan() to see the optimized plan)"
UNION
  PLAN 0:
    DF ["mpg", "cyl", "disp", "hp"]; PROJECT */11 COLUMNS; SELECTION: "None"
  PLAN 1:
    DF ["mpg", "cyl", "disp", "hp"]; PROJECT */11 COLUMNS; SELECTION: "None"
END UNION
> pl$concat(mtcars, pl$LazyFrame(mtcars))
shape: (64, 11)
┌──────┬─────┬───────┬───────┬───┬─────┬─────┬──────┬──────┐
│ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ … ┆ vs  ┆ am  ┆ gear ┆ carb │
│ ---  ┆ --- ┆ ---   ┆ ---   ┆   ┆ --- ┆ --- ┆ ---  ┆ ---  │
│ f64  ┆ f64 ┆ f64   ┆ f64   ┆   ┆ f64 ┆ f64 ┆ f64  ┆ f64  │
╞══════╪═════╪═══════╪═══════╪═══╪═════╪═════╪══════╪══════╡
│ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ … ┆ 0.0 ┆ 1.0 ┆ 4.0  ┆ 4.0  │
│ 21.0 ┆ 6.0 ┆ 160.0 ┆ 110.0 ┆ … ┆ 0.0 ┆ 1.0 ┆ 4.0  ┆ 4.0  │
│ 22.8 ┆ 4.0 ┆ 108.0 ┆ 93.0  ┆ … ┆ 1.0 ┆ 1.0 ┆ 4.0  ┆ 1.0  │
.....

> rbind(data.table(mtcars),mtcars) |> str()
Classes ‘data.table’ and 'data.frame':	64 obs. of  11 variables:
 $ mpg : num  21 21 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 ...
 $ cyl : num  6 6 4 6 8 6 8 4 4 6 ...
.
.
.
 $ am  : num  1 1 1 0 0 0 0 0 0 0 ...
 $ gear: num  4 4 4 3 3 3 3 4 4 4 ...
 $ carb: num  4 4 1 1 2 1 4 2 2 4 ...
 - attr(*, ".internal.selfref")=<externalptr> 
> rbind(mtcars,data.table(mtcars)) |> str()
'data.frame':	64 obs. of  11 variables:
 $ mpg : num  21 21 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 ...
 $ cyl : num  6 6 4 6 8 6 8 4 4 6 ...
 $ disp: num  160 160 108 258 360 ...

```

R/functions__eager.R
sorhawell: pcase by convention must trigger on something and return something. If nothing else then it runs or_else. Since pl$concat should be able to handle any valid input or_else should be unreachable. Or_else is set to raise an ugly loud "internal error"  if some developer or user manages to pass inputs with no case instead of returning a silent NULL.

maybe even just or_else =`stop("in pl$concat(): internal error:, how, not handled")`

R/functions__eager.R
sorhawell: oups

R/dataframe__frame.R
```diff
@@ -1057,6 +1057,25 @@ DataFrame_first = function() {
   self$lazy()$first()$collect()
 }
 
+
+#' @title Get the number of chunks of the Series' in a DataFrame
```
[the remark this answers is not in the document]
sorhawell: I'm an arse :) I struggled to understand chunkedArrays, and now I don't care to explain to others

R/functions__eager.R
etiennebacher: I thought we should mimic `py-polars` as much as possible. This automatic collecting can lead to unexpected output. For example if I want to concat several LazyFrames at once, but I forgot the first one of them is a DataFrame, then all other LazyFrames are collected, which can take some time. I just think it's weird that this function doesn't only concatenate but collect as well, especially since it's the only one (I think?) that collects automatically.

I'm fine with appending Series to a DataFrame, but I'd rather error if DataFrames and LazyFrames are mixed in the input. Or maybe adding an argument `auto_collect` which would be `FALSE` by default?

R/functions__eager.R
etiennebacher: Ok, I just didn't know enough about `pcase()`

R/functions__eager.R
sorhawell: You're right, there should be no implicit auto collect. I actually disliked that from spark, I was not always sure if a collect would be triggered.

R/dataframe__frame.R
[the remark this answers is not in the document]
etiennebacher: Thanks for adding the explanations, I'll slightly rephrase some stuff when the other comments will be addressed ;)

R/functions__eager.R
```diff
@@ -37,41 +38,89 @@
...
+    to_supertypes = FALSE) {
+  # unpack arg list
+  l = unpack_list(..., skip_classes = "data.frame")
+
+  # nothing becomes NULL
+  if (length(l) == 0L) {
+    return(NULL)
+  }
```
etiennebacher, line 53: I guess we should error instead in this situation, this can only come from a user input misspecification

R/functions__eager.R
sorhawell: it is just a runtime assertion, to avoid a future silent bug in some refactor

R/functions__eager.R
sorhawell: how about when using pl$concat  "dynamically" ?

```R
lf_all_logs_that_day =
  list.files("./error_logs/2022-01-01/", full = TRUE, pattern = "*.parquet") |>
  lapply(pl$scan_parquet) |>
  pl$concat(how = "vertical")
```

it could be there was no error log parquet files on that specific day

```R
> cbind()
NULL
> rbind()
NULL
> 
```

sorhawell: I personally prefer when an API take into account the zero case, because it more tedious to code around. But if you prefer Error, I'll remove it again.

R/functions__eager.R
sorhawell: I have added a comment why it still needs to be there

R/functions__eager.R
etiennebacher: `data.table::rbindlist()` also returns `NULL` when we provide a list of `NULL`s so it's fine to let this as-is

sorhawell:
@etiennebacher  I think we have addressed all first round of comments.

etiennebacher:
Thanks, I think this should be good but I'll just do some tweaks in docs and fix typos

etiennebacher (approved):

PR #409, opened 2023-10-03
Refactor signature

sorhawell:
allow concurrent interface to R have other signatures than fn(Series) -> Series, such as fn(Series, Series) -> Series needed for fold and reduce.

also fix a subtle bug in concurrent R discovered when implementing fold + reduce (cf #403)


src/rust/src/utils/extendr_concurrent.rs
```diff
@@ -234,7 +236,8 @@ where
                 flume::RecvTimeoutError::Timeout => {
                     //check user interrupts flags in R in a fast high-level way with Sys.sleep(0)
                     let res_res = extendr_api::eval_string("Sys.sleep(0)");
-                    //dbg!(&res_res);
+                    #[cfg(feature = "rpolars_debug_print")]
+                    dbg!(&res_res);
```
etiennebacher, line 240: should this `dbg!` be commented?

src/rust/src/concurrent.rs
```diff
@@ -11,83 +11,157 @@ use extendr_api::prelude::*;
...
-        extendr_api::error::Error::Other(format!(
-            "provided input is not an R function but a: {:?}",
-            probj.0
-        ))
-    })?;
+// define any possible signature of R lambdas
+#[derive(Debug)]
+pub enum RFnSignature {
+    //... read as function with input 1 Series mapped to output 1 Series
+    FnSeriesTOSeries(ParRObj, pl::Series),
```
etiennebacher, line 18: Is the `TO` on purpose or should it be `FnSeriesToSeries` instead?

etiennebacher (approved):
I just did some minor formatting in docs, and I have a couple of tiny comments on the Rust side

src/rust/src/concurrent.rs
sorhawell: thx :)

PR #412, opened 2023-10-05
Implement `$cat$get_categories()`

etiennebacher:
``` r
library(polars)

df = pl$DataFrame(
  cats = factor(c("z", "z", "k", "a", "b")),
  vals = factor(c(3, 1, 2, 2, 3))
)
df
#> shape: (5, 2)
#> ┌──────┬──────┐
#> │ cats ┆ vals │
#> │ ---  ┆ ---  │
#> │ cat  ┆ cat  │
#> ╞══════╪══════╡
#> │ z    ┆ 3    │
#> │ z    ┆ 1    │
#> │ k    ┆ 2    │
#> │ a    ┆ 2    │
#> │ b    ┆ 3    │
#> └──────┴──────┘

df$select(
  pl$col("cats")$cat$get_categories()
)
#> shape: (4, 1)
#> ┌──────┐
#> │ cats │
#> │ ---  │
#> │ str  │
#> ╞══════╡
#> │ z    │
#> │ k    │
#> │ a    │
#> │ b    │
#> └──────┘
df$select(
  pl$col("vals")$cat$get_categories()
)
#> shape: (3, 1)
#> ┌──────┐
#> │ vals │
#> │ ---  │
#> │ str  │
#> ╞══════╡
#> │ 3    │
#> │ 1    │
#> │ 2    │
#> └──────┘
```


sorhawell (approved):
looks good to me :) thanks

etiennebacher:
Thanks for the small fix. Quick question, how do you know that `get_categories()` is not fallible?

sorhawell:
> Thanks for the small fix. Quick question, how do you know that `get_categories()` is not fallible?

It the rust-polars return type Expr aka. pl::Expr
https://docs.rs/polars/latest/polars/prelude/struct.CategoricalNameSpace.html#method.get_categories

It would return PolarsResult<Expr> if it was fallible. I could in theory also panic but those we just let crash and burn. In generel any r-polars method which take no input and call a rust function/method which returns no result will be infallible. robj_to!() is fallible because user could give an erroneous input.

etiennebacher:
I see, thanks for the explanation, slowly getting more comfortable with the Rust side of `polars` :)

PR #414, opened 2023-10-06
Implement `$write_csv()` for `DataFrame`

etiennebacher:
TODO: more checks for several parameters (`datetime_format`, etc.)


R/dataframe__frame.R
```diff
@@ -1599,3 +1599,87 @@ DataFrame_sample = function(
...
+#'   then quotes will be used even if they aren`t strictly necessary.
+
+# TODO: include "never" when bumping rust-polars to 0.34
+# * `"never"`: This never puts quotes around fields, even if that results in
+#   invalid CSV data (e.g.: by not quoting strings containing the separator).
+
+#' @return
+#' This doesn't return anything but creates a CSV file.
+#' @export
+#' @rdname IO_write_csv
```
etiennebacher, line 1643: This name is on purpose, I have another branch where I create the category "Input/Output" in the reference page

src/rust/src/rdataframe/mod.rs
```diff
@@ -420,7 +422,57 @@ impl DataFrame {
...
+        batch_size: Robj,
+        datetime_format: Robj,
+        date_format: Robj,
+        time_format: Robj,
+        float_precision: Robj,
+        null_value: Robj,
+        quote_style: Robj,        
+    ) -> RResult<()> {
+
+        let null = robj_to!(Option, String, null_value).unwrap_or_default().unwrap();
```
sorhawell, line 442: you may want to change this to `robj_to!(string, null_value)?` then NULL is automatically prohibited and return a fine error, and "" is already the default from R.

tests/testthat/test-csv.R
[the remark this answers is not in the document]
etiennebacher: good idea, it should be easier than using write + read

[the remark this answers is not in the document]
etiennebacher: Actually, I don't really see how to do that. It seems that some other packages (like `readr`) test their write functions by writing to a tempfile and then reading it back.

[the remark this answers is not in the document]
eitsupi: How about 6c221fe0611752e7a6b182727ef3b1fb0745f6d1?

R/dataframe__frame.R
```diff
@@ -1680,3 +1680,83 @@ DataFrame_sample = function(
...
+    date_format = NULL,
+    time_format = NULL,
+    float_precision = NULL,
+    null_values = "",
+    quote_style = "necessary"
+) {
+
+  if (file_ext(path) != "csv") {
+    stop("Argument `path` must the path to a CSV file.")
+  }
```
eitsupi, line 1747: I don't understand the need for this.

R/dataframe__frame.R
```diff
@@ -1680,3 +1680,83 @@ DataFrame_sample = function(
...
+
+  if (length(quote_style) == 0 ||
+      !quote_style %in% c("always", "necessary", "non_numeric")) {
+    stop("Argument `quote_style` must be one of 'always', 'necessary', or 'non_numeric'.")
+  }
```
eitsupi, line 1752: See https://github.com/pola-rs/r-polars/pull/424#discussion_r1359319634

I think `match.arg` can generally be used for this kind of processing.
However, with the R polars package, the processing is done on the Rust side, so checking on the R side seems not to be necessary.

@sorhawell Thoughts?

etiennebacher: If I remove this and use `unreachable!` on the rust side, then the error is "Error in .pr$DataFrame$write_csv(self, path, has_header, utf8ToInt(separator), : user function panicked: write_csv"

which I don't find very helpful

tests/testthat/test-csv.R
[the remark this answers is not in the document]
etiennebacher: Looks good, thanks

src/rust/src/rdataframe/mod.rs
```diff
@@ -442,7 +444,57 @@ impl DataFrame {
...
 }
+
+pub fn parse_quote_style(x: Robj) -> QuoteStyle {
+    match robj_to!(Option, String, x).unwrap_or_default().unwrap().as_str() {
+        "always" => QuoteStyle::Always,
+        "necessary" => QuoteStyle::Necessary,
+        "non_numeric" => QuoteStyle::NonNumeric,
+        // "never" is available in rust-polars devel only for now (will be added in 0.34)
+        // "never" => QuoteStyle::Never,
+        _ => panic!("polars internal error: `quote_style` must be 'always', 'necessary' or 'non_numeric'.")
```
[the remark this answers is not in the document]
etiennebacher: I don't see why, in `rust-polars` this might be unreachable but here the user can provide anything they want on the R side so this line can actually be reached

etiennebacher:
> It seems that there is an error with `non_numeric`.

`has_header` is the 2nd arg and takes a boolean, hence the error. Should it be handled differently?

src/rust/src/rdataframe/mod.rs
```diff
@@ -442,7 +444,57 @@ impl DataFrame {
...
+            .with_float_precision(robj_to!(Option, usize, float_precision).unwrap())
+            .with_null_value(null)
+            .with_quote_style(qs)
+            .finish(&mut self.0)
+            .map_err(polars_to_rpolars_err)
+    }
 }
+
+pub fn parse_quote_style(x: Robj) -> QuoteStyle {
+    match robj_to!(Option, String, x).unwrap_or_default().unwrap().as_str() {
```
sorhawell: I would change .unwrap_or_default().unwrap().as_str() to ?.as_str()

wrap Ok(QuoteStyle::Always) and other lines

not use panic! nor unreachable as a use could supply a "wrong_name"

instead use rerr().plain("polars internal error: `quote_style` must be 'always', 'necessary' or 'non_numeric'.") which returns Err(rpolarserr) which is same type as Ok(QuoteStyle) being  RResult<QuotoStyle>.

Then after match arms use ? again to get bubble the error and only work with ok value

etiennebacher:
I added some tests for remaining options but some of your comments are still unclear to me. I mark this as ready for review but feel free to push directly on this branch

sorhawell:
in rust unwrap leads to panic and ? "bubbles" the error very nicely back to R. Parsing user input we should always prefer `?`- operator.

`?` operator is kinda short hand for

```rust
if x.is_err() {
 return( x.into())
} else {
 x.unwrap_unchecked()
}
```

the .into() allows for some errors to be implicitly converted... but we do use that at `let f = std::fs::File::create(path)?;` where std::fs::File errors implicitly converted RPolarsError.

If we were to unwrap parsing of user inputs we will be panicking instead of returning errors. Panic stack traces are forwarded to the StdError prompt no the R prompt. Depending on IDE, we can still read the panic message if it catches the StdError prompt. 

In rust if we really believe some thing should never go on, we can use panic!  and unreachable if we believe it could never come to that. unwrap is kinda cowboy coding and should at least be replaced with .expect("me the dev, says this would never be a None/Err because I know that ...") where we state our run time assertion.

When we don't know if something could be an Err we have to model this with the rust error type system. Any one who calls `fallible_foo() -> RResult<Bar>`cannot forget to address the Result has two possible states. When prototyping it is fine to use unwrap(), but should be replaced with something else. The rust value based error system becomes very powerful when dealing with nested error scenarios. Anyways for 95% of our code we can `?`

R/dataframe__frame.R
sorhawell: we could use match.arg but then it should be match.arg(. . . ) |> result() |> uw() to not let a "match.arg" error fly off without context, and that would be all fine. We need to also match.arg on rust side to convert to the underlying enum, and this matching needs error handling too, so we might as well do it there finally. But we can get back to that.

sorhawell: unreachable is mostly used to assure the compiler some match arm could not possibly be reached, and rarely tell a developer there is some flawed internal logic here.

```rust
pub fn robj_to_utf8_byte(robj: extendr_api::Robj) -> RResult<u8> {
    let mut utf8_byte_iter = robj_to_str(robj)?.as_bytes().iter();
    match (utf8_byte_iter.next(), utf8_byte_iter.next()) {
        (Some(s), None) => Ok(*s),
        (None, None) => rerr().plain("cannot extract single byte from empty string"),
        (Some(_), Some(_)) => rerr().plain("multi byte-string not allowed"),
        (None, Some(_)) => unreachable!("the iter() cannot yield Some after None(depleted)"),
    }
}
```

here I used unreachable because the compiler could not figure some fact about rust-iterators. Sometimes it can figure these things out, but not here. If one is very hardcore one could use unreachable_unchecked() which becomes "undefined behavior" if actually reached. unreachable_unchecked() is a few nanoseconds faster.

etiennebacher: Cool, thanks!

etiennebacher:
Thanks for improving the code @sorhawell

PR #416, opened 2023-10-08
fix: not drop first queue member if drop handler

### LINKED ISSUE #413: RBackGroundPool bug when modifying cap multiple times
In an attempt to show how freely cap can be set whenever in comment in #410 I actually found some bug.
It is really a funky edge case ,but this causes some hang up, and that is not OK.

the combination of cap and n_job to get to this bug is quite specific. Bug found on 4 core machine with 8 polars thread workers.

```R
> library(polars)
> pl$get_global_rpool_cap()
$available
[1] 0

$capacity
[1] 4

> pl$set_global_rpool_cap(12)
> l_expr = lapply(1:15,\(i) pl$lit(i)$map(\(x) {Sys.sleep(i/5); -i}, in_background = TRUE)$alias(paste0("lit_",i)))
> lf = pl$LazyFrame()$select(l_expr)
> unlist(pl$get_global_rpool_cap()) |> (\(x) paste(names(x),unname(x)))() |> print()
[1] "available 0" "capacity 12"
> handle = lf$collect(collect_in_background = TRUE)
> for (i in 1:10) {
+   unlist(pl$get_global_rpool_cap()) |> (\(x) paste(names(x),unname(x)))() |> print()
+   pl$set_global_rpool_cap(max(pl$get_global_rpool_cap()$capacity - 1,1))
+   Sys.sleep(.3)
+ }
[1] "available 8" "capacity 12"
[1] "available 8" "capacity 11"
[1] "available 8" "capacity 10"
[1] "available 8" "capacity 9" 
[1] "available 8" "capacity 8" 
[1] "available 7" "capacity 7" 
[1] "available 6" "capacity 6" 
[1] "available 5" "capacity 5" 
[1] "available 4" "capacity 4" 
[1] "available 3" "capacity 3" 
> handle$join()
0: During function call [handle$join()]
1: Disconnected
```

sorhawell:
resolves #413

If limiting cap and thereby active processes, while job queue was not empty, the first queue member would be dropped and never handled.

Solution: Just put first queue member back into queue, if no process is available due to downsizing of cap and active.


sorhawell:
@eitsupi if I want to limit a test to only run in CI and not in interactive dev, because very exotic and takes a few seconds to run, should I do something like  "skip if not en var CI='true' "?

tests/testthat/test-rbackground.R
```diff
@@ -85,3 +85,24 @@ test_that("rpool errors", {
...
+test_that("reduce cap and active while jobs in queue",{
+  pl$set_options(rpool_cap = 3)
+  l_expr = lapply(1:5,\(i) {
+    pl$lit(i)$map(\(x) {Sys.sleep(.4); -i}, in_background = TRUE)$alias(paste0("lit_",i))
+  })
+  lf = pl$LazyFrame()$select(l_expr)
+  handle = lf$collect(collect_in_background = TRUE)
+  Sys.sleep(.2)
+  pl$set_options(rpool_cap = 2)
+  Sys.sleep(.1)
```
eitsupi, line 98: > should I do something like "skip if not en var CI='true' "?

Do you mean this?
I think there is probably no problem with setting it like that.

sorhawell: Yeah that and potentially other process tests. They have some overhead as they spin-up new processes, and do not touch this too often. Any PR check should still run the tests.

eitsupi (approved):
I do not know how to gauge this. (Example in #413 no longer works.)
Since it seems to pass the test, is it safe to merge it?

sorhawell:
It should also work with example in #413 I will check again and then merge. I added a test which should cause the same error , but it is as simple as possible to still cause the original error.

PR #424, opened 2023-10-13
fix devel R now requires no unquoted {} in doc text

sorhawell:


R/expr__expr.R
```diff
@@ -1373,7 +1373,7 @@ Expr_rechunk = "use_extendr_wrapper"
 #' @aliases Expr_cumsum
 #' @name Expr_cumsum
 #' @details
-#' Dtypes in {Int8, UInt8, Int16, UInt16} are cast to
+#' Dtypes in `{Int8, UInt8, Int16, UInt16}` are cast to
```
eitsupi, line 1376: Thanks, but I'm not sure if it's appropriate to express these as "code" here.

How about

```suggestion
#' Dtypes in Int8, UInt8, Int16, UInt16 are cast to
```

sorhawell: makes sense, how about something like this commit for all the braces?
https://github.com/pola-rs/r-polars/pull/424/commits/45eb76638518c2fc23f6f16b07ffc9e0ff2d9847

eitsupi: I think it is good.

In this style it should be possible to check the arguments using the `match.arg` function.
Like this:
https://github.com/apache/arrow/blob/c5bce96ba626ad2558e10acddf8ef19ea06940e8/r/R/dataset-write.R#L125-L136

sorhawell: We don't need match.arg as that is already taken care of in the conversion on rust-side and prints a very nice error, mentioning all valid options. If using match.arg we also have to catch raised errors and forward them

eitsupi: Makes sense, thanks.

sorhawell: The match.arg - like conversion

https://github.com/pola-rs/r-polars/blob/326c370abc88f23e0e7aa611363da98b72617784/src/rust/src/rdatatype.rs#L299

R/expr__expr.R
```diff
@@ -3066,13 +3066,14 @@ Expr_rolling_std = function(
...
-    closed = "left" # ;: Nullable<String>,
+    closed = c("left", "right", "both", "none")
     ) {
   wargs = prepare_rolling_window_args(window_size, min_periods)
-  unwrap(.pr$Expr$rolling_std(
+  .pr$Expr$rolling_std(
     self, wargs$window_size, weights,
-    wargs$min_periods, center, by, closed
-  ))
+    wargs$min_periods, center, by, closed[1L]
```
sorhawell, line 3074: missing before merge, roll out this solution to all. Consider not using [1L] subsetting but simply pass as is to rust side. Likely extendr as_str() don't care if it is a string or a char vec, and will only access first element.

sorhawell: will get back to [1L] in new PR

sorhawell:
a better conversion of R choice/option c("a", "b", "c") similar to match.arg() might arise #437 but that would be out of scope of this PR

eitsupi (approved):
Thanks!

PR #429, opened 2023-10-18
Add arguments to `$profile()`

etiennebacher:
* Add all optimization parameters
* Add arg `show_plot`

TODO:
- [x] more plot-related args: ~`figsize`~, `truncate_nodes`
- [x] bump news

We can't test this with snapshots because the timing changes at every run. I used `ggplot2` because doing the same graph with base R is just a pain, but if one of you wants to replace it with base R I don't have a problem with that.

---

Example:

``` r
library(polars)

countries = c(
  "France", "Germany", "United Kingdom", "Japan", "Columbia",
  "South Korea", "Vietnam", "South Africa", "Senegal", "Iran"
)

set.seed(123)
df_test = pl$LazyFrame(
  country = sample(countries, 1e7, TRUE),
  x = sample(1:100, 1e7, TRUE),
  y = sample(1:1000, 1e7, TRUE)
)$
  with_columns(contains_na = pl$col("country")$str$contains("na"))$
  sort(pl$col("country"))$
  filter(pl$col("country")$is_in(pl$lit(c("United Kingdom", "Japan", "Vietnam")))    )$
  with_columns(is_uk = pl$col("country") == "United Kingdom")

x <- df_test$profile(show_plot = TRUE)
```

``` r

x <- df_test$profile(show_plot = TRUE, truncate_nodes = 5)
```


eitsupi (commented):
Thanks, this is an interesting feature.
While there are other packages out there for creating great plots besides `ggplot2`, it makes sense to go with `ggplot2` here given that it is one of the most popular R package.

Some questions:

- Would it make sense to add an option feature that would allow the display of the plot to be the default?
- I think it might be useful if the ggplot object is the return value, is it possible to do that?

R/lazyframe__lazy.R
```diff
@@ -1397,8 +1403,104 @@ LazyFrame_fetch = function(
...
+      comm_subexpr_elim,
+      streaming
+    ) |> unwrap("in $profile():")
+  }
+
+  out = lf |>
+    .pr$LazyFrame$profile() |>
+    unwrap("in $profile()")
+
+  if (isTRUE(show_plot) && requireNamespace("ggplot2", quietly = TRUE)) {
```
eitsupi, line 1452: Would it be useful to display some kind of warning if ggplot2 is not installed?

etiennebacher:
> Would it make sense to add an option feature that would allow the display of the plot to be the default?

I don't really see the point here

> I think it might be useful if the ggplot object is the return value, is it possible to do that?

We should return a list of 3 with the two tables + the plot when `show_plot = TRUE`, I'll update the function

R/lazyframe__lazy.R
etiennebacher: I added a clearer error message

eitsupi:
> I don't really see the point here

Sorry for the lack of clarity. I mean, I was imagining something like specifying it with the `options` function.

etiennebacher:
> Could you also update the document?

I added details on the returned value

> I was imagining something like specifying it with the `options` function.

I don't think this is necessary, this function shouldn't be called many times so that we need a global option for this

R/lazyframe__lazy.R
```diff
@@ -1397,8 +1404,108 @@ LazyFrame_fetch = function(
...
+      total_timing = paste0(total_timing/1000, "ms")
+      timings$start = timings$start / 1000
+      timings$end = timings$end / 1000
+    } else {
+      unit = "\U00B5s"
+      total_timing = paste0(total_timing, "\U00B5s")
+    }
+
+    # for some reason, there's an error if I use rlang::.data directly in aes()
+    .data = rlang::.data
```
sorhawell, line 1476: If using rlang directly I guess it should be a soft dependency and/or check for it is available together with ggplot2 right?

R/lazyframe__lazy.R
```diff
@@ -1397,8 +1404,108 @@ LazyFrame_fetch = function(
...
+      comm_subexpr_elim,
+      streaming
+    ) |> unwrap("in $profile():")
+  }
+
+  out = lf |>
+    .pr$LazyFrame$profile() |>
+    unwrap("in $profile()")
+
+  if (isTRUE(show_plot)) {
```
sorhawell, line 1453: maybe place plot in an internal R function plot_profile and wrap it in result to catch any errors from ggplot

R/lazyframe__lazy.R
etiennebacher: `rlang` is in Suggests and we don't need to check here because  if `ggplot2` is installed then `rlang` is installed too

R/lazyframe__lazy.R
sorhawell: excellent

sorhawell (approved):
thanks :)

Issue #408, opened 2023-10-01
`arrow` like "R source package with Rust library binary" installation


eitsupi:
The arrow package will attempt to download a pre-built binary from the Internet if `NOT_CRAN=true`.
<https://arrow.apache.org/docs/11.0/r/articles/install.html#r-source-package-with-libarrow-binary>

This method also has the advantage that many features disabled by default (i.e., on CRAN) during source installation are enabled in the pre-built binaries.

I thought the same thing could be done here and have recently added that functionality to `prqlr` (eitsupi/prqlr#195).
The arm64 architecture can also be supported due to cargo's excellent cross-compilation, and the glibc version should be negligible by selecting the musl target. (e.g. #86)

Unlike the binary installation via GitHub releases that this repository currently offers, it has the advantage of being valid for installation from anywhere. Like:

```r
Sys.setenv(NOT_CRAN="true")
remotes::install_github("eitsupi/prqlr")
```

```r
Sys.setenv(NOT_CRAN="true")
install.packages("prqlr", repos = "https://eitsupi.r-universe.dev/") # Of course, Windows and macOS do binary installations
```

For example, the installation on arm64 Linux from R-universe is as follows:

```log
$ docker run --rm -it rocker/r-ver bash
root@3e5a9638fea4:/# R

R version 4.3.1 (2023-06-16) -- "Beagle Scouts"
Copyright (C) 2023 The R Foundation for Statistical Computing
Platform: aarch64-unknown-linux-gnu (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> Sys.setenv(NOT_CRAN="true")
> install.packages("prqlr", repos = "https://eitsupi.r-universe.dev/")
Installing package into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
trying URL 'https://eitsupi.r-universe.dev/src/contrib/prqlr_0.5.3.9000.tar.gz'
Content type 'application/x-gzip' length 191638 bytes (187 KB)
==================================================
downloaded 187 KB

[...]

PR #435, opened 2023-10-20
chore: add configure scripts and update CI

### LINKED ISSUE #431: Rust cache does not seem working on Windows
I was wondering about the long build times on Windows, and it probably has something to do with symlinks not working on Windows.

https://github.com/pola-rs/r-polars/blob/2fc7ff5e5294f9f4fc9d9b8c3ee8a0254476e4cf/src/Makevars.win#L47-L52

eitsupi:
Part of #408, close #431

- Move some logics on Makevars to the configure script
- Rewrite CI workflows
- Drop R 4.1.x support (for not to use old R on Windows)

Port the configure scripts and GitHub Actions workflows from <https://github.com/eitsupi/prqlr>.


eitsupi:
The release workflow seems to be building twice. It needs to be fixed.

eitsupi:
Merging this PR should reduce CI execution time due to Rust caching.
I would like to do a major rewrite of the CI after this is merged.

.github/actions/setup/action.yaml
```diff
@@ -5,9 +5,8 @@ inputs:
     description: Set up Rust nightly toolchain?
     required: false
   target:
-    description: Rust toolchain to use
-    required: false
-    default: ""
+    description: Rust target triple to use
```
etiennebacher, line 8: is "target triple" a typo?

.github/actions/setup/action.yaml
```diff
@@ -43,8 +42,26 @@ runs:
         run: |
           rustup target add ${{ inputs.target }}
 
+      - name: Set up Rust pfofile
```
etiennebacher, line 45:
```suggestion
      - name: Set up Rust profile
```

etiennebacher (commented):
I don't know much in bash so I can't properly review this but reducing CI time would be great. It looks like it still takes about 20min in this PR, is it because it has to rebuild everything before caching is used?

Also, I didn't understand why we bumped the required R version to 4.2. I'd rather keep it 4.1 if there are no reasons other than for CI.

eitsupi:
> It looks like it still takes about 20min in this PR, is it because it has to rebuild everything before caching is used?

I think that's why it hasn't been saved yet because the cache is only saved from the main branch.
Once this is merged and run once, the build time should be significantly faster as the binaries are cached.

> Also, I didn't understand why we bumped the required R version to 4.2. I'd rather keep it 4.1 if there are no reasons other than for CI.

To support R 4.1, the Makevars file for Windows must be split into `Makevars.win` and `Makevars.ucrt`.
Let's stop supporting R 4.1 because the cost of managing two files is high.

Also, the latest version of libR-sys no longer supports R 4.1 (extendr/libR-sys#170), so it will be necessary to end support for R 4.1 in the near future.

.github/actions/setup/action.yaml
eitsupi: I don't understand the intent of your question. I think this is an official term.
<https://doc.rust-lang.org/beta/rustc/platform-support.html#platform-support>

etiennebacher: My bad, I just didn't know this expression

etiennebacher (approved):

.github/workflows/release.yaml
```diff
@@ -78,25 +77,32 @@ jobs:
...
-          echo "RPOLARS_RUST_SOURCE=${PWD}/src/rust" >> $GITHUB_ENV
+          if [ "${{ runner.os }}" == "Windows" ]; then
+            TARGET=x86_64-pc-windows-gnu
+          else
+            TARGET="$(rustc -vV | grep host | cut -d' ' -f2)"
+          fi
+          Rscript -e 'pkgbuild::compile_dll()'
+          mkdir -p tools
+          mv "$(pwd)/src/rust/target/${TARGET}/release/libr_polars.a" tools
+          echo "LIBR_POLARS_PATH=$(pwd)tools/libr_polars.a" >>$GITHUB_ENV
```
eitsupi, line 100:
```suggestion
          echo "LIBR_POLARS_PATH=$(pwd)/tools/libr_polars.a" >>$GITHUB_ENV
```

PR #438, opened 2023-10-20
read_parquet err handling + examples

sorhawell:


R/parquet.R
```diff
@@ -73,9 +96,26 @@ pl$read_parquet = function(
     row_count_name = NULL,
     row_count_offset = 0L,
     low_memory = FALSE) {
+
+  #construct a derived call
   mc = match.call()
```
sorhawell, line 101: @etiennebacher  Interesting using quote + eval instead `...`, I have not thought of that. What would be the pros and cons?

sorhawell: I agree scan_parquet needs a bigger overhaul. I would try to do that in separate PR.

etiennebacher: This quote + eval() was already done in `$read_csv()`, so I just took it there but I'm not the one who implemented it. I can't think of cons for now

sorhawell: @eitsupi what do you think pros and cons of using ... vs match/eval ?

eitsupi: I don't really understand the intent of your question, but I think `match.call` is more convenient here since using `...` as a function argument has the disadvantage that auto-complete cannot be used.

R/parquet.R
```diff
@@ -14,7 +14,23 @@
 #' @name scan_parquet
 #' @rdname IO_scan_parquet
 #' @examples
-#' # TODO write parquet example
+#' #write example file
+#' my_parquet = tempfile(fileext = ".parquet")
```
eitsupi, line 18: Isn't it necessary to add processing to delete temporary files?
Also, is an extension necessary?

R/parquet.R
sorhawell: sry my bad, I blame the fever :)

sorhawell:
superseded by #492

PR #423, opened 2023-10-13
feat! <DataFrame>$transpose

sorhawell:
closes #400

PR #440, opened 2023-10-23
implement <DataFrame>$transpose()

sorhawell:
replaces #423


NEWS.md
```diff
@@ -50,6 +50,7 @@
 - New method `$dt$time()` to extract the time from a `datetime` variable (#428).
 - New method `pl$read_parquet()` that is a shortcut for `pl$scan_parquet()$collect()` (#434).
 - Rename `$str$str_explode()` to `$str$explode()` (#436).
+- New method `pl$transpose()` for `DataFrame` (#440).
```
etiennebacher, line 53:
```suggestion
- New method `$transpose()` for `DataFrame` (#440).
```

R/dataframe__frame.R
```diff
@@ -1682,6 +1682,74 @@ DataFrame_sample = function(
...
+#' @param include_header boolean (default FALSE) if TRUE, add previous column names as
+#' a new Utf8 column.
+#' @param header_name string name of the header column containing previous column names. Only
+#' relevant if include_header = TRUE.
+#' @param column_names character vector or default NULL. How to name previous rows as columns.
+#' NULL is default and is "column_1", "column_2" ... . Length of column names must match n rows of
+#' input DataFrame.
+#'
+#' @details
+#' This is a very expensive operation. Perhaps you can do it differently.
```
etiennebacher, line 1696:
```suggestion
#' This is a very expensive operation.
```

R/dataframe__frame.R
```diff
@@ -1682,6 +1682,74 @@ DataFrame_sample = function(
 }
 
 
+#' Transpose a DataFrame over the diagonal.
+#'
+#' @param include_header boolean (default FALSE) if TRUE, add previous column names as
+#' a new Utf8 column.
```
etiennebacher, line 1688:
```suggestion
#' @param include_header If `TRUE`, the column names will be added as first column.
```

R/dataframe__frame.R
```diff
@@ -1682,6 +1682,74 @@ DataFrame_sample = function(
...
+#' @param header_name string name of the header column containing previous column names. Only
+#' relevant if include_header = TRUE.
```
etiennebacher, line 1690:
```suggestion
#' @param header_name If `include_header` is `TRUE`, this determines the name of the column 
#' that will be inserted.
```

R/dataframe__frame.R
```diff
@@ -1682,6 +1682,74 @@ DataFrame_sample = function(
...
+#' @param column_names character vector or default NULL. How to name previous rows as columns.
+#' NULL is default and is "column_1", "column_2" ... . Length of column names must match n rows of
+#' input DataFrame.
```
etiennebacher, line 1693:
```suggestion
#' @param column_names Character vector indicating the new column names. If `NULL` (default),
#' the columns will be named as "column_1", "column_2", etc. The length of this vector must match
#' the number of rows of the original input.
```

R/dataframe__frame.R
```diff
@@ -1682,6 +1682,74 @@ DataFrame_sample = function(
...
+#' @examples
+#'
+#' # simple use-case
+#' pl$DataFrame(mtcars)$transpose(include_header = TRUE, column_names = rownames(mtcars))
+#'
+#' # All rows must have one shared supertype, recast Categorical to Utf8 which is a supertype
+#' # of f64, and then dataset "Iris" can be transposed
+#' pl$DataFrame(iris)$with_columns(pl$col("Species")$cast(pl$Utf8))$transpose()
+#'
+#' # An example where transpose is warranted for speed.
```
etiennebacher, line 1714: I feel like the whole example below is useful but shouldn't be placed here. It shows a lot of things (chunks, fold, reduce, etc.) that are not related to `$transpose()`. 

Maybe there should be a kind of vignettes like "Useful tips and tricks with polars" where we can gather this kind of examples

etiennebacher (changes requested):

R/dataframe__frame.R
sorhawell: You are probably right. I remove the example.

R/dataframe__frame.R
```diff
@@ -1682,6 +1682,42 @@ DataFrame_sample = function(
...
+#' that will be inserted.
+#' @param column_names Character vector indicating the new column names. If `NULL` (default),
+#' the columns will be named as "column_1", "column_2", etc. The length of this vector must match
+#' the number of rows of the original input.
+#'
+#' @details
+#' This is a very expensive operation.
+#'
+#' If you need to perform non foldable (see fold / reduce) row operations like median.
+#' Transpose may be the fastest option, for that.
```
etiennebacher, line 1698:
```suggestion
#' Transpose may be the fastest option to perform non foldable (see `fold()` or `reduce()`) 
#' row operations like median.
```

etiennebacher (approved):
Just one more change in the docs, otherwise looks good to me, thanks!

PR #452, opened 2023-10-29
docs: updates about pre-built binaries

eitsupi:


eitsupi:
Once this is merged I would like to release 0.9.0 and pin R-universe to the latest release.

README.md
```diff
@@ -106,40 +106,19 @@ install.packages(
...
-``` r
-# Linux
-install.packages(
-  "https://github.com/pola-rs/r-polars/releases/latest/download/polars_cross_aarch64-unknown-linux-gnu.tar.gz",
-  repos = NULL
-)
-```
+For source installation, pre-built Rust libraries may be available if
+the environment variable `NOT_CRAN` is set to `"true"`. (Or, set
+`LIBR_POLARS_BUILD` to `"false"`)
```
etiennebacher, line 113: I suppose that setting `NOT_CRAN` to `"true"` could have some side effects if a package developer installs `polars` in the same environment as the package they develop. Maybe we should prefer setting `LIBR_POLARS_BUILD = "false"` instead?

etiennebacher (approved):
Looks good, I just have a few comments.

Also, does this mean that the original binary size will be greatly reduced since `libr_polars` is obtained afterwards? If so, this is good for the CRAN release, right?

eitsupi:
> Also, does this mean that the original binary size will be greatly reduced since `libr_polars` is obtained afterwards? If so, this is good for the CRAN release, right?

This mechanism should not work on CRAN since CRAN basically does not allow binary downloads from outside.
(It seems that string2path has been criticized by CRAN for doing similar behavior in the past.)

README.md
eitsupi: Yes, `NOT_CRAN` can have meaning in multiple packages, so we should use `LIBR_POLARS_BUILD` to have it affect only the polars package.

Indeed, downloading binary libraries when `NOT_CRAN=ture` is a behavior borrowed from the `arrow` package.

README.Rmd
```diff
@@ -116,45 +116,29 @@ install.packages(
...
-#### Linux (aarch64) / macOS (aarch64)
-
-GitHub releases also provide pre-compiled packages for arm64 (aarch64) platforms.
+### Build from source
 
-These are a little different from above. To install, requiring make (and Xcode for macOS)
-but not rustc / cargo. They are actually source releases bundled with a pre-compiled object file
-`./inst/libr_polars.a`. The final linking / building of any R packages
-must be done on the native OS/architecture, but that should take only ~20 seconds.
+````{asis, echo = file.exists("tools/lib-sums.tsv")}
```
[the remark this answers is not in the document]
eitsupi: Yes, this script will remove the tsv file if the library version in the source package and the latest binary library version on the GitHub release do not match.

https://github.com/pola-rs/r-polars/blob/adb54320f5f682df5ea31489b1915da7b963eee0/dev/generate-lib-sums.R#L32-L42

PR #455, opened 2023-11-01
Refactor `$scan_csv()` and `$read_csv()`

### LINKED ISSUE #444: Scan multiple files
Like this:

``` r
readr::read_csv(c(readr::readr_example("mini-gapminder-africa.csv"), readr::readr_example("mini-gapminder-asia.csv")))
#> Rows: 12 Columns: 5
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr (1): country
#> dbl (4): year, lifeExp, pop, gdpPercap
#>
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
#> # A tibble: 12 × 5
#>    country           year lifeExp       pop gdpPercap
#>    <chr>            <dbl>   <dbl>     <dbl>     <dbl>
#>  1 Algeria           1952    43.1   9279525     2449.
#>  2 Angola            1952    30.0   4232095     3521.
#>  3 Benin             1952    38.2   1738315     1063.
#>  4 Botswana          1952    47.6    442308      851.
#>  5 Burkina Faso      1952    32.0   4469979      543.
#>  6 Burundi           1952    39.0   2445618      339.
#>  7 Afghanistan       1952    28.8   8425333      779.
#>  8 Bahrain           1952    50.9    120447     9867.
#>  9 Bangladesh        1952    37.5  46886859      684.
#> 10 Cambodia          1952    39.4   4693836      368.
#> 11 China             1952    44   556263527      400.
#> 12 Hong Kong, China  1952    61.0   2125900     3054.
```

<sup>Created on 2023-10-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

It seems that `pl$scan_csv()` can only read one file at the moment.

```r
pl$scan_csv(c(readr::readr_example("mini-gapminder-africa.csv"), readr::readr_example("mini-gapminder-asia.csv")))
#> Error in if (!file.exists(path)) { : the condition has length > 1
```

etiennebacher:
Close #444 

This PR:

* updates names of parameters to match py-polars
* adds missing parameters
* deals with multiple paths to CSV files
* improves docs

TODO:

- [x] more tests
- [x] deal with `row_count_` params
- [x] clean Rust side
- [x] deal with multiple download URLs
- [x] clean code for `dtypes` and `null_values` 
- [x] replace the `panic!` related to wrong encoding value
- [x] ~integers are parsed as i64 which can't be converted to R~, see #465

---

Multiple URLs isn't implemented yet in Python, this fails:
```python
import polars as pl

pl.read_csv(
    ["https://vincentarelbundock.github.io/Rdatasets/csv/AER/BenderlyZwick.csv",
     "https://vincentarelbundock.github.io/Rdatasets/csv/AER/BenderlyZwick.csv"]
)
```


R/csv.R
```diff
@@ -136,33 +138,96 @@ pl$scan_csv = function(
...
 }
 
-#' Shortcut for pl$scan_csv()$collect()
+#' New DataFrame from CSV
+#'
+#' @description
+#' Read a file from path into a polars DataFrame.
+#' @name read_csv
 #' @rdname IO_read_csv
+#' @param path Path to a file or URL. It is possible to provide multiple paths
```
eitsupi, line 151: Is it possible to avoid writing the same description twice?
I tried using the `inheritParams` tag but it doesn't seem to work.

Why not avoid duplication by combining them into the same Rd file?

etiennebacher: >  I tried using the inheritParams tag but it doesn't seem to work.

Yeah, I don't know why, that's annoying

> Why not avoid duplication by combining them into the same Rd file?

I'd rather have separate files, but I think I saw somewhere that we could store the roxygen docs as "template" and re-use them, so I'll try to find this again

DESCRIPTION
```diff
@@ -39,6 +40,8 @@ Suggests:
     tibble,
     tools,
     withr
+Remotes:
+    r-lib/testthat
```
eitsupi, line 44: Why is this necessary?

etiennebacher: Because of https://github.com/r-lib/testthat/issues/1880

etiennebacher:
I'm blocked on the `encoding` handling on the Rust side. Using the following code compiles but panicks if I run `pl$read_csv(tmpf, encoding = "foo")` instead of returning the expected error message:
```rust
let encoding = match encoding {
        "utf8" => pl::CsvEncoding::Utf8,
        "utf8-lossy" => pl::CsvEncoding::LossyUtf8,
        e => {
            // panic!("encoding {} not implemented.", e);
            let result = Err(format!("encoding {} not implemented.", e)).unwrap();
            let out = Ok(result)
                .map_err(polars_to_rpolars_err)
                .map(LazyFrame);
            return out
        }
    };
```
@eitsupi can you take a look?

eitsupi:
I will look at it tomorrow if I have time.

eitsupi:
Done.
I think the basic strategy is to notify R of the error message, so we should avoid panic without doing something like `unwrap`.

NEWS.md
```diff
@@ -27,7 +27,14 @@
...
 - New methods `$peak_min()` and `$peak_max()` to find local minima and maxima in
-  a Series.
+  a Series (#462).
+- `$scan_csv()` and `$read_csv()` now match more closely the Python-Polars API (#455):
+  - `sep` is renamed `separator`, `overwrite_dtypes` is renamed `dtypes`, 
+    `parse_dates` is renamed `try_parse_dates`.
+  - new arguments `missing_utf8_is_empty_string`, `rechunk`, `eol_char`, 
+    `raise_if_empty`, `truncate_ragged_lines`
+  - `path` can now be a vector of characters indicating several paths to CSV files. 
+    This only works if all CSV files have the same schema.
```
eitsupi, line 37: Could you change the section since this is a breaking change?

etiennebacher:
Thank you @eitsupi. Two things left:

- integer columns are parsed as i64, which is badly converted to R. One solution is to detect which columns are i64 and shrink they to i32 after reading the file. This can be done either on Rust or R side but will add some overhead. I don't know if there's a way to prevent parsing as i64.
- arg `missing_utf8_is_empty_string` doesn't work (cf commented test) and I don't get why. I tried to remove all `.with_*` lines on Rust side (except `.with_missing_is_null()`) and to hardcode the value as `true` or `false` but it doesn't make any difference. I also didn't find any recent change in upstream polars that would have fixed this behavior after 0.34.

NEWS.md
etiennebacher: done

eitsupi:
> * integer columns are parsed as i64, which is badly converted to R. One solution is to detect which columns are i64 and shrink they to i32 after reading the file. This can be done either on Rust or R side but will add some overhead. I don't know if there's a way to prevent parsing as i64.

I'm not sure how many use cases there are to convert to R data.frame. And, just having the `bit64` R package installed avoids that problem, right?

etiennebacher:
On the R side, we could check that there are i64 columns in the data. If there are some, we check that `bit64` is installed and throw an error if it's not (sort of like `duckplyr`)

eitsupi:
> sort of like `duckplyr`

Maybe you mean `duckdb`?

etiennebacher:
> > sort of like `duckplyr`
> 
> Maybe you mean `duckdb`?

Yes, I misread your previous message

eitsupi:
In any case, these are not directly related to CSV leadership and should be discussed in a separate issue.

eitsupi:
I have created an issue #465 about int64 handling.

etiennebacher:
Thanks, there's just the problem with `missing_utf8_is_empty_string` left before this is ready to review

R/csv.R
etiennebacher: We can include Rmd files in roxygen docs but not in the parameters section, so I don't have a solution for this. I'd still like to keep separate docs so that `scan_csv` and `read_csv` appears separately in the sidebar of "Reference" on the website

etiennebacher:
I don't understand why `missing_utf8_is_empty_string` doesn't work. It shouldn't block this whole PR so I mark this ready for review but I can remove this arg for now if needed

sorhawell:
I could not see either why `missing_utf8_is_empty_string` is not responding. Did it work in py-polars?

sorhawell (approved):

eitsupi:
I'm sorry, but I'm on a business trip and can't let it run on hand for a few days.
Please merge without waiting for me. (I think it's better to just check the operation in Python)

etiennebacher:
@sorhawell yes it works on Python. I decided to remove this argument for now since we know it doesn't work.

@eitsupi no problem, do you have time to release 0.10 after this is merged? Otherwise I can do it if it's just `usethis::use_github_release()`

eitsupi:
> do you have time to release 0.10 after this is merged? Otherwise I can do it if it's just `usethis::use_github_release()`

https://github.com/pola-rs/r-polars/issues/466#issuecomment-1802012570

PR #456, opened 2023-11-01
fix: fix read_* functions when the package is not loaded

eitsupi:
This PR fixes the error `Error in eval(expr, p) : object 'pl' not found` when execute the read functions e.g. `polars::pl$read_csv(path)`.


etiennebacher (approved):
thanks!

PR #457, opened 2023-11-02
feat: add basic SQL features

eitsupi:
There is no `register_globals` setting present in py-polars for now.

By the way, I would like to create a `DBI` backend based on this feature out of interest, but is it okay to create a repository in the rpolars organization? (It is more convenient to exist in the same R-universe)

cc @paleolimbot


eitsupi:
Oops, it seems that this feature can only be built using the nightly-toolchain...

etiennebacher:
> Oops, it seems that this feature can only be built using the nightly-toolchain...

Should I review now?

eitsupi:
> Should I review now?

Thanks for checking this.
Yes, the problem should be fixed.......

NEWS.md
```diff
@@ -21,6 +21,11 @@
   - `$scan_parquet()` gains an argument `hive_partitioning`.
   - `$meta$tree_format()` has a better formatted output.
 
+## What's changed
+
+- New class `RPolarsSQLContext` and its methods.
```
etiennebacher, line 26: Can you add some information on what it allows to do? Something like: 
```suggestion
- New class `RPolarsSQLContext` and its methods to perform any SQL query on a Polars DataFrame or LazyFrame.
```
(not sure what I added is correct)

R/sql.R
```diff
@@ -0,0 +1,167 @@
...
+#' @examplesIf pl$polars_info()$features$sql
+#' lf = pl$LazyFrame(a = 1:3, b = c("x", NA, "z"))
+#' res = pl$SQLContext(frame = lf)$execute(
+#'   "SELECT b, a*2 AS two_a FROM frame WHERE b IS NOT NULL"
+#' )
+#' res$collect()
+RPolarsSQLContext
+
+
+#' @title auto complete $-access into a polars object
```
etiennebacher, line 16:
```suggestion
#' @title Auto complete $-access into a polars object
```

R/sql.R
```diff
@@ -0,0 +1,167 @@
+#' @title Run SQL queries against DataFrame/LazyFrame data.
```
etiennebacher, line 1:
```suggestion
#' Run SQL queries against DataFrame/LazyFrame data.
```
This tag is not actually needed

R/sql.R
```diff
@@ -0,0 +1,167 @@
...
+#' lf = pl$LazyFrame(a = 1:3, b = c("x", NA, "z"))
+#' res = pl$SQLContext(frame = lf)$execute(
+#'   "SELECT b, a*2 AS two_a FROM frame WHERE b IS NOT NULL"
+#' )
+#' res$collect()
+RPolarsSQLContext
+
+
+#' @title auto complete $-access into a polars object
+#' @description called by the interactive R session internally
```
etiennebacher, line 17:
```suggestion
#' @description Called by the interactive R session internally
```

R/sql.R
```diff
@@ -0,0 +1,167 @@
...
+#' @param x RPolarsSQLContext
+#' @param pattern code-stump as string to auto-complete
```
etiennebacher, line 19: what's a code-stump?

R/sql.R
```diff
@@ -0,0 +1,167 @@
...
+#' @return char vec
```
etiennebacher, line 20:
```suggestion
#' @return Character vector
```
Can you clean the rest?

docs/make-docs.R
```diff
@@ -136,7 +136,7 @@ make_doc_hierarchy = function() {
     "pl", "Series", "DataFrame", "LazyFrame", "GroupBy",
     "LazyGroupBy", "ExprList", "ExprBin", "ExprCat", "ExprDT",
     "ExprMeta", "ExprName", "ExprStr", "ExprStruct",
-    "Expr", "IO", "RThreadHandle"
+    "Expr", "IO", "RThreadHandle", "SQLContext"
```
etiennebacher, line 139: "SQL" would be clearer

etiennebacher (approved):
Looks cool, thanks! I personally don't use SQL but there are probably many people who will be happy to see this feature.

Mostly comments about docs

etiennebacher (changes requested):
actually tests fail so I'm "unapproving" for now

docs/make-docs.R
eitsupi: I am not sure how this works, I assumed that SQLContext is the name of the actual structure used upstream and here it is related to the name of the Rd file.

R/sql.R
eitsupi: I believe this is copied from elsewhere.

R/sql.R
eitsupi: ditto

R/sql.R
eitsupi: I believe this is copied from elsewhere.

R/sql.R
eitsupi: I know it is, but I used it because it appeared to be used elsewhere.

eitsupi:
Thanks for your review.
Since `.DollarNames.RPolarsSQLContext` is completely copied from elsewhere, it would be better to revise it with other places instead of this PR.

docs/make-docs.R
etiennebacher: Yeah, I think the change should happen below actually, where we rename the Expr variants

etiennebacher: Not even that, forget about this comment, I'll take care of this later

etiennebacher:
> By the way, I would like to create a DBI backend based on this feature out of interest, but is it okay to create a repository in the rpolars organization?

I don't know but you should open an issue about that

etiennebacher (approved):

PR #464, opened 2023-11-03
Fix docs for `DataFrame$explode()` and `LazyFrame$explode()`

etiennebacher:
Close #461

PR #467, opened 2023-11-04
Add test for `DataFrame$explode()` and `LazyFrame$explode()`

etiennebacher:
For the behavior of `$explode()` on character columns, that might change.

Following https://github.com/pola-rs/r-polars/pull/464#issuecomment-1793191297


eitsupi (approved):
Thanks!

PR #468, opened 2023-11-04
Remove `stopf()`

### LINKED ISSUE #460: Code copied from the data.table package
https://github.com/pola-rs/r-polars/blob/0e730bccb48a0126283235cbe3834f57ae678fcd/R/translation.R#L1-L2

If I understand correctly, the data.table package is licensed under the MPL-2.0 license, so distributing it as an MIT license is a license violation.

etiennebacher:
Close #460


sorhawell:
I don't mind this PR.

Just for context, except from copying from data.table ;) My original intention of the stopf indirection was to
 - have a level of indirection to allow styling of all future errors one place. unwrap() has overtaken that role a lot today.
 - not to use plain stop() as I personally think it has a bad call info behavior. It will as default inform the user of some random internal function, enough to confuse, but not detailed enough to be a full source code trackback as in python.
 
 ```R
 > some_user_f = \() {
+   some_inner_f = \() stopf("oups")
+     some_inner_f()
+ }
> some_user_f()
Error: oups
> some_user_f = \() {
+   some_inner_f = \() stop("oups")
+     some_inner_f()
+ }
> some_user_f()
Error in some_inner_f() : oups
```

sorhawell (approved):

etiennebacher:
Thanks for the additional info, we should remove the use of `stop()` in the future but I'd prefer avoid lagging behind with this PR since it solves an issue

PR #472, opened 2023-11-07
Use last CRAN version of `altdoc`

etiennebacher:
`altdoc` is currently being worked on, which can lead to docs failure here. I made a small release 0.2.2 (not on CRAN yet) that points to before these changes so that we have a stable behavior here.

Merge this when `altdoc` 0.2.2 is on CRAN.


eitsupi (approved):
Looks working fine now.

PR #471, opened 2023-11-06
Implement `$read_ndjson()` and `$scan_ndjson()`

etiennebacher:
TODO:
- [x] tests
- [x] bump news
- [x] ~maybe refactor common bits with csv functions on the Rust side~

PR #474, opened 2023-11-08
Refactor again CSV reading

etiennebacher:
Following #471

### DIFF
--- a/R/csv.R
+++ b/R/csv.R
@@ -101,15 +101,12 @@ pl$scan_csv = function(
   # capture all args and modify some to match lower level function
   args = as.list(environment())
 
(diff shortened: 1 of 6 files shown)


R/csv.R
```diff
@@ -248,6 +246,10 @@ pl$read_csv = function(
 
 
 check_is_link = function(path, reuse_downloaded, raise_error = FALSE) {
+  # do nothing let path fail on rust side
+  if (is.na(path)) {
+    return(NULL)
+  }
```
etiennebacher, line 252: Otherwise, `NA` goes through all the code about connections (without erroring), which is surprisingly slow. Better to have an early exit here

eitsupi (approved):
Thanks!

PR #477, opened 2023-11-08
refactor: auto formatting R files

eitsupi:
I have ran:

```sh
touch R/translation.R
GIT_DIF_TARGET=v0.9.0 make fmt-r
```

### DIFF
--- a/R/csv.R
+++ b/R/csv.R
@@ -97,16 +97,16 @@ pl$scan_csv = function(
     raise_if_empty = TRUE,
     truncate_ragged_lines = FALSE,
     reuse_downloaded = TRUE) {
-
   # capture all args and modify some to match lower level function
   args = as.list(environment())
 
-  args[['path']] = lapply(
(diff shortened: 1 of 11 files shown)


etiennebacher (approved):
Thanks, I think I would lint and style a bit more if the tools were faster, I'm a bit jealous of `ruff` in python ^^
(I'm also not used to using a makefile)

eitsupi:
> Thanks, I think I would lint and style a bit more if the tools were faster, I'm a bit jealous of `ruff` in python ^^

If I understand correctly, the main function of ruff up to now is a linter, and the one used here is an auto formatter, so I think its role is slightly different.
(Of course, ruff has started adding Black features, so there is also an auto formatter feature.)

The problem here is that we are not running the linter on CI. (#214)

etiennebacher:
Yeah they have a formatter now: https://astral.sh/blog/the-ruff-formatter

PR #483, opened 2023-11-09
Add `"never"` in `quote_style`

etiennebacher:


etiennebacher:
@eitsupi there are some failures here with pre-built binaries but I don't understand why

tests/testthat/test-sink_stream.R
```diff
@@ -99,12 +99,12 @@ test_that("sink_csv: separator works", {
...
 
 test_that("sink_csv: quote_style and quote works", {
-  dat_pl2 = pl$LazyFrame(iris)
+  dat_pl2 = pl$LazyFrame(head(iris))
 
   # wrong quote_style
   ctx = dat_pl2$sink_csv(temp_out, quote_style = "foo") |> get_err_ctx()
   expect_identical(ctx$BadArgument, "quote_style")
-  expect_identical(ctx$Plain, "a `quote_style` must be 'always', 'necessary' or 'non_numeric'.")
+  expect_identical(ctx$Plain, "`quote_style` should be one of 'always', 'necessary', 'non_numeric', or 'never'.")
```
eitsupi, line 107: How about use parametrized test?

eitsupi (changes requested):
> there are some failures here with pre-built binaries but I don't understand why

This is caused by incompatibility with pre-built binaries.
Please increment the library version in Cargo.toml and execute `make all` to update the files.

src/rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "r-polars"
-version = "0.34.0"
+version = "0.35.0"
```
eitsupi, line 3: For now, I think 0.34.1 is better because it matches the rust-polars version and is not a breaking change.

eitsupi (changes requested):

tests/testthat/test-sink_stream.R
etiennebacher: Done

src/rust/Cargo.toml
etiennebacher: Done

eitsupi (approved):
Thanks!

PR #484, opened 2023-11-09
ci: use pak's sys reqs installation

eitsupi:
Recent versions of `pak` install system dependencies by default, so setting `CI: false` was meaningless. (See r-lib/rig#184)
And, installing system dependencies by pak does not seem to cause any harm, so we can leave the default settings to simplify configuration.


etiennebacher (approved):
Thanks

PR #489, opened 2023-11-09
Update LICENSE file

stinodego:
Hi all,

I was reviewing the repositories in the Polars organization for license files. I found that the open source license in this repo is non-standard:

`LICENSE` containing only copyright - but no full name included
`LICENSE.md` with an MIT license, again with no name included.

I set the year to 2022 (year of the first commit) and the name to the original author @sorhawell . Of course there have been excellent contributions by many authors since then, but I believe this is the correct way to go about it with regards to licensing. Though I am not sure (not a lawyer). 

The `LICENSE` file is automatically picked up by GitHub. I am not sure the Markdown file is needed, but I don't want to break anything so I left it intact.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,2 +1,21 @@
-YEAR: 2023
-COPYRIGHT HOLDER: polars authors (polars the R package)
+MIT License
+
+Copyright (c) 2022 Søren Havelund Welling
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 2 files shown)


sorhawell (approved):
Hi @stinodego thank you for adding this. Seems good to me.

Sometimes `R CMD check` (auto R package checker) will raise errors if license file is out of R CRAN definition of correctness. It seems our workflow check is not triggered for this PR, not sure why? I have run the check locally and `R CMD check` says ok.

sorhawell:
[oh I see CI check is not triggered by license anymore]( https://github.com/pola-rs/r-polars/blob/6c1f07c6b54a3ddfe6ca03e88189fd890dc20c9f/.github/workflows/check.yaml#L8)

LICENSE
eitsupi, line 1: This change is not allowed.

<https://r-pkgs.org/license.html#key-files> says:

> As described above, the LICENSE file is used in one of two ways. Some licenses are templates that require additional details to be complete in the LICENSE file. The LICENSE file can also contain the full text of non-standard and non-open source licenses. You are not permitted to include the full text of standard licenses.

eitsupi:
Hi @stinodego, thank you for making a PR.
These files are automatically generated by the `usethis` package maintained by Posit PBC.
This package's automatic LICENSE configuration has been in use for many years, and I don't see any need to modify anything other than the year.

PR #491, opened 2023-11-09
Revert "Update LICENSE file"

eitsupi:
Reverts #489

Including the full text of the MIT License in the LICENSE file is not allowed by CRAN.

<https://cran.r-project.org/doc/manuals/r-release/R-exts.html> says:

> One of the “standard” short specifications
> 
> GPL-2 GPL-3 LGPL-2 LGPL-2.1 LGPL-3 AGPL-3 Artistic-2.0
> BSD_2_clause BSD_3_clause MIT

> Whereas you should feel free to include a license file in your source distribution, please do not arrange to install yet another copy of the GNU COPYING or COPYING.LIB files but refer to the copies on https://www.r-project.org/Licenses/ and included in the R distribution (in directory share/licenses). Since files named LICENSE or LICENCE will be installed, do not use these names for standard license files. To include comments about the licensing rather than the body of a license, use a file named something like LICENSE.note.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,2 @@
-MIT License
-
-Copyright (c) 2022 Søren Havelund Welling
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


stinodego (approved):
Right - my apologies. I was a bit too quick to hit the green button. Should have let you sign off & merge.

Will follow your expertise on this!

eitsupi:
Thanks for your quick review!

PR #492, opened 2023-11-09
Refactor and improve `$scan/read_parquet()`

sorhawell:


R/parquet.R
```diff
@@ -85,14 +79,22 @@ pl$read_parquet = function(
...
     row_count_name = NULL,
     row_count_offset = 0L,
-    low_memory = FALSE) {
+    # storage_options,#: dict[str, object] | None = None, #seems fsspec specific
+    low_memory = FALSE,
+    hive_partitioning = TRUE) {
   mc = match.call()
-  mc[[1]] = get("pl", envir = asNamespace("polars"))$scan_parquet
-  eval.parent(mc)$collect()
+  mc[[1]] = pl$scan_parquet
```
eitsupi, line 95:
```suggestion
  mc[[1]] = get("pl", envir = asNamespace("polars"))$scan_parquet
```

If we do not do this, it seems that an error will occur if we use the package without loading it like `polars::pl$read_parquet("foo.parquet")`.

eitsupi (changes requested):

sorhawell:
@eitsupi  sorry for circling around the match.call eval.

I noticed in new utest `"scan read parquet from other process"` that`eval.parent` and not `eval` is needed to scope some args. But using eval.parent makes the frame stack a bit strange, and parent frame could very well be ´R_GlobalEnv´. Maybe that could give some unexpected behavior one day.

How about `do.call(f, args)`?
```R

pl$read_parquet(x,y,z) { 
  args = as.list(environment())
    result({
      do.call(pl$scan_parquet, args)$collect()
    }) |>
      unwrap("in pl$read_parquet(): ")
  }
```

etiennebacher:
> How about do.call(f, args)?

This looks more robust indeed

eitsupi:
> How about `do.call(f, args)`?

Actually I am not familiar with these but they seem to work well, thanks!

eitsupi (approved):

tests/testthat/test-parquet.R
```diff
@@ -0,0 +1,40 @@
...
+    )
+  }
+
+  # bad parallel args
+  ctx = pl$read_parquet(tmpf, parallel = "34") |> get_err_ctx()
+  expect_true(startsWith(ctx$BadValue, "ParallelStrategy choice"))
+  expect_identical(ctx$BadArgument, "parallel")
+  ctx = pl$read_parquet(tmpf, parallel = 42) |> get_err_ctx()
+  expect_identical(ctx$NotAChoice, "input is not a character vector")
+})
```
etiennebacher, line 40: can you split these tests into several `test_that()` calls?

etiennebacher:
Thanks! I'll just tweak a few things before merging

PR #494, opened 2023-11-10
Document S3 methods

### LINKED ISSUE #493: Documentation for base R API (S3 Methods)
There are several functions that can use the same API as base R's `data.frame` etc., but they are not well documented.

etiennebacher:
Close #493 

Screenshot_2023-11-10_at_09-24-23_subset_-_Polars_R_Package


R/s3_methods.R
```diff
@@ -80,131 +88,230 @@
...
 head.DataFrame = function(x, n = 6L, ...) x$limit(n = n)
 
 #' @export
-#' @noRd
+#' @rdname S3_head
 head.LazyFrame = head.DataFrame
 
+#' Take the last n rows
+#'
+#' @param x A Polars DataFrame or LazyFrame
```
eitsupi, line 110:
```suggestion
#' @param x A [DataFrame][DataFrame_class] or [LazyFrame][LazyFrame_class]
```

How about linking to the classes?
Like this.

https://github.com/pola-rs/r-polars/blob/103f95601df3842b7d017517ea4a0b752c22f74a/R/sql.R#L64

R/s3_methods.R
```diff
@@ -80,131 +88,230 @@
...
 
 #' @export
-#' @noRd
+#' @rdname S3_dimnames
 dimnames.LazyFrame = function(x) list(NULL, names(x))
 
+#' Convert to a data.frame
+#'
+#' @param x A Polars DataFrame or LazyFrame
+#' @param ... Any arguments passed to `data.frame()`.
```
eitsupi, line 185:
```suggestion
#' @param ... Any arguments passed to [data.frame()].
```

eitsupi (commented):
Awesome! Thanks!

A few comments about crossreff of roxygen comments.
<https://roxygen2.r-lib.org/articles/index-crossref.html>

R/s3_methods.R
etiennebacher: I'll apply this but not that the links will only appear in RStudio, not on the website

R/s3_methods.R
etiennebacher: ditto

R/s3_methods.R
eitsupi: Yes, but the link should be inserted in the generated Rd file rather than in RStudio.
I think it will be properly linked in the documentation on R-universe.

eitsupi (approved):
Thanks!

R/s3_methods.R
```diff
@@ -80,131 +88,236 @@
...
-#' as.character for polars Series
-#' @param x Series
-#' @param ... Additional arguments are ignored.
-#' @param str_length an integer. If specified,
-#' utf8 or categorical type Series will be formatted to a string of this length.
-#' @return character vector
-#' @examples
-#' s = pl$Series(c("foo", "barbaz"))
-#' as.character(s)
-#' as.character(s, str_length = 3)
```
eitsupi, line 206: @etiennebacher These examples must be removed?

etiennebacher: it was a mistake, I added them back in https://github.com/pola-rs/r-polars/pull/503

Issue #485, opened 2023-11-09
Split README.md to some docs


eitsupi:
- [x] `DEVELOPMENT.md` for developers to develop this package. (#538, #586)
- [x] `vignettes/install.Rmd` for users to install this package. (#643)

etiennebacher:
I'm not convinced we need a vignette for installation instructions, this should appear in the README. Also the procedure is simpler now thanks to your improvements with the binaries, right?

eitsupi:
> I'm not convinced we need a vignette for installation instructions, this should appear in the README.

My intention is something like the arrow package has.
<https://arrow.apache.org/docs/r/articles/install.html>

In my opinion the installation details are too long to write in the README.

etiennebacher:
I'm fine with having a vignette for more detailed instructions, but I'd still want to keep basic installation instructions on the README. I often go to the website homepage just to copy the `install.packages()` line

PR #500, opened 2023-11-11
chore: support binary library 0.34.1 from the GitHub release

eitsupi:


README.md
etiennebacher, line 1: why does this chunk of the README keeps popping on and off? shouldn't it be shown all the time?

etiennebacher (approved):

README.md
eitsupi: My intention was to indicate that certain versions (dev version) do not necessarily support the installation of binary libraries.
Since it is undesirable to update the README as often as you suggest, I would like to move this section to a vignette (#485) and change it so that it is included with the package and appears automatically.......

eitsupi: On second thought, since the R-universe is currently fixed to the latest release version, it would be the same if it always displayed this.

PR #508, opened 2023-11-13
Implement `* _horizontal()` functions

etiennebacher:
Maybe the behaviour of `pl$min()` and `pl$max()` should be updated now since they are now applied on each column independently in py-polars


eitsupi (changes requested):
Awesome, thanks!

Could you also replace `pl$all` and `pl$any`?

etiennebacher:
`pl$all()` selects all columns so it's not the same, and `pl$any()` doesn't exist.

There are `$all()` and `$any()` expressions but they already apply on columns only, not rowwise

eitsupi:
Sorry, I may be mistaken.
I saw py-polars renamed `any` and `all` in another repository a while ago.
https://github.com/eitsupi/querying-with-prql/commit/4dcdfd6b58d7860ae6d29f9299ef1175f309061e

etiennebacher:
Oh yeah `pl.any()` exists but is not implement in r-polars yet

eitsupi:
Can you add examples of "exclude rows with `null` values in any column" and "exclude rows with `null` values in all columns"?
I believe these are very common use cases.

etiennebacher:
Done, is this what you had in mind?

eitsupi (approved):
Yes, thanks again!

Issue #497, opened 2023-11-10
"From Arrow" enhancement


eitsupi:
It was a surprise to me that something like the below didn't work.

```r
polars::pl$DataFrame(arrow::arrow_table(mtcars))
```

I'm wondering if it's possible to create some functions like `as_polars_df` to make converting from more objects more intuitive with S3 methods.

For example, arrow package has `as_arrow_table`,
https://github.com/pola-rs/r-polars/blob/2472c24d22b8a4e8d77b6351c78e60410287b193/R/pkg-nanoarrow.R#L49-L52

eitsupi:
@paleolimbot Sorry to tag you, but any advice on converting from nanoarrow objects to polars DataFrame etc?

paleolimbot:
No worries!

An S3 method is probably helpful, although the S3-ness of it is mostly if you want another package to be able to control that conversion: `as_arrow_table()` is implemented (I think, or if not, it should be) in nanoarrow to collect a `nanoarrow_array_stream` to an `arrow::Table`. I'm not sure that's something you need in rpolars but it's also not hard to do. If you solve `nanoarrow_arrray_stream` to DataFrame, you could leverage the `nanoarrow::as_nanoarrow_array_stream()` generic, which is already implemented for Arrow objects (or should be).

With respect to array stream conversion...you may have to look in the polars code to see if part of this is done already. I believe there's ADBC support somewhere, and ADBC results are also `ArrowArrayStream`, so I imagine there's something already. It may be that all you need is to pass the `nanoarrow_array_stream` object (i.e., external pointer to `ArrowArrayStream` into Rust and cast the pointer to `arrow2::ffi::ArrowArrayStream`.

There is also `nanoarrow::collect_array_stream()`, which will give you a `list()` of `nanoarrow_array` (i.e., external pointer to `ArrowArray`). For complicated reasons, if you go this route you need to call `nanoarrow_pointer_export()` on each array (using a Rust-allocated `ArrowArray` or a `nanoarrow_allocate_array()`).

Let me know if any of that doesn't make sense...I'm also happy to review a PR when you get there!

eitsupi:
Thanks for the detailed explanation!

According to the documentation, py-polars has ADBC integration, so perhaps that is the clue...

sorhawell:
We do actually fully support direct conversion from arrow (Table, RecordBatchReader, ArrowDatum) to polars via arrow2::ffi::ArrowArrayStream. I implemented it here #67. 

[...]

PR #519, opened 2023-11-18
feat!: add `as_polars_df` and `as_polars_lf` generic functions

eitsupi:
Related to #497
Add generic functions for creating DataFrames and LazyFrames.

The PR also includes breaking changes to some existing method arguments, but the probability of having code that depends on the order of these function arguments externally is low.

The default value of `n_rows` as `Inf` is based on dbplyr's `collect` function.


R/as_polars.R
```diff
@@ -0,0 +1,109 @@
...
+
+#' @rdname as_polars_df
+#' @export
+as_polars_df.DataFrame = function(x, ...) {
+  x
+}
+
+
+#' @rdname as_polars_df
+#' @param n_rows Integer or `Inf`
```
eitsupi, line 51: TODO: this line should be removed

R/lazyframe__lazy.R
```diff
@@ -338,6 +338,7 @@ LazyFrame_set_optimization_toggle = function(
 #' @description `$collect()` performs the query on the LazyFrame. It returns a
 #' DataFrame
 #' @inheritParams LazyFrame_set_optimization_toggle
+#' @param ... Ignored.
```
etiennebacher, line 341: I don't understand why you add this here. I get that the `as_polars()` methods must have these dots but nothing forces us to pass them to `$collect()` or `fetch()` in the method for `LazyFrame`

R/as_polars.R
```diff
@@ -0,0 +1,109 @@
...
+#'
+#' For [LazyFrame][LazyFrame_class] objects, this function is a shortcut for
+#' [$collect()][LazyFrame_collect] or [$fetch()][LazyFrame_fetch], depending on
+#' whether the number of rows to fetch is infinite or not.
+#' @rdname as_polars_df
+#' @param x Object to convert to a polars DataFrame.
+#' @param ... Additional arguments passed to methods.
+#' @examples
+#' # Convert a data.frame to a polars LazyFrame
+#' lf = as_polars_df(mtcars)$lazy()
```
etiennebacher, line 16: it would probably have more interest to show the example with the ArrowTable because for classic data.frames `pl$DataFrame()` works fine

tests/testthat/test-as_polars.R
```diff
@@ -0,0 +1,28 @@
...
+  "col_dbl" = (1:10) / 10,
+  "col_chr" = letters[1:10],
+  "col_lgl" = rep_len(c(TRUE, FALSE, NA), 10)
+)
+
+make_cases = function() {
+  tibble::tribble(
+    ~.test_name, ~x,
+    "data.frame", test_df,
+    "pldf", pl$DataFrame(test_df),
```
etiennebacher, line 12: this case is not very useful since it's already used as the expectation

etiennebacher (commented):
Looks good, just a few comments, thanks!

R/lazyframe__lazy.R
eitsupi: The intent is to force these arguments to be specified by name.

etiennebacher (approved):
Thanks, can you also bump NEWS?

eitsupi:
I also added `as_polars_lf()` and Rust macro rewritten with `as_polars_*`.
This causes horizontal concat of multiple unnamed Series or R vectors to fail because they are now converted to DataFrame on the R side when horizontally concatting unnamed Series or R vectors.
However, the Python polars does not allow horizontal concat of Series in the first place, so this is not a problem.

NEWS.md
```diff
@@ -37,6 +37,8 @@
  is aimed for r-polars extensions, and will be kept stable as much as possible (#504).
 - New functions `pl$min_horizontal()`, `pl$max_horizontal()`, `pl$sum_horizontal()`,
   `pl$all_horizontal()`, `pl$any_horizontal()` (#508).
+- New generic functions `as_polars_df()` and `as_polars_lf` to create polars DataFrames
```
etiennebacher, line 40:
```suggestion
- New generic functions `as_polars_df()` and `as_polars_lf()` to create polars DataFrames and LazyFrames
```

etiennebacher (approved):

eitsupi:
As a complete aside, I am surprised that Polars has no way to revert from LazyGroupBy to LazyFrame.
I am impressed that tidypolars seems to have implemented this well.......
(Personally, I think it is safer to rely on the `by` argument instead of using `dplyr::group_by`, as I had trouble dealing with groups even in the `arrow` package.)

etiennebacher:
> I am impressed that tidypolars seems to have implemented this well

Yeah that was annoying. The trick was to almost never use `$group_by()` but to store the groups in the attributes of the Data/LazyFrame and then internally check if there are some groups. It's one of the things I made in a specific PR if you want to take a look: https://github.com/etiennebacher/tidypolars/pull/22

eitsupi:
Looking at the source code, it is not possible to retrieve `opt_state` from LazyGroupBy, but it is almost the same as LazyFrame, so it seems possible to return to LazyFrame by recording `opt_state` on the r-polars side and then writing it back.

https://github.com/pola-rs/polars/blob/5240a35f42467227d9bca894e96ebfb31afe664c/crates/polars-lazy/src/frame/mod.rs#L62-L67
https://github.com/pola-rs/polars/blob/5240a35f42467227d9bca894e96ebfb31afe664c/crates/polars-lazy/src/frame/mod.rs#L1611-L1621

I don't know if it is worth doing.

R/as_polars.R
```diff
@@ -0,0 +1,156 @@
...
+as_polars_df.DataFrame = function(x, ...) {
+  x
+}
+
+
+#' @rdname as_polars_df
+#' @export
+as_polars_df.GroupBy = function(x, ...) {
+  x$to_data_frame()
+}
```
eitsupi, line 56: Sorry, this was incorrect.......
Not polars DataFrame, but R data.frame.
I will open a follow-up PR.

eitsupi: #522

PR #526, opened 2023-11-20
feat: `%/%`, `%%` translation and add `floor_div`, `mod`, `eq_missig` and `neq_missing`

eitsupi:
The order of functions has been slightly changed in the file for readability.

Note that there is currently a problem with Polars' mod (pola-rs/polars#10570).
(I wasn't sure if I should stop subscribing to `%%` because of this problem, but I am anyway)

Also note that the function name are different here than in rust-polars (rem v.s. mod) because `mod` is a reserved word in Rust.

Another problem is that py-polars and rust-polars may have different names for these functions.
For these functions r-polars seems to have adopted many of the ones from rust-polars so far, so I'm going to use `floor_div` and `neq_missing` instead of `floordiv` and `ne_missing` in py-polars.


eitsupi:
The error seems to come from a network failure.

R/expr__expr.R
```diff
@@ -1489,9 +1567,9 @@ Expr_gather = function(indices) {
 #' @examples
 #' pl$DataFrame(a = c(1, 2, 4, 5, 8))$
 #'   with_columns(
-#'     pl$col("a")$shift(-2)$alias("shift-2"),
-#'     pl$col("a")$shift(2)$alias("shift+2")
-#'   )
+#'   pl$col("a")$shift(-2)$alias("shift-2"),
+#'   pl$col("a")$shift(2)$alias("shift+2")
+#' )
```
etiennebacher, line 1572: isn't this indentation wrong? We do want some additional indentation when we are in `with_columns()` (I guess this is `styler`'s fault)

Same thing for all other changes like this

eitsupi: Certainly.
We'll probably need to create an issue in the styler repository.

PR #531, opened 2023-11-21
feat!: rename `$not_` -> `$not`, `$first` -> `$first_distinct`, and add `$last_distinct`

### LINKED ISSUE #524: `$and_()`, `$or_()`, `$not_()` v.s. `$and()`, `$or()`, `$not()`
I noticed that py-polars has `and_()` and `or_()`, not `and()` and `or()`.

We introduced `not_()` in #511.
However, this discussion pola-rs/polars#10807 makes me wonder if the underscore suffix comes from Python supervision and if we need to emulate it in R.
(Not to mention that the function name is rather odd, and I don't think such names are used in common packages)

I feel it would be better to change `not_()` to `not()` before the 0.11 release.

### LINKED ISSUE #527: Rename `is_first` to `is_first_distinct`
Already renamed on the Rust side.

eitsupi:
Close #524
Close #527

- Rename `$not_` to `$not`. (`not` is a reserved word of Python, so py-polars uses `not_`. But we can use `not` in Rust and R)
- Rename `$first` to `$first_distinct`. The remaining `$first` warns.
- Add `$last_distinct`, related to `$first_distinct`.


NEWS.md
```diff
@@ -23,8 +23,10 @@
   columns to these functions will now compute the min/max/sum in each column 
   separately. Use `pl$min_horizontal()` `pl$max_horizontal()`, and 
   `pl$sum_horizontal()` instead for rowwise computation (#508).
-- `$is_not()` is deprecated and will be removed in 0.12.0. Use `$not_()` instead 
-  (#511).
+- `$is_not()` is deprecated and will be removed in 0.12.0. Use `$not()` instead 
+  (#511, #531).
+- `$is_first` is deprecated and will be removed in 0.12.0. Use `$is_first_distinct()`
```
etiennebacher, line 28:
```suggestion
- `$is_first()` is deprecated and will be removed in 0.12.0. Use `$is_first_distinct()`
```

src/rust/src/lazy/dsl.rs
```diff
@@ -1600,8 +1600,7 @@ impl Expr {
         self.0.clone().is_finite().into()
     }
 
-    // TODO: rename to is_first_distinct
-    pub fn is_first(&self) -> Self {
+    pub fn is_first_distinct(&self) -> Self {
```
etiennebacher, line 1603: can you return `RResult<Self>`?

src/rust/src/lazy/dsl.rs
```diff
@@ -1613,7 +1612,9 @@ impl Expr {
         self.0.clone().is_infinite().into()
     }
 
-    // TODO: is_last_distinct
+    pub fn is_last_distinct(&self) -> Self {
```
etiennebacher, line 1615: ditto

src/rust/src/lazy/dsl.rs
eitsupi: There are a large number of other functions that return Self in the same way, so if we want to do that, it would be better to do that in a separate PR.

sorhawell: I think just `Self` is correct where there is no fallibility due to `robj_to` conversion or other error sources. Then the function does not need to wrap return type in a `RResult`.

sorhawell (approved):
Looks good to me. Thx.

src/rust/src/lazy/dsl.rs
etiennebacher: Ah right

PR #533, opened 2023-11-21
refactor!: remove stale, ambiguous, broken functions

eitsupi:
Remove some functions:

- `as_polars_df.RecordBatchReader`: Broken (Sorry, I should add tests in #519. See https://github.com/pola-rs/r-polars/issues/497#issuecomment-1821161702)
- `DataFrame_as_data_frame` (`$as_data_frame`): In favor of `$to_data_frame`
- `GroupBy_to_data_frame` and `GroupBy_as_data_frame`: It is ambiguous whether it is converted to polars DataFrame or R data.frame, and I think it is sufficient to use the new `$ungroup()` method and `as.data.frame()` together.


etiennebacher (approved):
Thanks!

PR #534, opened 2023-11-21
Rename `$apply()` and `$map()` to `$map_elements()` and `$map_batches()`

### LINKED ISSUE #512: Rename the `$map()` and `$apply()` Expr and variants
See at the bottom: https://pola-rs.github.io/polars/releases/upgrade/0.19

etiennebacher:
Close #512 

We haven't implemented `$apply()` for DataFrame or Groupby yet. I also cleaned the docs of both methods.


sorhawell:
Feel free to drop any example from map/apply you don't see necessary. Some of these examples I think take a little too long to run CI-wise.

etiennebacher:
> Some of these examples I think take a little too long to run CI-wise.

Yes but they are still useful I think, maybe I'll move them in a "tips and tricks" vignette with your examples for `$transpose()`

etiennebacher:
@eitsupi sorry I forgot again, should I bump r-polars version in cargo.toml to 0.36.0 or 0.35.1?

eitsupi:
Actually, we are only looking at different version numbers here, so it should not be the same number, but 0.35.1 is a good choice to match the minor version of rust-polars.
https://github.com/pola-rs/r-polars/blob/2fbf67d0d8e4d2bcfcd317873c08c9fb464b0cc2/dev/generate-lib-sums.R#L32

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # polars (development version)
 
+## Breaking changes and deprecations
+
+- `$apply()` on an Expr or a Series is renamed `$map_elements()`, and `$map()` 
+  is renamed `$map_batches()`. `$map()` and `$apply()` will be removed in 0.13.0 (#534).
```
eitsupi, line 6: Wondering if this could be included in 0.11.1 and removed in 0.12.0.

etiennebacher: do you have other things you'd like to include in 0.11.1 apart from that? We release one version every 3-4 weeks so I think we'll get pretty quickly to 0.13.0

eitsupi: I am ok to release 0.11.1 immediately after this is merged as I have nothing further in particular to add.

etiennebacher: My point was that I don't see the need to make a new release just for this deprecation. If you have other things to include then sure but otherwise let's just include this in 0.12.0?

eitsupi (approved):
Thanks!

PR #543, opened 2023-11-26
feat: more imitate the `[` (extract) function for data.frame and vector

eitsupi:
- Rename the Rd file
- Support `x[i]` case for DataFrame and LazyFrame
- Impliment `<Series>[i]`
- Replace internal `stop` to `Err_plain`


etiennebacher (approved):
Thanks

etiennebacher:
This should be merged before #544, right?

eitsupi:
Thanks for taking a look at this. This is a change on the R side only, so either merge first is the same.

PR #516, opened 2023-11-18
WIP: chore: update conversions and error-handling

sorhawell:

PR #554, opened 2023-11-29
refactor!: Rename structs in Rust to rename classes in R

### LINKED ISSUE #149: Open discussion on renaming polars classes

etiennebacher:
Close #149 

I just renamed the LazyFrame class (on R side) and struct (on rust side) to RPolarsLazyFrame. This compiles and works fine. We don't need to change the docs since this is only an internal (breaking) change.

@sorhawell @eitsupi if you agree with this then I can move forward and rename DataFrame and Expr


sorhawell:
This looks very good! I would see it as minor change as some user may have written using class() or inherits() and there have to adjust the class name.

sorhawell:
This looks very good! I would see it as minor change as some user may have written using class() or inherits() and there have to adjust the class name.

Funny thing @paleolimbot suggested this change very early in the project history, and I was like no that would be terrible. I finally came around to like it :)

sorhawell:
This looks very good! I would see it as minor change as some user may have written using class() or inherits() and there have to adjust the class name.

Funny thing @paleolimbot suggested this change very early in the project history, and I was like no that would be terrible. I finally came around to like it :)


If you like you can also split it into a couple of PRs.

sorhawell (dismissed):

etiennebacher:
> I would see it as minor change as some user may have written using class() or inherits() and there have to adjust the class name.

Yes I'll put this in the NEWS

> If you like you can also split it into a couple of PRs.

I think I'll do all at once here

eitsupi:
Thanks for working at this!
I'm definitely in favor of making all class name changes.

etiennebacher:
I'm wondering if I should add the `RPolars` prefix to the other classes too, like `LazyGroupBy` or `DataType`:

https://github.com/pola-rs/r-polars/pull/554/files#diff-afa4237c90018088718e50ed6bf29c5d33bcf94ba925a9a5089683ba258e1343R270-R272

What do you think?

eitsupi:
> I'm wondering if I should add the RPolars prefix to the other classes too

I think so.

src/rust/Cargo.toml
```diff
@@ -127,6 +127,7 @@ features = [
   "search_sorted",
   "serde-lazy",
   "sign",
+  "sql",
```
eitsupi, line 130: Is this intentional change?
The polars/sql feature is already included in the sql feature.

etiennebacher: I thought it would fix the error in the docs CI, I can't reproduce it locally. I'll remove this.

eitsupi: The error message seems come from here.
https://github.com/pola-rs/r-polars/blob/7bbe2c69fcad744a5442b4cff4ae4e071f2230d6/R/dataframe__frame.R#L74

IIUC, this line can changed to be `NULL`

etiennebacher: Thanks, nice catch! Unfortunately there's still one error I can't track down though

eitsupi:
I think we need update these.
https://github.com/pola-rs/r-polars/blob/7bbe2c69fcad744a5442b4cff4ae4e071f2230d6/R/zzz.R#L14-L27

See this one.
https://github.com/pola-rs/r-polars/blob/7bbe2c69fcad744a5442b4cff4ae4e071f2230d6/R/zzz.R#L104

etiennebacher:
> See this one.

That works because you already used `RPolarsSQLContext$` in `extendr-wrappers.R`, cf my previous comment (that we posted at the same time)

eitsupi:
> I haven't renamed all the `DataFrame$` in `extendr-wrappers.R`: https://github.com/pola-rs/r-polars/pull/554/files#diff-d9fa8df226b01f446a1391e800e573e396bdf7b99a514a2ffb2f4d1e9004c740
> 
> That would imply that we need to rename all `DataFrame_` functions to `RPolarsDataFrame_` (and similarly for all `LazyFrame_` and friends). I thought that would be unnecessary because we just want to change the class of the output. Should I replace all `DataFrame$` in `extendr-wrappers.R` too?

I'm sorry, but I don't know what you mean. The file is automatically generated, so you don't have to edit it manually, right?

My understanding is that they come from Rust struct names, so it's funny that currently it's not updated.

eitsupi:
> That works because you already used `RPolarsSQLContext$` in `extendr-wrappers.R`, cf my previous comment (that we posted at the same time)

I haven't edited that file manually. It's just that the Rust struct name is used.

etiennebacher:
I forgot this file is automatically generated, I always added the new methods by hand. My workflow is not the right one, but if I do `rextendr::document()` then `RPolarsSQLContext` doesn't appear in `extendr-wrappers.R`, even if I do `make build`, which is supposed to compile with full features, right?

sorhawell:
> I think we may have to merge #516 first.

I have updated #516 , ready for review again

etiennebacher:
Thanks for completing this PR @eitsupi

PR #558, opened 2023-12-02
Implement `$str$to_date()`, `$str$to_time()`, `$str$to_datetime()`

etiennebacher:
They were already there but not available through the namespace `$str` because they were only used in `$str$strptime()`

Also updated the arguments because some were deprecated/removed upstream


R/expr__string.R
```diff
@@ -84,18 +84,103 @@ ExprStr_strptime = function(
...
+#' @return Expr
+#' @name ExprStr_to_date
+#'
+#' @examples
+#' pl$DataFrame(str_date = c("2009-01-02", "2009-01-03", "2009-1-4", "2009 05 01"))$
+#'   with_columns(date = pl$col("str_date")$str$to_date(strict = FALSE))
+ExprStr_to_date = function(format = NULL, strict = TRUE, exact = TRUE, cache = TRUE) {
+  .pr$Expr$str_to_date(self, format, strict, exact, cache) |>
+    unwrap("in $str$to_date():")
+}
```
eitsupi, line 121: Probably applicable to other functions as well, but such functions may be better off using `...` to enforce named arguments.

eitsupi (approved):
Looks good, thanks!

Just a minor comment.

R/expr__string.R
etiennebacher: but then you lose the description of each param in the docs, no? In any case, if that concerns several other functions it can be done in another PR so I'm merging this

eitsupi: > but then you lose the description of each param in the docs, no?

I'm sorry, but I don't understand. Here's what I mean:

```r
ExprStr_to_date = function(..., format = NULL, strict = TRUE, exact = TRUE, cache = TRUE)
```

etiennebacher: Oh right, could be added later but that could also apply to basically all functions that have several args

PR #559, opened 2023-12-03
test: switch to use the new release version of `testthat`

eitsupi:
We have been using the edge version of `testthat` since https://github.com/pola-rs/r-polars/pull/455#discussion_r1380464052, but `testthat` 3.2.1 is now on CRAN.
So we can switch to that.


etiennebacher (approved):
Thanks!

PR #561, opened 2023-12-03
feat: `rownames` option of `as_polars_df`

eitsupi:
Similar to `tibble::as_tibble`'s `rownames` argument.
However, it differs in that it implicitly overwrites existing column (`as_tibble` holds two columns have same name, and `tibble::rownames_to_column` raise an error).

### DIFF
--- a/R/as_polars.R
+++ b/R/as_polars.R
@@ -12,10 +12,13 @@
 #' @param x Object to convert to a polars DataFrame.
 #' @param ... Additional arguments passed to methods.
 #' @examplesIf requireNamespace("arrow", quietly = TRUE)
-#' at = arrow::as_arrow_table(mtcars)
+#' # Convert the row names of a data frame to a column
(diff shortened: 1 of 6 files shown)


R/as_polars.R
```diff
@@ -36,9 +39,28 @@ as_polars_df.default = function(x, ...) {
 
 
 #' @rdname as_polars_df
+#' @param rownames How to treat existing row names of a data frame:
+#'  - `NULL`: Remove row names. This is the default.
+#'  - A string: The name of a new column, which will contain the row names.
+#'    If `x` already has a column with that name, the existing column will be removed.
```
etiennebacher, line 45: Shouldn't it error instead? I think it's more likely people give the same name by accident rather than knowingly replacing a column. They can just drop the column before converting to DataFrame

eitsupi: I have no strong opinion.
Let's switch to raise error.

PR #562, opened 2023-12-03
refactor: `full_features` is just a short cut of `simd` & `sql`

eitsupi:


etiennebacher (approved):

eitsupi:
> Thanks! I'm actually not sure `$to_titlecase()` requires this flag anymore but to check

I have checked the upstream.
https://github.com/pola-rs/polars/blob/c7c0677d5d74aad99a16d50fe5c43fcce1747744/crates/polars-ops/src/chunked_array/strings/case.rs#L134-L135

PR #563, opened 2023-12-03
ci: set dependabot for GitHub Actions and Rust

eitsupi:


.github/dependabot.yaml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "ci: "
+
+  - package-ecosystem: "cargo"
```
etiennebacher, line 10: how does it work with rust crates? Will it automatically bump the versions?

eitsupi: Yes. This is very widely used.

etiennebacher: But will it also bump things that we want to do manually, like `rust-polars` or `extendr`?

eitsupi: Is there any reason we have to change them manually?
If it's not compatible anyway, the test will just fail.

etiennebacher: Just because updating it comes with a lot of work to adapt our code, so I'd rather open a separate PR for this (as before). But I guess we can also work in the PR opened by the dependabot

etiennebacher (approved):
Thanks! Let's see how this works

PR #573, opened 2023-12-06
DO NOT MERGE: show case auto-completion tricks

sorhawell:
This PR show cases that r-polars with minimal code could support full code completion of `$` even in long chains.

This works because R by default has enabled  code completion for any `[]` function, whereas it is disabled by default for `()` function.

This PR add class to any public function or method `pl_f`

and then adds this generic method 
```R
"[.pl_f" = \(x, ...) {
  if(is.null(formals(x))) {
    x()
  } else {
    x(...)
  }
}
```

to alias `[]`with `()`

this method does not clash with any subsetting as e.g. RPolarsDataFrame is not a function


sorhawell:
#### also works with a native R terminal
image

sorhawell:
@etiennebacher @eitsupi  what do you think of this auto-completion?

eitsupi:
I feel this is great and could be added here as an experimental feature.

etiennebacher:
I like the exploration on this, it's very interesting to see the internals of autocompletion, thanks!

However I'm really not convinced we should add this feature, even labelled "experimental". The polars syntax is already quite strange for classic R users and enabling the use of `[]` for function arguments would be super weird. If we support both kinds of syntax (`()` and `[]`), we will also end up with situations where we mix them in the docs, which will very likely lead to confusion.

It's annoying that we don't have autocompletion after the first `$` but I don't think adding this will make it easier

sorhawell:
@etiennebacher I agree this is not pretty and will look confusing and should not at all be mentioned in examples.

I really like this and want to use it my self though, because it makes scripting so much faster. I can't remember existence or spelling every methods and their arguments.
image

I would argue there is a long way to get completion right and it ok good to ship various experimental feature locked code completions to try them out.

The kind of styling we recommend in examples is not really useful in practice because it does not support method/arg-completion.

e.g.

```R
pl$col("a")$add(52)$sum() 
```
or
```R
pl$
  col("a")$
  add(52)$
  sum() 
```

I find my self break up the polars syntax with assignment while composing to accessing completion, and then rewrite in a stylish way.

```R
# while composing the same expression as above could temporarily look like this
e = pl$col("a")
e = e$add(42)
e = e$sum()
```

I think this is bad practice for any query language, that the user have to write it twice.

sorhawell:
If still a "No", I will ofc accept that :)

etiennebacher:
I get that it makes scripting easier, I would also like to have better autocompletion for polars methods. I just think this issue should be fixed upstream and not with a hack. I also understand that you personally want to use that, but if you share code in issues / PRs / whatever, this will lead to more confusion.

Besides that, introducing this `[.pl_f` has side effects on other packages, e.g `data.table`. When this is enabled, the autocompletion for `data.table` will show all row numbers:

image

Finally, using `$` to chain methods is similar to `R6` so this is not completely unknown in the R world. Using `[` to pass arguments to functions would be unique here and I worry it leads to a lower attractiveness of polars. 

In #580, you linked to [completeme](https://github.com/jimhester/completeme), I suppose you already tried that and it didn't work?

Happy to discuss more about this of course.

---

An additional couple of drawbacks about this implementation, just in case there's a way to fix them (if this gets merged):

* on a moderately sized query, listing all the previous code in the autocomplete can lead to extra wide popup

image

* validating the autocompletion with `ENTER` inserts parenthesis instead of square brackets

PR #577, opened 2023-12-07
fix: `dim()` should return integer

### LINKED ISSUE #576: `dim()` should return integer
Now returns double for RPolarsDataFrame.

```r
> dim(mtcars) |> is.integer()
[1] TRUE

> polars::as_polars_df(mtcars) |> dim() |> is.integer()
[1] FALSE
```

eitsupi:
Fix #576

Because `<DataFrame>.shape()` returns `(usize, usize)` in Rust, it converted to real in R.


etiennebacher:
Should this be updated for the other length / width / ncol ... methods?

eitsupi:
> Should this be updated for the other length / width / ncol ... methods?

From what little I have seen, it seems unnecessary as they are still returning ints at this time.

etiennebacher (approved):
Thanks, can you add a quick note in news?

PR #582, opened 2023-12-08
refactor!: rename `<Expr>$lit_to_s()` to `<Expr>$to_series()` and remove `<Expr>$lit_to_df()`

eitsupi:
- These functions do not exist in py-polars
- Their names differ from the rules for other functions (should be `$to_*` instead of `$lit_to_*`)
- It is ambiguous whether "df" represents data.frame or RPolarsDataFrame

So I think it is better to remove them.


etiennebacher (commented):
This is fine for me but I'd like @sorhawell to take a look too

sorhawell:
I disagree with dropping `lit_to_s()`. 

How about keeping `lit_to_s()` or renaming to `_to_s()`?

I find it very practical and is one of my favorite methods. All the tests become more verbose by using `as_polars_series()` and it is less easy to fast tab-autocomplete and it breaks the polars syntax if required to use `|>` for conversion.


 >  These functions do not exist in py-polars
 
 Neither do `as_polars_series()` right?
 
 > Their names differ from the rules for other functions (should be $to_* instead of $lit_to_*)
 
 That is because only the subvarient Expr::Literal or (Expr derived thereof) can be converted into a Series. I prefixed lit_ point this out.
 
 ```R
 #example of non literal conversion to Series
> as_polars_series(pl$col())

0: In R: in $select()
0: During function call [as_polars_series(pl$col())]
not found:

Error originated just after this operation:
DF []; PROJECT */0 COLUMNS; SELECTION: "None"
```
 
 We can drop the prefix lit, but documentation should state that it will cause an error converting non Literal-derived expressions to Series. We can consider change to `to_s()`. 
 
 

> It is ambiguous whether "df" represents data.frame or RPolarsDataFrame

agreed but that could be resolved by renaming to  `lit_to_xyz`. I don't think lit_to_df is as important as `lit_to_s()`.

**Comment by sorhawell** [comment]
a use case for `lit_to_s()`: If working with a Series, we have no cast. to_lit() and lit_to_s() allows jumping back and forth with method chaining only.
```R
s = pl$Series[1:3]
s$
  to_lit[]$
  # work with a Series as a Literal
  cast[pl$Float64]$
  lit_to_s[]$
  # back to Series interface
  append[4]$
  chunk_lengths()
```

eitsupi:
I see.
Then how about deleting `lit_to_df` and renaming `lit_to_s` to `to_series`?

sorhawell:
> I see. Then how about deleting `lit_to_df` and renaming `lit_to_s` to `to_series`?

sounds good to my ears :)

eitsupi:
@sorhawell I have updated. Could you take a look at this?

sorhawell (approved):
looks good to me :)

PR #583, opened 2023-12-09
feat: better print for RPolarsSQLContext

eitsupi:
Change to show registered tables. Almost the same as below.
https://github.com/rpolars/r-polarssql/blob/0cd1fb3d56d4d76a50bfcbad464a96a42db1e73d/R/show_Connection.R#L4-L11


etiennebacher:
Thanks, can you add a snapshot test for this?

etiennebacher (approved):

PR #538, opened 2023-11-22
docs: add dev doc

eitsupi:
Part of #485

TODO:

- [ ] Remove the development part of the README

PR #586, opened 2023-12-09
docs: remove the development part of the README.md in favor of DEVELOPMENT.md

eitsupi:
Part of #485, a follow up for #538

This PR also remove `renv` related files no longer used.


etiennebacher (approved):
Great, LGTM. Just wondering if @sorhawell still uses the `renv` setup since he set this up I think?

eitsupi:
`renv` became version 1.0.0 a few months ago, but there is no record of it being used here, so I'm pretty sure no one is using it.

PR #587, opened 2023-12-09
docs: some minor fixes for the polars vignette

eitsupi:


etiennebacher (commented):
We could add more details here, like "To avoid collision with classes provided by other packages, the class name of all objects created by polars starts with 'RPolars'. For example, a polars DataFrame has the class 'RPolarsDataFrame'."

eitsupi:
> We could add more details here, like "To avoid collision with classes provided by other packages, the class name of all objects created by polars starts with 'RPolars'. For example, a polars DataFrame has the class 'RPolarsDataFrame'."

I think it would be great to add these explanations. Feel free to edit this PR.

etiennebacher:
I moved the explanation a bit lower in the text

etiennebacher (approved):

eitsupi:
Thanks!

PR #594, opened 2023-12-11
rpolars_style + utils

sorhawell:


eitsupi (changes requested):
Thanks for working on this!
I'm sorry, but I'm on a business trip and can't see this in detail for a while.

For the time being, since the tools directory is for storing files that are used when installing R packages, could you please move this to the dev directory?

sorhawell:
I will try to incorporate this in the current Makefile as a suggestion of how it could be used. I know there might be an upcoming Tasks file to use instead

sorhawell:
besides running like this to style changes, it takes 1-8 seconds
```sh
make fmt-r
```

sometimes every .e.g. .R file could be styled something like like this, it would probably take 1-2 min
```R
source("dev/styler_utils.R")
style_files(list.files(".",full.names=TRUE, pattern = "*.R$"))
```

or using the styler package directly. This could take 2-10 min
```R
source("dev/styler_utils.R")
styler::style_pkg(".", style = rpolars_style)
```

sorhawell:
@eitsupi  I don't know why but it seems check is disabled for some reasone

Styling is now fairly verbose, we can set `verbose = FALSE`

dev/styler_utils.R
```diff
@@ -0,0 +1,171 @@
...
+    pattern = "\\.R$|\\.Rmd$",
+    # specific files never to style
+    excludes = c("R/extendr-wrappers.R")) {
+  # temporarily set GIT_DIF_TARGET
+  old_env = Sys.getenv("GIT_DIF_TARGET")
+  on.exit({
+    Sys.setenv("GIT_DIF_TARGET" = old_env)
+  })
+  if (is.null(branch)) {
+    Sys.setenv("GIT_DIF_TARGET" = "main")
```
sorhawell, line 68: Should touch GIT_DIF_TARGET if branch is NULL

eitsupi (approved):
Seems working fine, thanks!

eitsupi:
@sorhawell If we use this, it seems that `<<-` will be changed to `=`. Could you fix it?

eitsupi:
@sorhawell Could you fix this?

PR #597, opened 2023-12-17
Add experimental custom autocompletion

### LINKED ISSUE #580: better code completion with r-polars
I miss better auto completion in r-polars. This issue tracks attempts to improve.

sorhawell:
Close #580 

- [x]  implement custom Rstudio code completion
- [ ] ~~native R / radian code completion, to be as good as Rstudio completion or what possible.~~
- [ ] ~~support vscode-R completion via their webserver feature~~ 
- [x]  add `pl$polars_code_completion()`

also to support completion:
 - added `$columns()`-method and `names()` for `GroupBy` and `LazyGroupBy`
 - improved "method not found" error message
 - added  missing .DollarNames.RLazyGroupBy  method
 - deprecated `pl$extra_auto_completion()` the old way to activate native R completion.


sorhawell:
have made some progress. It is a bit slow as these Rstudio internals has no public documentation

image

sorhawell:
completions for chains, icons shows if method / sub-namespace ("list") 
image

... and  assignment (columns<-)
image



Auto completions can be used to "tab-start" an expression from a column name
image


works for lazy too

image

sorhawell:
interactive syntax explaining

image

alias is missing an argument
image



image

sorhawell:
@eitsupi @etiennebacher 

It is possible to make very sweet Rstudio auto completion.
I would be possible to also add interactive help-docs or table views of DataFrame.


So far only data.table, dplyr and reticulate and maybe 3 other packages have their completions in the rstudio-source code.

Here I make use the completions methods can be wrapped with custom handlers and injected into the rstudio ide environment. In theory multiple packages could all wrap the same rstudio completion without conflicting.


This approach is using the private api of Rstudio-api, but judging from the blame history. This code part is quite stable on a yearly basis.

sorhawell:
to activate auto-completion run this

```R
source("~/Documents/projs/r-polars/R/rstudio_completion.R")
.dev$activate_polars_rstudio_completion()
library(polars)
```

the weird `.dev`-prefix  would not stay. It is just a convenient way to update completion functions without restarting session or building polars package to do faster try&error.

sorhawell:
We also have some [reasonable  code completion for native R terminals](https://github.com/pola-rs/r-polars/blob/main/R/autocompletion.R) which could be further refined. In the past I was discouraged how bad the native R utils completion worked with Rstudio. But that is no longer a problem, with this custom completion for Rstudio.

etiennebacher:
Love this, the examples look amazing! I'll try a bit more later.

etiennebacher:
That would be cool to have in the next release (not before a few weeks I think). @sorhawell if you're short on time, maybe you can let aside the radian/VS Code objectives and address them in another PR later?

I also think this kind of completion could be enabled by default in RStudio and in the terminal using `.Platform$GUI %in% c("RStudio", "RTerm")` in `.onLoad()`. What do you think?

eitsupi:
> I also think this kind of completion could be enabled by default in RStudio and in the terminal using `.Platform$GUI %in% c("RStudio", "RTerm")` in `.onLoad`. What do you think?

I'm not familiar with these, but perhaps it can be enabled for interactive sessions only?

etiennebacher:
This is what `rstudioapi` uses to detect whether RStudio is used:
``` r
rstudioapi::isAvailable
#> function (version_needed = NULL, child_ok = FALSE) 
#> {
#>     if (child_ok && isJob()) 
#>         return(callRemote(sys.call(), parent.frame()))
#>     identical(.Platform$GUI, "RStudio") && version_ok(version_needed)
#> }
#> <bytecode: 0x00000285110f1b98>
#> <environment: namespace:rstudioapi>
```

sorhawell:
> This is what `rstudioapi` uses to detect whether RStudio is used:
> 
> ```r
> rstudioapi::isAvailable
> #> function (version_needed = NULL, child_ok = FALSE) 
> #> {
> #>     if (child_ok && isJob()) 
> #>         return(callRemote(sys.call(), parent.frame()))
> #>     identical(.Platform$GUI, "RStudio") && version_ok(version_needed)
> #> }
> #> <bytecode: 0x00000285110f1b98>
> #> <environment: namespace:rstudioapi>
> ```

without rstudioapi it is possible to check for if it is an rstudio session like this

```R
"tools:rstudio" %in% search()
```

that is the internal rstudio API which is being patched by the code completion. I think it should always  be actively opt-in rstudio-completion, because patching another API at run-time is brute manners.

etiennebacher:
> I think it should always be actively opt-in rstudio-completion, because patching another API at run-time is brute manners.

Are there some clear downsides in hacking this API? Is the autocompletion for other packages affected by those changes? 

I would be fine with keeping it experimental and opt-in for now (maybe with a message when loading the package?) but I mean at some point we should enable it by default as it changes drastically the experience with `polars`.

eitsupi:
Any update on this?

@etiennebacher Are you willing to take this on?
I am not very interested in this as I do not use RStudio and have no motivation to test it.

R/rstudio_completion.R
```diff
@@ -0,0 +1,222 @@
+# storing autocompletion functions in an environment,
+# allows modification at run-time, without rebuilding package.
+# This is useful as rstudio to do fast trial&error experimentation.
+.dev = new.env(parent = emptyenv())
```
eitsupi, line 4: Why this is named `.dev`?

R/rstudio_completion.R
```diff
@@ -0,0 +1,222 @@
...
+  tryCatch(
+    {
+      base::setTimeLimit(elapsed = timelimit)
+      lhs = eval(parse(text = string), envir = envir)
+      TRUE
+    },
+    error = function(e) {
+      message("could no auto complete syntax because:")
+      if (inherits(e, "RPolarsErr_error")) {
+        message(polars:::result(stop(e))$err$rcall(string))
```
eitsupi, line 20: `polars:::` is needed here?

R/rstudio_completion.R
```diff
@@ -0,0 +1,222 @@
...
+.dev = new.env(parent = emptyenv())
+
+
+# function to parse eval left-hand-side string
+# timelimit: will stop eval attempt and throw error
+.dev$eval_lhs_string = function(string, envir, timelimit = 3) {
+  lhs = NULL
+  tryCatch(
+    {
+      base::setTimeLimit(elapsed = timelimit)
```
eitsupi, line 13: `base::` is needed here?

R/utils.R
```diff
@@ -68,21 +68,20 @@ verify_method_call = function(Class_env, Method_name, call = sys.call(sys.nframe
...
+        "$ - syntax error:", Method_name, "is not a method/attribute of the class", class_name
+        # ,
+        #
+        #         # add call to error messages
+        #         if (!polars_optenv$do_not_repeat_call) {
+        #           paste(
+        #             "\n when calling method:\n",
+        #             paste(capture.output(print(call)), collapse = "\n")
+        #           )
+        #         }
```
eitsupi, line 82: These lines are needed?

eitsupi (changes requested):

etiennebacher:
> @etiennebacher Are you willing to take this on?

Yes I'll finish this one

etiennebacher:
I don't know how to test completion in "native R". I ran vanilla R in bash and in Powershell and couldn't get any completion (for any function, not just `polars`)

R/autocompletion.R
```diff
@@ -1,26 +1,78 @@
...
-#' pl$extra_auto_completion(activate = FALSE) # deactivate
-pl_extra_auto_completion = function(activate = TRUE) {
+#' # Arg + column-name completion
+#' # press tab inside group_by() to see args and/or column names.
+#' pl$LazyFrame(iris)$group_by()
+#'
+#' # deactivate like this or restart R session
+#' pl$code_completion(activate = FALSE)
+#' }
+pl_code_completion = function(
```
eitsupi, line 44: Is there any reason to justify putting this under `pl` env?

eitsupi (commented):
Thanks.

I am wondering if there is any worth in the option to enable/disable this.

etiennebacher:
> I am wondering if there is any worth in the option to enable/disable this.

I wanted to enable this automatically at first, but I'm not confident on how it will interact with other packages that might add their own custom completion (also @sorhawell wrote most of this, I'm not so familiar with the autocompletion internals). I'd rather keep this experimental and opt-in for now

R/autocompletion.R
etiennebacher: Not really, I didn't write this and didn't have a good reason to move it outside of `pl`. Would you prefer naming it `polars_completion()` instead?

sorhawell:
native R terminal has no drop down for multiple suggestions, and there is something annoying where it will
delete what alraady has been written. Right now it works best for the very final completion.

```R
> library(polars)
> pl$code_completion()
Using code completion in 'native' mode.
> # writes this and tab:  pl$col()$ab
> pl$col()$abs() # completes s()
polars Expr: col("").abs()
> rc.status()$comps # use rc.status()$comps for unit testing
[1] "$abs()"
>
```

it works much better with the radian terminal

# writes this and tab:  pl$col()$a
image


image

sorhawell:
I found this for unit testing

```R

for unit testing it is possible to set the 

```R
> utils:::.assignToken("pl$col()$a")
> utils:::.completeToken()
> utils:::.retrieveCompletions()
 [1] "pl$col()$abs()"             "pl$col()$add()"             "pl$col()$agg_groups()"      "pl$col()$alias()"          
 [5] "pl$col()$all()"             "pl$col()$and()"             "pl$col()$any()"             "pl$col()$append()"         
 [9] "pl$col()$approx_n_unique()" "pl$col()$arccos()"          "pl$col()$arccosh()"         "pl$col()$arcsin()"         
[13] "pl$col()$arcsinh()"         "pl$col()$arctan()"          "pl$col()$arctanh()"         "pl$col()$arg_max()"        
[17] "pl$col()$arg_min()"         "pl$col()$arg_sort()"        "pl$col()$arg_unique()"      "pl$col()$argsort()"        
[21] "pl$col()$arr()"   
```

R/autocompletion.R
eitsupi: Hmmm, do you expect the user to execute this function directly?
If not, could a prefix like `.polars_` make sense?

etiennebacher: Yes the user can execute this directly to activate the custom completion since its opt-in for now

R/autocompletion.R
```diff
@@ -1,26 +1,78 @@
...
+#' some time for the eager API depending on the size of the data and of the
+#' query.
+#'
+#' @return NULL
 #'
 #' @examples
-#' # auto completion via .DollarNames method
-#' e = pl$lit(42) # to autocomplete pl$lit(42) save to variable
-#' # then write `e$`  and press tab to see available methods
+#' \dontrun{
```
eitsupi, line 29: How about use `if (interactive())` instead of `dontrun`

R/autocompletion.R
eitsupi: I am not comfortable with the design of the function.
Why not make activation and deactivation separate functions, like `polars_code_completion_activate()` and `polars_code_completion_deactivate()`, respectively?

etiennebacher: Agreed on renaming. One question is how do we know which `mode` is used when we run `polars_deactivate_code_completion`? One solution would be to add a `completion` field in `polars_options()`, so that we (and the user) can always check if autocompletion is activated or not, and if so in which mode. Do you agree with this?

eitsupi: Does the former mode matter when executing those functions?
I think we can display the information, perhaps via `polars_info()`?
(Similar to the one currently showing `thread_pool_size`)

etiennebacher: > Does the former mode matter when executing those functions?

Yes, it's not the same code to deactivate autocompletion depending on the mode

> I think we can display the information, perhaps via polars_info()?

We could display it there, but we still need to store the info somewhere so that `polars_info()` can retrieve it. That's why I thought we could store it as a `polars.autocompletion` option.

eitsupi: Hmmm, I don't know if we can use option that way for preservation of the state.
Wouldn't you generally use environment type objects?

etiennebacher: I don't understand what you mean, here's what I have in mind for the value of `polars.autocompletion` option:

* default is autocompletion disabled, so it's `""` (or `NULL`)
* `polars_activate_code_completion(mode = "X")` =>  `X`
* `polars_activate_code_completion()` => retrieve the mode that was set by the user with `getOption("polars.autocompletion")`, deactivate autocompletion, and set the option to `""` (or `NULL`)

The option is reset when the session is restarted, but the autocompletion would be automatically deactivated so I don't think there's a problem here

etiennebacher: For now, in any case we require the user to run `polars_activate_code_completion()` so we don't have to take envvars into account

eitsupi: The environment I am talking about is not an environment variable.
https://adv-r.hadley.nz/environments.html

Environment is commonly used in the R world to preserve state, right?
I don't think that options set by the user should be updated by the function.

etiennebacher: Oh right, I confused environments and envvars, I'll deal with that tomorrow or beginning of next week

R/autocompletion.R
```diff
@@ -1,26 +1,98 @@
...
-#' pl$extra_auto_completion(activate = FALSE) # deactivate
-pl_extra_auto_completion = function(activate = TRUE) {
+#'   # Arg + column-name completion
+#'   # press tab inside group_by() to see args and/or column names.
+#'   pl$LazyFrame(iris)$group_by()
+#'
+#'   # deactivate like this or restart R session
+#'   polars_deactivate_code_completion()
+#' }
+polars_activate_code_completion = function(
```
eitsupi, line 47: I think a name like `polars_code_completion_*` is better than `polars_*_code_completion` in terms of autocompletion.

etiennebacher: Done

eitsupi (approved):
Greate, let's merge.

PR #598, opened 2023-12-17
feat: allow concat conditions with `,` instead of `&` in `<LazyFrame>$filter` and `pl$when`

eitsupi:


eitsupi:
TODO: Add examples and tests like https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.when.html

etiennebacher (changes requested):
This looks great, thanks! 

Can you add a test (and maybe some docs) on what happens when there's a trailing comma in the `...`? I'm not sure whether it's ignored or if it errors (and if so if the error message is clear)

eitsupi:
> This looks great, thanks!
> 
> Can you add a test (and maybe some docs) on what happens when there's a trailing comma in the `...`? I'm not sure whether it's ignored or if it errors (and if so if the error message is clear)

Currently this results in an error.
The error messages currently don't seem very nice, but perhaps this can be improved in the future.

eitsupi:
I am currently getting the same error with `with_columns`.
The meaning of this error is quite difficult to understand. The easiest way to fix this is to use `rlang::list2`, but we can't do that here, so this error should be allowed.

```r
> pl$DataFrame(mtcars)$with_columns(foo = TRUE,)
Error in unpack_list(...) : argument is missing, with no default
```

NEWS.md
```diff
@@ -32,6 +32,8 @@
   double (#577).
 - The conversion of R's `POSIXct` class to Polars datetime now works correctly with millisecond
   precision (#589).
+- `<LazyFrame>$filter()`, `<DataFrame>$filter()`, and `pl$when()` allow multiple conditions
+  input like `lf$filter(pl$col("foo") == 1, pl$col("bar") != 2)` (#598).
```
etiennebacher, line 36:
```suggestion
- `<LazyFrame>$filter()`, `<DataFrame>$filter()`, and `pl$when()` now allow multiple conditions
  to be separated by commas, like `lf$filter(pl$col("foo") == 1, pl$col("bar") != 2)` (#598).
```

etiennebacher (approved):

etiennebacher:
> The meaning of this error is quite difficult to understand

We could do a `tryCatch()` and throw our own error but that should be done in another PR anyway

eitsupi:
I created an issue about the issue of extra comma. #600

PR #602, opened 2023-12-17
Add extensions and remove examples from README

### LINKED ISSUE #537: Mention extensions in the README?
So far I think  `polarssql` and `tidypolars`  are the only extensions based on `polars`

etiennebacher:
Close #537

I removed easy examples and point to the "Get started" vignette instead. This is just a proposal, happy to modify if needed. We could also organize it like `dplyr` with:

1. Overview
2. Extensions
3. Installation
4. Short examples

@eitsupi btw I replaced "backend" by "methods"


eitsupi (changes requested):
I'm wondering if it would be enough to delete the Rmd and just have a simple md file.

> btw I replaced "backend" by "methods"

I feel it's not common to use the method in this context.

etiennebacher:
> I feel it's not common to use the method in this context.

I just think that "`polarssql` provides `DBI` backend and `dbplyr` backend." is not right, it should be "`polarssql` provides a `polars` backend for `DBI` and `dbplyr` interfaces".

> I'm wondering if it would be enough to delete the Rmd and just have a simple md file.

Depends if we want to remove the examples from the README

etiennebacher:
I moved the "get started" part before the installation part. `arrow` and `dplyr` also go in this order and I think it makes sense to list what a package can do before a long list of various installation processes

grantmcdermott:
> I removed easy examples and point to the "Get started" vignette instead.

My 2c: I'm a fan of providing README quickstart examples in general, since it gives users a concrete preview of what your package actually does. (Potentially of greater value in this particular context, given the number of competing data manipulation frameworks already available to R users.) The main polars repo also provides [quickstart examples](https://github.com/pola-rs/polars#python) on their README, so I think you're being consistent here.

etiennebacher:
@grantmcdermott I can add them back, where do you think they should be placed? Before or after installation instructions?

grantmcdermott:
After installation.

eitsupi:
I imagined that it would be a good idea to put the exact same thing in the README as the README of pola-rs/polars.

etiennebacher:
I addressed your comments, what do you think now @eitsupi @grantmcdermott ?

README.Rmd
```diff
@@ -1,7 +1,5 @@
 ---
-output:
-  github_document:
-    html_preview: false
+output: github_document
```
eitsupi, line 2: Why this change?

README.Rmd
eitsupi, line 1: Could you copy from the main branch instead of manually updating or git revert?

eitsupi (commented):
Sorry for the delay.

README.Rmd
etiennebacher: just because I like seeing the HTML version when I render it locally, it doesn't change anything for CI or else I think

README.Rmd
etiennebacher: I'll open a new PR instead, not sure what's the best way to clean it now

grantmcdermott:
> I addressed your comments, what do you think now @eitsupi @grantmcdermott ?

Sorry, I somehow missed this. Lmk if there's anything I need to look at in the new PR.

PR #607, opened 2023-12-18
improve unpack_list error

### LINKED ISSUE #600: Error messages from `with_columns()` is difficult to understand when extra comma
The meaning of this error is quite difficult to understand. The easiest way to fix this is to use `rlang::list2`, but we can't do that here, so this error should be allowed.

```r
> pl$DataFrame(mtcars)$with_columns(foo = TRUE,)
Error in unpack_list(...) : argument is missing, with no default
```

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/issues/598#issuecomment-1859177891_

sorhawell:
resolves #600 

this PR improves the bad error message for trailing commas for internal function `unpack_list()`.

Also the function now has `.context=` and `.call` args just as `unwrap()`.

I have provided context for any use of `unpack_list()` 

In future when/if extendr+polars supports/allows trailing commas. This extra error styling can be removed again.

examples 

```R

> unpack_list(1,2,)
0: During function call [unpack_list(1, 2, )]
1: trailing argument commas are not (yet) supported with polars
2: argument is missing, with no default
> pl$DataFrame()$select(1,2,)
0: In R: in $select()
0: During function call [pl$DataFrame()$select(1, 2, )]
1: trailing argument commas are not (yet) supported with polars
2: argument is missing, with no default

```


etiennebacher (commented):
LGTM, can you add some tests and bump news?

eitsupi (commented):
Thanks for working on this.

Could you also update `unpack_bool_expr`?

tests/testthat/test-dotdotdot.R
eitsupi, line 1: If we keep this in a separate file, wouldn't it be more consistent to also move the original functions to dotdotdot.R?

tests/testthat/test-dotdotdot.R
```diff
@@ -0,0 +1,56 @@
...
+  # unpacks a single list, return ok-result, get same polars expression
+  a = unpack_bool_expr_result(pl$col("a") == 1, pl$col("b") > 2)
+  b = unpack_bool_expr_result(list(pl$col("a") == 1, pl$col("b") > 2))
+  expect_true(a$ok$meta$eq(b$ok))
+
+  # ... not if more args
+  expect_failure(expect_equal(
+    unpack_list(list(1, 2, 3), "a"),
+    unpack_list(1, 2, 3, "a")
+  ))
```
eitsupi, line 48: Ditto.

tests/testthat/test-dotdotdot.R
```diff
@@ -0,0 +1,56 @@
...
+  expect_identical(
+    unpack_list(list(1, 2, 3)),
+    unpack_list(1, 2, 3)
+  )
+
+  # ... not if more args
+  expect_failure(expect_equal(
+    unpack_list(list(1, 2, 3), "a"),
+    unpack_list(1, 2, 3, "a")
+  ))
```
eitsupi, line 12: My understanding is that such usage is not common and should not be done as it is confusing with the number of tests displayed in the UI.

eitsupi (changes requested):

tests/testthat/test-dotdotdot.R
sorhawell: Normally I don't unittest internal functions too much. @etiennebacher  asked for it and since this one is very central I wrote the tests

I think this is in fact a very normal situation and a central  expected behavior 


```R
# inputs to DataFrame a list and a vector of same length
> pl$DataFrame(list(1,2,3),1:3)
shape: (3, 2)
┌────────────┬──────────────┐
│ new_column ┆ new_column_1 │
│ ---        ┆ ---          │
│ list[f64]  ┆ i32          │
╞════════════╪══════════════╡
│ [1.0]      ┆ 1            │
│ [2.0]      ┆ 2            │
│ [3.0]      ┆ 3            │
└────────────┴──────────────┘

# in the single case where input is a list and no other args, the list is unpacked
> pl$DataFrame(list(1,2,3))
shape: (1, 3)
┌────────────┬──────────────┬──────────────┐
│ new_column ┆ new_column_1 ┆ new_column_2 │
│ ---        ┆ ---          ┆ ---          │
│ f64        ┆ f64          ┆ f64          │
╞════════════╪══════════════╪══════════════╡
│ 1.0        ┆ 2.0          ┆ 3.0          │
└────────────┴──────────────┴──────────────┘

```

sorhawell: in py-polars dicts are "unpacked" to columns. We allow both `data.frame` like interface where each column is an argument but also the py-polars way where all columns can be passed as a single named list.

```py
In [9]: pl.DataFrame({"a": [1,2,3], "b": [1,2,3]})
Out[9]: 
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 1   │
│ 2   ┆ 2   │
│ 3   ┆ 3   │
└─────┴─────┘
```

sorhawell:
```R
> pl$DataFrame(list(list(1,2,3)))
shape: (3, 1)
┌────────────┐
│ new_column │
│ ---        │
│ list[f64]  │
╞════════════╡
│ [1.0]      │
│ [2.0]      │
│ [3.0]      │
└────────────┘
> pl$DataFrame(list(list(1,2,3)),"a")
shape: (1, 2)
┌───────────────────────┬──────────────┐
│ new_column            ┆ new_column_1 │
│ ---                   ┆ ---          │
│ list[list[f64]]       ┆ str          │
╞═══════════════════════╪══════════════╡
│ [[1.0], [2.0], [3.0]] ┆ a            │
└───────────────────────┴──────────────┘
```

tests/testthat/test-dotdotdot.R
eitsupi: What I would like to point out is that `expect_*` functions are nested

sorhawell: ohh misunderstood what you meant, what you prefer instead?

tried this 
image

it seem the counting of number of tests is correct
image


I guess expect_success will catch the fail_test exception, such that testthat only sees the outer exception, if any.

eitsupi: Thanks for the clarification. There doesn't seem to be a problem with this usage

etiennebacher: > Normally I don't unittest internal functions. @etiennebacher asked for it 

Just to clarify I wanted to test the new error messages from `$select()` and others, not necessarily test the internal functions directly. I'm fine with this.

etiennebacher:
Apart from moving the functions dealing with `...` in a separate `dotdotdot.R`, is there anything else to do here?

eitsupi (approved):

Issue #613, opened 2023-12-20
`make docs` compiles the Rust crates twice


etiennebacher:
When I run `make docs`, the makefile calls `build` and then `install`:

https://github.com/pola-rs/r-polars/blob/d6cc4dfe7e1045d172ae5aee4d11ba8e8f72a8a5/Makefile#L68

This compiles the Rust deps twice, which is a waste of time. Am I the only one with this? Is it possible to fix this?

sorhawell:
It may be due to a change of working directory as with `R CMD check`. I can try to see if symlinking will fix it.

sorhawell:
I think there is a bad habbit in many R build tools like extendr, roxygen2 etc. to automatically and with no opt-out to call `devtools::load_all()` to ensure package is loaded. This is ok for small packages, but for a mammuth like polars, even the R part slows down the process significantly. 


for `make docs`  I count `loading polars`  4 times in the log.  I saw one build of rust code before an error with forkcluster with the `future package`

```sh
(.venv) Srens-MacBook-Pro:r-polars sorenwelling$ make docs
export NOT_CRAN=true \
        && export LIBR_POLARS_BUILD=true \
        && export RPOLARS_FULL_FEATURES=true \
        && Rscript -e 'if (!(require(arrow) && require(nanoarrow) && require(knitr))) warning("could not load arrow/nanoarrow/knitr, ignore changes to nanoarrow.Rd or knit_print.Rd"); rextendr::document()'
Loading required package: arrow

Attaching package: ‘arrow’

The following object is masked from ‘package:utils’:

    timestamp

Loading required package: nanoarrow
Loading required package: knitr
ℹ Generating extendr wrapper functions for package: polars.
✔ Writing 'R/extendr-wrappers.R'
ℹ Updating polars documentation
ℹ Loading polars
export NOT_CRAN=true \
        && export LIBR_POLARS_BUILD=true \
        && export RPOLARS_FULL_FEATURES=true \
        && R CMD INSTALL --no-multiarch --with-keep.source .
* installing to library ‘/Library/Frameworks/R.framework/Versions/4.3-x86_64/Resources/library’
* installing *source* package ‘polars’ ...
** using staged installation

--------------------------- [RUST FOUND] ---------------------------
cargo 1.73.0-nightly (7ac9416d8 2023-07-24)

rustc 1.73.0-nightly (0d95f9132 2023-07-26)
binary: rustc
[...]

Issue #556, opened 2023-12-01
[Develop] The Makefile is used?


eitsupi:
Previous discussion: #173

Currently, this repository allows users to use Makefile to auto-format and run tests with a single command, but I suspect that the current state of the art is not very useful.
Since the Makefile does not have the ability to record the previous execution, each time the command is run, even if the source code has not changed, the installation of the polars package and execution of all tests will take quite some time. Developers who do not like the time it takes may stop running tests.

@sorhawell @etiennebacher What do you think about this?

I was surprised to see that @etiennebacher manually edited the `extendr-wrappers.R` file in #554.
Essentially this file should be automatically updated in the developer workflow, so I think it shows that the current workflow is not good.

As I noted in an earlier discussion, I believe [Task](https://taskfile.dev/) is more appropriate than Make for this type of use, even if additional installation is required.

eitsupi:
> Does that still require clang/gcc and Make?

These are always necessary when installing any R source package, not just this package.
In other words, there is little chance that a Linux R user will not have them.

> Maybe we could have some script to bootstrap a developer stack.

In my opinion, it is not a good idea to have a script install such a tool because there are multiple ways to install it. Simply prompt them to install with documentation. (Same as Rust, Python, or R itself).

PR #617, opened 2023-12-23
build: add Taskfile.yml to use task as a build tool

eitsupi:
A part of #556

Allows use of `task` instead of `make`.
Most of the content was copied from the prqlr repository.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -176,6 +176,7 @@ tasks:
       - Rscript -e 'if (desc::desc_get("RoxygenNote") < packageVersion("roxygen2")) quit(status = 1)'
     deps:
       - build-extendr-wrappers
+      - format-r
     cmds:
       - Rscript -e
         '{{.R_FUNC_IMPORT_REQUIRED_PKGS}};
@@ -190,6 +191,8 @@ tasks:
       - "{{.MANIFEST}}"
       - "{{.CARGO_LOCK}}"
       - "{{.RUST_SOURCE}}"
+    deps:
+      - format-rs
     generates:
       - R/extendr-wrappers.R
     status:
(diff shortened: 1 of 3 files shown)


eitsupi:
At the moment, only the main targets have been ported from the Makefile, and the minor targets have not yet been ported.
But building and testing the entire package should work. Like `task build-all`, `task build-website`, etc.

@etiennebacher Could you try this?
Task should work on Windows since it doesn't depend on Unix shell, but I haven't tested it here.

etiennebacher:
Thanks @eitsupi, I confirm it works fine on Windows once I downloaded `Task` (I just tried `build-all` and `setup-dev`). I'll need more time to see if it fixes #613 but at least the basic usage works

Taskfile.yml
```diff
@@ -0,0 +1,239 @@
...
+    cmds:
+      - R CMD INSTALL --no-multiarch --with-keep.source .
+
+  build-website:
+    desc: Build the website.
+    sources:
+      - man/*.Rd
+      - altdoc/altdoc_preprocessing.R
+      - altdoc/*.yml
+      - vignettes/*.Rmd
```
eitsupi, line 229: We can use the `{{.VIGNETTES}}` variable instead of this.

eitsupi:
I believe I have ported the full functionality of the Makefile.

```sh
task setup-dev
task build-all
task build-website
```

etiennebacher (changes requested):
I think the documentation workflow needs to be updated as well to use task instead

eitsupi:
> I think the documentation workflow needs to be updated as well to use task instead

I'd like to do that in a follow-up PR. This PR includes an update to Rbuildignore, so it triggers the R workflow needlessly.

etiennebacher (approved):
Thanks @eitsupi

eitsupi:
Thanks for taking a look at this.

etiennebacher:
@eitsupi I use the taskfile more regularly now and the Rust crates are only compiled once with `task build-all`, thanks!

Issue #616, opened 2023-12-23
0.12.0 release


eitsupi:
If there is nothing special, we can create a release right away.

etiennebacher:
If #597 is close to be merged, it would be nice to include it. Apart from that, I'm fine with releasing 0.12.0

eitsupi:
We may want to include pola-rs/polars#13239

etiennebacher:
Yes I think we should

sorhawell:
> If #597 is close to be merged, it would be nice to include it. Apart from that, I'm fine with releasing 0.12.0


I will try to wrap up the PR before new year. It should only be branded as a experimental feature so far. There are for sure bugs.


status quo on a 5 star rating scale:

Rstudio **** (method + arg completion + polars error explanation)
Radian *** (method completion + polars error explanation)
Native R terminal * (method completion, no drop down, sometime completion disappear)
vscode-R ** (method completion + shows method formals, [NEEDS this PR  merged](https://github.com/sorhawell/vscode-R/pull/1), can only complete the last expression of code file :/ )

eitsupi:
It seems like it will take some time, so I think we can release 0.12.0 first and include it in 0.12.1.

etiennebacher:
Yes I agree

eitsupi:
Done in #631

eitsupi:
> I will try to wrap up the PR before new year.

Any update?

PR #622, opened 2023-12-25
Bump rust-polars to unreleased 2023-12-25

etiennebacher:
Related to #616


etiennebacher:
@eitsupi I'm stuck on the failures of release-lib workflows. I tried bumping the version of the r-polars crate but no success. Maybe the only thing to do is to bump the rust version in cargo.toml?

eitsupi:
> I'm stuck on the failures of release-lib workflows. I tried bumping the version of the r-polars crate but no success. Maybe the only thing to do is to bump the rust version in cargo.toml?

lib 0.35.1 has not been released, so no version change is required.
I think the build failure is not a problem with our Rust source code, but with the toolchain, since it is only happening at nightly and is failing to build polars-arrow in the dependent crate, not r-polars.

etiennebacher:
Thanks for the fix @eitsupi

sorhawell (approved):

eitsupi (approved):

PR #631, opened 2023-12-28
release: 0.12.0

eitsupi:


NEWS.md
```diff
@@ -1,6 +1,6 @@
-## polars (development version)
+# polars 0.12.0
```
eitsupi, line 1: @etiennebacher It seems that the headers of each version of NEWS are not recognized by usethis unless they are set to L1.

etiennebacher: It should work fine with altdoc but I'll double check today

eitsupi: I built the website and realized that toc is not generated as expected.
Maybe altdoc needs to change the header level?
image

eitsupi: I created an issue for this #632

etiennebacher: I think it would actually be better to have only one h1 tag named "NEWS" or "Changelog", so that  the TOC doesn't change on the website. Mkdocs requires only one h1 tag, so we would need to convert all h1 to h2 in `altdoc`. That could be done eventually, but for now I think it's better to keep the NEWS as-is. Can you revert the changes on NEWS.md?

eitsupi: In general Markdown notation, yes, but in R's NEWS.md, this is the common use of headers, and I think it creates friction to deviate from it.

eitsupi: Anyway, now we need to merge this and make follow-up corrections.

etiennebacher: NEWS files can have several formats (even if there is one more used than the others) so we decided to not modify them ourselves in `altdoc`. It's not hard for the users to bump the version by hand when necessary. (just for info I don't want to spend much more time on `altdoc`. It was already more than planned so if we have something that works for docs I'd rather keep it as is)

etiennebacher (approved):

NEWS.md
etiennebacher: > Anyway, now we need to merge this and make follow-up corrections.

I'll make a PR to revert this

eitsupi: Unless we automate that process, we will have to manually fix it with every release.......

PR #633, opened 2023-12-28
build: add devcontainer.json for to support Dev Containers

eitsupi:
Replace #312.

Add a container definition to be able to work with GitHub Codespaces etc.
(I'm using this locally and it seems to be working fine.)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,26 @@
+{
+	"image": "ghcr.io/rocker-org/devcontainer/r-ver:4",
+	"features": {
+		"ghcr.io/devcontainers/features/rust:1": {},
+		"ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
+			"packages": "cmake,python3-venv"
+		},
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+			"type": "volume"
+		},
+		{
+			"source": "devcontainer-cargo-cache-${devcontainerId}",
+			"target": "/usr/local/cargo",
+			"type": "volume"
+		}
+	],
+	"updateContentCommand": {
+		"prqlr-build-deps": "task setup-dev"
```
etiennebacher, line 24: Is it a mistake to have prqlr here?

eitsupi: Yeah, it looks like this was copied from prqlr so I forgot to change it.
Thanks for pointing this out.

eitsupi, line 24:
```suggestion
		"setup-r-polars-requirements": "task setup-dev"
```

etiennebacher (approved):

PR #636, opened 2023-12-28
docs(website): fix indentation in changelog

etiennebacher:
Currently nested lists are flattened, this fixes it


eitsupi:
Does this mean altdoc or something is expecting an indent size of 4?
In that case, I think we need to force the indent size using something like Editorconfig.

etiennebacher:
I think this is just required by `mkdocs`. The NEWS file in `altdoc` has nested lists with only 2 spaces and it works fine with the `docute` generator

>  In that case, I think we need to force the indent size using something like Editorconfig.

I'm not familiar with this, does it work with any IDE?

eitsupi:
> I'm not familiar with this, does it work with any IDE?

Yes, except for RStudio IDE. (rstudio/rstudio#1711)

etiennebacher:
@eitsupi I don't know how to enable Editorconfig, so feel free to merge this when ready

eitsupi (approved):
Looks fine, thanks!

eitsupi:
This is the upstream issue https://github.com/mkdocs/mkdocs/issues/545#issuecomment-522196661

PR #641, opened 2024-01-01
Implement string methods `reverse()`,  `contains_any()`, `replace_many()`

etiennebacher:


R/expr__string.R
```diff
@@ -822,3 +822,83 @@ ExprStr_explode = function() {
 ExprStr_parse_int = function(radix = 2, strict = TRUE) {
   .pr$Expr$str_parse_int(self, radix, strict) |> unwrap("in str$parse_int():")
 }
+
+#' Returns string values in reversed order
+#'
+#' @name ExprStr_reverse
```
eitsupi, line 828: I think these `@name` tags can be removed.

eitsupi (approved):
Thanks! A minor comment.

R/expr__string.R
etiennebacher: Yes this won't matter anymore once #642 is merged. I can already remove them but don't pay attention to R CMD check failures due to this.

PR #626, opened 2023-12-26
docs(website): better display of "Usage" section

etiennebacher:
This only replaces e.g `DataFrame_drop()` by `<DataFrame>$drop()` in the "Usage" section. It doesn't address the issue of missing "Usage" section for expressions in a subnamespace.

PR #642, opened 2024-01-01
Fix absence of "Usage" section for Expr subnamespaces

etiennebacher:
None of the Expr subnamespaces (DT, list, etc.) have a "Usage" section in the docs. This is because the internal function `macro_new_subnamespace()` is run before the Rd files are generated by `roxygen`.  The problem is that `macro_new_subnamespace()` removes the functions that are defined (e.g `ExprDT_days()`) and renames them, so when `roxygen` tries to detect how to use `ExprDT_days()`, it doesn't exist anymore and therefore the usage section is empty. This is also why we need to manually add the `@name` tag.

This PR fixes this by removing the argument `remove_f` in `macro_new_subnamespace()`, which means that the original functions are no longer deleted. I don't think this has an effect on the users: the snapshots that list all methods are not modified and all the tests pass. I also removed the `@name` tags, since they're no longer necessary.

In a follow-up PR, I'll deal with renaming the `ExprDT_` into something cleaner on the website (something similar to #626).

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -458,7 +458,7 @@ restruct_list = function(l) {
 #' # e$my_sub_ns$add2() #use the sub namespace
 #' # e$my_sub_ns$mul2()
 #'
-macro_new_subnamespace = function(class_pattern, subclass_env = NULL, remove_f = TRUE) {
+macro_new_subnamespace = function(class_pattern, subclass_env = NULL) {
   # get all methods within class
   class_methods = ls(parent.frame(), pattern = class_pattern)
   names(class_methods) = sub(class_pattern, "", class_methods)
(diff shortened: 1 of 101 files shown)


R/utils.R
etiennebacher, line 1: This is the main change

R/utils.R
```diff
@@ -458,7 +458,7 @@ restruct_list = function(l) {
 #' # e$my_sub_ns$add2() #use the sub namespace
 #' # e$my_sub_ns$mul2()
 #'
-macro_new_subnamespace = function(class_pattern, subclass_env = NULL, remove_f = TRUE) {
+macro_new_subnamespace = function(class_pattern, subclass_env = NULL) {
```
etiennebacher, line 461: This is the main change

man/S3_knit_print.Rd
eitsupi, line 1: Unnecessary change.

man/nanoarrow.Rd
eitsupi, line 1: Ditto.

eitsupi (changes requested):
Thanks for working on this!

eitsupi (approved):

PR #654, opened 2024-01-03
build: remove the Makefile in favor of the Taskfile.yml

### LINKED ISSUE #556: [Develop] The Makefile is used?

eitsupi:
Close #556


.github/workflows/docs.yaml
```diff
@@ -86,12 +87,18 @@ jobs:
...
-        run: make docs
-        shell: bash
+        run: task build-website
+
+      - name: upload docs
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: docs
+          path: docs
```
etiennebacher, line 101: what does this do?

eitsupi: Upload the website generated in PR as an artifact.
Download the artifact to check if the website is rendering correctly.
Without this, it is difficult to check whether your website was built correctly.

etiennebacher (approved):
Thanks for all the work on the tooling/workflows

PR #656, opened 2024-01-03
docs: Minor corrections to documents

eitsupi:
- `print` methods should not have Rd files.
- Rename the `<DataFrame>$to_list()` function's Rd file.


etiennebacher (approved):
Thanks

PR #657, opened 2024-01-03
docs: remove `@usage NULL` for datetime methods

etiennebacher:
Something forgotten in previous PRs. These were the last `@usage NULL` to remove


eitsupi (approved):
Thanks!

PR #658, opened 2024-01-03
feat: add `is_polars_*` functions

eitsupi:
In the R world, the `is*` functions that pairs with the `as*` functions are common, and for example, tibble and data.table also have such a function, so I think polars should also have these.


etiennebacher (approved):
Thanks

PR #659, opened 2024-01-03
feat: Bump rust-polars to 0.36.2

etiennebacher:
https://github.com/pola-rs/polars/releases/tag/rs-0.36.2

We updated most things in the previous bump to unreleased 2023-12-25. The main change here is the renaming of `Utf8` to `String`


src/rust/src/rdatatype.rs
```diff
@@ -73,7 +73,7 @@ impl RPolarsDataType {
             "Float32" | "float32" | "double" => pl::DataType::Float32,
             "Float64" | "float64" => pl::DataType::Float64,
 
-            "String" | "character" => pl::DataType::String,
+            "Utf8" | "String" | "character" => pl::DataType::String,
```
eitsupi, line 76: I was just getting started on this too. I think we can leave Utf8 as an alias, as py-polars does the same thing.

etiennebacher: Yes, we should keep an alias but I guess docs should be updated to prefer "String" instead

etiennebacher: Go ahead if you want, let me know when I can jump in ;)

eitsupi:
Breaking changes affect R side:

- pola-rs/polars#13224
- pola-rs/polars#13307

eitsupi:
I think I passed all the tests.

I'm going to bed now, so I won't update any more. It would be helpful if you could update the documentation etc.

etiennebacher:
@eitsupi I think it's good to go. There are no breaking changes here, so I think we could release 0.12.1 so that we don't get two rust-polars updates in one release

PR #669, opened 2024-01-05
docs: minor updates for the install vignette

eitsupi:
- pl$polars_info -> polars_info
- Add note about `MY_UNIVERSE`
- Other minor update


etiennebacher (approved):
Thanks, can you also replace the title by something like "Installation details"?

PR #671, opened 2024-01-05
ci: fix typo of the path in the ci trigger

eitsupi:
I noticed that #669 was not triggering the docs workflow due to a typo in this file.


etiennebacher (approved):
Good catch

PR #682, opened 2024-01-08
Implement `$rolling()` for `DataFrame` and `LazyFrame`

etiennebacher:


eitsupi (changes requested):
Thanks for working on this.
I don't think this currently supports `$ungroup()`. Need to add tests.

Also, rather than making the implementation of RPolarsGroupBy complicated like this, I thought it would be simpler and better to delete `by_agg` on the Rust side and rewrite it to always call `$lazy()` and `$collect()` on the R side.
https://github.com/pola-rs/r-polars/blob/ffc9a5c63f145793b6be8c5b64bde9e12d2b7dc8/src/rust/src/rdataframe/mod.rs#L310-L327

etiennebacher:
> I thought it would be simpler and better to delete by_agg on the Rust side and rewrite it to always call $lazy() and $collect() on the R side.

I'll open a new PR for that as it touches the `$agg()` behavior overall, not just for rolling computations. Would it be ok for you to have a "dirty" implementation of `$ungroup()` similar to the change in `$agg()` here in the meantime?

eitsupi:
> Would it be ok for you to have a "dirty" implementation of `$ungroup()` similar to the change in `$agg()` here in the meantime?

Of course it's okay. Thanks!

R/dataframe__frame.R
```diff
@@ -1807,3 +1807,27 @@ DataFrame_write_ndjson = function(file) {
     unwrap("in $write_ndjson():") |>
     invisible()
 }
+
+#' @inherit LazyFrame_rolling title description params details
+#' @return A GroupBy object
```
eitsupi, line 1812:
```suggestion
#' @return A [GroupBy][GroupBy_class] object
```

R/lazyframe__lazy.R
```diff
@@ -1613,3 +1613,44 @@ LazyFrame_with_context = function(other) {
...
+#' @inherit Expr_rolling description details params
+#' @param index_column Column used to group based on the time window. Often of
+#' type Date/Datetime. This column must be sorted in ascending order (or, if `by`
+#' is specified, then it must be sorted in ascending order within each group). In
+#' case of a rolling group by on indices, dtype needs to be either Int32 or Int64.
+#' Note that Int32 gets temporarily cast to Int64, so if performance matters use
+#' an Int64 column.
+#' @param by Also group by this column/these columns.
+#'
+#' @return A LazyGroupBy object
```
eitsupi, line 1629:
```suggestion
#' @return A [LazyGroupBy][LazyGroupBy_class] object
```

eitsupi (approved):

R/dataframe__frame.R
etiennebacher: I always forget this, thanks

PR #684, opened 2024-01-09
feat(installation): support cross-compiling the R package via the `--host` option of the configure script

eitsupi:
Same as eitsupi/prqlr#247

R-universe started distributing R binary packages for arm64 macOS yesterday (r-universe-org/help#340).

image

This PR should enable both of the following two R package cross-compilation methods to work.

1. Compile from Rust source (confirmed with eitsupi/prqlr#247)
2. Download and use Rust binary libraries from the internet (Checked on my R-universe https://github.com/eitsupi/eitsupi.r-universe.dev/commit/2b5163fe86eac278eee3a1fe7a343b7c9bd37518)

### DIFF
--- a/configure
+++ b/configure
@@ -41,12 +41,12 @@ check_cargo() {
 }
 
 check_bin_lib() {
-  if [ "${NOT_CRAN}" = "true" ] && [ -z "${LIBR_POLARS_BUILD}"  ]; then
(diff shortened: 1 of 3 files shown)


etiennebacher (approved):
No clue about this, so just go ahead ;)

etiennebacher:
Ok for the release (can you bump news?)

Actions failing due to github: githubstatus.com

PR #689, opened 2024-01-10
Clean the `agg()` implementation for `DataFrame`

### LINKED ISSUE #552: Rewrite `ungroup` for LazyGroupBy with `From`
See pola-rs/polars#12562

etiennebacher:
Followup of #682, 

* I feel like the way `RollingGroupBy` object was handled in #682 is actually not that bad (but I'm open to ideas on how to improve more)
* added the Rd file for `LazyGroupBy_class` which was missing and caused R CMD check failure in #682

We could also fix #552 in this PR, but I don't know how to do that.

@eitsupi I have a hard time updating the snapshots in parametric tests, they always fail with this kind of error:
```
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error (test-after-wrappers.R:30:7): public and private methods of each class DataFrame
Error in `trace_back(top = getOption("testthat_topenv"), bottom = trace_env)`: Can't find `bottom` on the call tree.
Backtrace:
    ▆
 1. ├─rlang::eval_tidy(code, args)
 2. └─testthat::expect_snapshot(ls(.pr[[private_key]])) at test-after-wrappers.R:30:7
 3.   └─testthat:::expect_snapshot_helper(...)
 4.     └─testthat::expect(...)
 5.       └─rlang::trace_back(top = getOption("testthat_topenv"), bottom = trace_env)
 6.         └─rlang:::trace_find_bottom(bottom, frames)
 7.           └─rlang::abort(msg, call = call)
```
I have to comment the parametric test, run the tests (to remove the snapshots) and then run them again to regenerate them. Do you have the same issue?


eitsupi:
> I have to comment the parametric test, run the tests (to remove the snapshots) and then run them again to regenerate them. Do you have the same issue?

No such work is necessary. Run `testthat::snapshot_accept()` to update the snapshot after the test fails. (It would be nice if there was an option to force `devtools::test()` to update the snapshot, but there doesn't seem to be one at the moment.)
https://github.com/pola-rs/r-polars/blob/e457053db4c4e4547975bca326e2a1271e193257/Taskfile.yml#L294-L297

tests/testthat/_snaps/groupby.md
```diff
@@ -15,9 +15,7 @@
       │ one ┆ 4.0 │
       │ two ┆ 1.0 │
       └─────┴─────┘
-      groups: [[1]]
-      [1] "foo"
-      
+      groups: foo
```
eitsupi, line 18: Could you add tests for when there are multiple groups?

eitsupi (changes requested):
Thanks, please update GroupBy_ungroup too.

tests/testthat/_snaps/groupby.md
etiennebacher: done

etiennebacher:
> Thanks, please update GroupBy_ungroup too.

IINM, if we want to handle rolling groupby separately, we need to make a `RollingGroupBy` object and the equivalent Struct in Rust. I'm not necessarily against that but I'd rather let that for another PR

eitsupi (approved):
Ok, thanks. Let's merge this now.

PR #691, opened 2024-01-11
Implement `group_by_dynamic()` for `DataFrame` and `LazyFrame`

etiennebacher:


tests/testthat/test-groupby.R
```diff
@@ -7,6 +9,19 @@ df = pl$DataFrame(
...
+    pl$col("bar")$mean()$alias("bar_tail_sum")
+  )$to_data_frame()
+
+  expect_equal(
+    df2,
+    data.frame(foo = c("one", "two"), bar_sum = c(9, 6), bar_tail_sum = c(4.5, 2))
+  )
+})
+
+
```
etiennebacher, line 24: I only moved this above because otherwise RStudio doesn't provide the button "Test"

R/group_by.R
eitsupi, line 1: Why not completely rewrite GroupBy first? https://github.com/pola-rs/r-polars/pull/682#pullrequestreview-1812880011

etiennebacher:
@eitsupi I ended up removing the attributes when we print the `DynamicGroupBy` and `RollingGroupBy` objects because if we print this class of objects, it means they are last in the pipeline, so we already have all the info about groups and other args. Maybe that's what you said in #694, I'm not sure.

R/lazyframe__lazy.R
```diff
@@ -1720,3 +1720,114 @@ LazyFrame_rolling = function(index_column, period, offset = NULL, closed = "righ
...
+    index_column,
+    every,
+    period = NULL,
+    offset = NULL,
+    include_boundaries = FALSE,
+    closed = "left",
+    label = "left",
+    by = NULL,
+    start_by = "window",
+    check_sorted = TRUE) {
```
eitsupi, line 1821: (as well as other functions) Could you please enforce named arguments using `...`?
https://github.com/pola-rs/polars/blob/cdef09b98f5164271ac1a489b395accf4a363549/py-polars/polars/lazyframe/frame.py#L3113-L3127

eitsupi (changes requested):

eitsupi (approved):
Thanks!

PR #692, opened 2024-01-11
Implement `pl$duration()`

### LINKED ISSUE #661: Implement `pl$duration()`
https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.duration.html#polars.duration

etiennebacher:
Close #661


R/functions__lazy.R
```diff
@@ -1007,3 +1007,70 @@ pl_sum_horizontal = function(...) {
...
+pl_duration = function(
+    weeks = NULL,
+    days = NULL,
+    hours = NULL,
+    minutes = NULL,
+    seconds = NULL,
+    milliseconds = NULL,
+    microseconds = NULL,
+    nanoseconds = NULL,
+    time_unit = "us") {
```
eitsupi, line 1073: I think it very difficult to understand that these arguments can be specified by position.
Please put `...` so that we can force the argument name.

In fact, this seems to be the case with Python as well.
https://github.com/pola-rs/polars/blob/58b0da8bc31d71d21be8782c5013747516652f81/py-polars/polars/functions/as_datatype.py#L178-L189

eitsupi (changes requested):

R/functions__lazy.R
etiennebacher: Yes, good catch

eitsupi (approved):
Thanks!

Issue #445, opened 2023-10-28
release-optimized profile?


eitsupi:
Is there any point in continuing to have the `release-optimized` profile exist? If it is not needed, I would like to delete it.

ritchie46:
fat linking with a single thread leads to better optimized binaries as the compiler has all the information.

sorhawell:
> fat linking with a single thread leads to better optimized binaries as the compiler has all the information.

yeah and 40% slimmer in our case :)

sorhawell:
> Thank you for your reply. My question was, could I integrate them into the release profile?

Currently we use the "release" profile for development on local machines and "release.optimized" as the preferred for binary release. On my slow machine it takes about 20min compile r-polars + dependencies with "release", and 30 sec when dependencies are already compiled. 

When running "release.optimized" with lto="fat" it takes about 40 min to compile r-polars + dependencies and I think about 2-4 min for r-polars + linking.

Maybe we should dedicated "dev" profile instead.  Potentially with even lower optimizations.

Issue #300, opened 2023-07-04
Revise the contents of Cargo.toml


eitsupi:
Looking at rust-polars and python-polars, the features that require the nightly toolchain are named `nightly`.

https://github.com/pola-rs/polars/blob/126aef55c4909021176aad2cbb190f5a8daf731c/py-polars/Cargo.toml#L48
https://github.com/pola-rs/polars/blob/126aef55c4909021176aad2cbb190f5a8daf731c/polars/Cargo.toml#L17

Since we are now directly controlling the enabling and disabling of the `simd` feature here, it would be better to rewrite it to use `nightly` feature for consistency.

etiennebacher:
@eitsupi is this still an issue?

eitsupi:
Yes, I still want to work on this. I mean, we currently use a feature named `full_features`, which is realistically "nightly".
https://github.com/pola-rs/r-polars/blob/2472c24d22b8a4e8d77b6351c78e60410287b193/src/rust/Cargo.toml#L11-L16

I would like to keep `full_features` as a means to adjust MSRV for CRAN submissions and create `nightly` under it.
(For example, as of 0.10.0, it seems that it is `polars/lazy` that raises the MSRV from 1.71 to 1.73, so a possible response would be to add `polars/lazy` to `full_features`)

Issue #241, opened 2023-06-12
how to config max number of threads?


tdhock:
hi @sorhawell is there a function that I can call to tell polars the max number of threads I want it to use?
something like data.table::setDTthreads(4) if I only want to use 4 threads max.
Thanks!
I searched the documentation for "thread" and got a few hits but nothing that helped.

sorhawell:
The default is likely the double of number of cores. I did not know it was possible to modify, but Ifound this [function in py-polars docs](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.threadpool_size.html#polars.threadpool_size) and learned you can use `Sys.setenv("POLARS_MAX_THREADS"=3)` before loading library polars


`polars:::test_get_pool_size()` is not published yet but it shows that envvar option seems to work

```R
> polars:::test_get_pool_size()
[1] 8

Restarting R session...

> library(polars)

Restarting R session...

> Sys.setenv("POLARS_MAX_THREADS"=3)
> library(polars)
> polars:::test_get_pool_size()
```

on top of this r-polars reserves one extra thread for running the R interpreter. If not injecting R code into polars with $map() or $apply(), that thread will mostly be sleeping.

tdhock:
Hi again, thanks for the quick response. That is helpful to know that we can set an environment variable to set the max number of threads before polars is loaded, but is there a way to change it after it has already been loaded/attached via library(polars)? (similar to data.table::setDTthreads)
Also that is beneficial to know that a thread is reserved for running the R interpreter. So does that mean the minimal number for POLARS_MAX_THREADS is 0? or is it 1?
Would be nice to see some documentation about this topic at some point.
For context I am trying to do some benchmarking of single threaded vs multi-threaded csv reading code. 
Also, is there a polars function for csv writing? I did not see docs about that.

sorhawell:
> So does that mean the minimal number for POLARS_MAX_THREADS is 0? or is it 1?

POLARS_MAX_THREADS could have been named POLARS_RAYON_POOL_SIZE. I don't know what would happen on 0. Either 0 is not allowed or rayon hang I guess. Rayon is a rust crate which implements a thread pool, which polars is build upon. 1 would be lowest meaningful value I guess.


in the rust-polars for the csv reader and other readers it is possible to specifically decide on threads. These may not be bounded by POLARS_MAX_THREADS.  I will look into if r-polars could easily expose these settings.


Let's also keep this issue open until these thread findings have been added to docs.

[...]

PR #693, opened 2024-01-13
refactor!: simplify Makevars files and rename env vars for building the library

eitsupi:
Related to #80, #241, #300, and #445

- Allows to select any feature via `LIBR_POLARS_FEATURES` instead of `RPOLARS_FULL_FEATURES="true"`. This allows users to freely select features.
- Unify environment variable names from `RPOLARS*` to `LIBR_POLARS*` for consistency.
- Delete parts that are no longer needed.

These changes are brought over from [prqlr](https://cran.r-project.org/package=prqlr).


etiennebacher (commented):
Small nitpick but shouldn't these be named `LIB_R_POLARS_*` instead?

eitsupi:
> shouldn't these be named `LIB_R_POLARS_*` instead?

Good point.
The names are a very difficult issue, but I believe that `libr_polars` is the name of the actual binary library that will be generated, and adding an underscore like `LIB_R_POLARS` would make it confusing.

https://github.com/pola-rs/r-polars/blob/134ed5db04ae722a13203f0c4d8da5db6b5aeebc/.github/workflows/release-lib.yaml#L26

etiennebacher:
> I believe that `libr_polars` is the name of the actual binary library that will be generated

Would it be a breaking change to rename this binary `lib_r_polars`? It's just weird to have `libr`/`LIBR` as prefix. Since renaming envvars is a breaking change, we might as well rename the binary at the same time.

eitsupi:
> Would it be a breaking change to rename this binary `lib_r_polars`? It's just weird to have `libr`/`LIBR` as prefix.

This name is automatically generated from the crate name "r-polars". i.e. "lib" + "r-polars" -> "libr-polars" -> "libr_polars"
This is the cargo's default behavior, so I don't think there's anything strange about it.

etiennebacher:
Sorry for being nitpicky on this, but I'd prefer envvars to be named `R_POLARS_*`, e.g `R_POLARS_FEATURES`. This doesn't match the binary name but from the user perspective it makes more sense IMO. The users never interact directly with the binary generated by cargo, so for them there's no consistency issue in having the binary named `libr_polars` and the envvars named `R_POLARS_*`.

eitsupi:
> Sorry for being nitpicky on this, but I'd prefer envvars to be named `R_POLARS_*`, e.g `R_POLARS_FEATURES`. This doesn't match the binary name but from the user perspective it makes more sense IMO. The users never interact directly with the binary generated by cargo, so for them there's no consistency issue in having the binary named `libr_polars` and the envvars named `R_POLARS_*`.

That's a good point to consider, and I was thinking the same way when I opened this PR.

However, we can separate environment variables related to Rust build from other environment variables (for example, environment variables that affect the behavior of r-polars may be added) by using "LIB" as a prefix.
I thought maybe it wasn't a bad idea.

etiennebacher:
> However, we can separate environment variables related to Rust build from other environment variables (for example, environment variables that affect the behavior of r-polars may be added) by using "LIB" as a prefix.

Right, that's a good distinction to make, thanks for explaining.

At some point, we should add a vignette listing all possible envvars, both for the Rust build and for the use in R (in addition to what you put in the "installation" vignette).

etiennebacher (approved):

eitsupi:
Thanks for your review!
I would like @sorhawell to confirm this as well, so I will merge it afterwards.

sorhawell:
> Thanks for your review! I would like @sorhawell to confirm this as well, so I will merge it afterwards.

Hi @eitsupi  and @etiennebacher .

You probably noticed the frequency and quality of PRs and reviews have gone down lately. I think so. I will take a pause from open source for half year or more I think. I want to try out some new directions in a life. It has been a great honor to work with you. You can replace me as maintainer and/or "first author" whenever it makes sense.

I will try to commit the Rstudio completion PR and give up on the vscode completion PR.

eitsupi:
@sorhawell Thanks for the reply. Glad to hear you are doing well.

I will merge this and move on.

etiennebacher:
@sorhawell I hope this break will be good for you and you'll find things you enjoy doing. I'm always impressed by the amount of work you put to make this package working, thanks so much

grantmcdermott:
+1 OSS work is very rewarding... but it can also be incredibly draining. @sorhawell I'm among the many grateful people for all the work that you've put into `r-polars`. (Equally grateful for @eitsupi and @etiennebacher picking up the maintenance reins.) Enjoy the break and I hope that it brings you a sense of rest and rejuvenation.

PR #695, opened 2024-01-13
Some improvements to the "Reference" homepage

etiennebacher:


eitsupi (approved):

altdoc/reference_home.Rmd
```diff
@@ -26,10 +26,13 @@ to choose between eager and lazy evaluation, that require respectively a
...
-can also be used on `LazyFrame`s. Calling `<DataFrame>$lazy()` yields
-a `LazyFrame`. While calling `<LazyFrame>$collect()` starts a computation and
-yields a `DataFrame` as result.
+or `drop()`. Most functions that can be applied to `DataFrame`s can also be used 
+on `LazyFrame`s, but some are specific to one or the other. For example: 
+
+* `$lazy()` converts a `DataFrame` to a `LazyFrame`, and therefore only exists 
+  for `DataFrame`;
+* `$collect()` executes a lazy query, which means it can only be applied on 
+  a `LazyFrame`.
```
eitsupi, line 35: Actually in Python these can exist in both classes.
However, in R, we can probably use the `as_polars_*` functions for that purpose, so we don't need to create these methods in both.

etiennebacher: Ok what if I replace the first point with "$equals() exists for `DataFrame` but not for `LazyFrame`"?

https://docs.pola.rs/py-polars/html/search.html?q=equals#

PR #697, opened 2024-01-14
fix: rename pl$options' Rd file and enforce named arguments

eitsupi:
- Rename the Rd file
- The documentation stated that named arguments were required, but this was not the case. This is corrected.


etiennebacher (approved):
LGTM, thanks

PR #700, opened 2024-01-14
docs(website): do not use directory urls

### LINKED ISSUE #509: Set `use_directory_urls: false` for the Mkdocs website to mimic pkgdown site
Work around for r-lib/downlit#182

Currently, we cannot enable automatic links to reference pages via `downlit` due to the use of pretty URLs.
However, I’m wondering if we can execute automatic links via `downlit` by setting up `pkgdown.yml` in the site and discontinuing the use of pretty URLs.
https://www.mkdocs.org/user-guide/configuration/#use_directory_urls

“If this is set up, it should be possible to automatically link from the pkgdown documentation of packages such as tidypolars and polarssql to the polars documentation.

etiennebacher:
Close #509


etiennebacher:
Actually I need to check if we need to update some links

README.md
```diff
@@ -83,20 +84,22 @@ df$sort("fruits")$select(
...
+#> │ ---    ┆ ---    ┆ ng_fruits    ┆ --- ┆ s            ┆ its          ┆ its          ┆ by_fruits    │
+#> │ str    ┆ str    ┆ ---          ┆ i32 ┆ ---          ┆ ---          ┆ ---          ┆ ---          │
+#> │        ┆        ┆ str          ┆     ┆ i32          ┆ i32          ┆ i32          ┆ i32          │
+#> ╞════════╪════════╪══════════════╪═════╪══════════════╪══════════════╪══════════════╪══════════════╡
+#> │ apple  ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 7            ┆ 4            ┆ 4            │
+#> │ apple  ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 7            ┆ 3            ┆ 3            │
+#> │ banana ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 8            ┆ 5            ┆ 5            │
+#> │ banana ┆ audi   ┆ fruits       ┆ 11  ┆ 2            ┆ 8            ┆ 2            ┆ 2            │
+#> │ banana ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 8            ┆ 1            ┆ 1            │
+#> └────────┴────────┴──────────────┴─────┴──────────────┴──────────────┴──────────────┴──────────────┘
```
eitsupi, line 98: I think this is an unrelated change.

eitsupi (changes requested):
Thanks, could you add the `pkgdown.yml` file as a follow up work in another PR?
(Or I may be able to do it.)

Something like this: https://github.com/r-lib/pkgdown/blob/09457ac814017f0dac83219553d95c08368d55b7/pkgdown.yml

README.md
etiennebacher: Should be reverted now

etiennebacher:
> Thanks, could you add the pkgdown.yml file as a follow up work in another PR?

I added `pkgdown.yml` but I removed some stuff that seemed unnecessary in our context. Is this so that `downlit` knows where to link the functions to?

eitsupi:
> Is this so that `downlit` knows where to link the functions to?

Yes, downlit will check to see if `pkgdown.yml` is on the URL in the DESCRIPTION file, and if so, it will create links to the reference pages or articles there.
Ideally, it would be great if altdoc could automatically generate that file.

pkgdown.yml
eitsupi, line 1: Perhaps it should be placed in the altdoc directory instead of here? (Should eventually reside in the root of the website)
I checked the artifact in GitHubActions and it did not appear to be correctly placed.

eitsupi (changes requested):

pkgdown.yml
etiennebacher: But will it be found by `pkgdown` or `downlit` (I don't know which of the two does the linking) in this case? On the `pkgdown` repo, it's placed in the `pkgdown` folder. Maybe that's the only acceptable location for it?

eitsupi: I meant like these:

- https://pkgdown.r-lib.org/pkgdown.yml
- https://tidypolars.etiennebacher.com/pkgdown.yml

eitsupi: https://downlit.r-lib.org/index.html#cross-package-links

eitsupi (approved):
Thanks, let's see if the automatic link works......

PR #701, opened 2024-01-14
build: add functions to update usage sections in Rd files

### LINKED ISSUE #698: Fix usage sections in Rd files side
Related to this, we shoud update usage section like the followings:
https://github.com/pola-rs/r-polars/blob/396caf97ec9c21a639851bc0d357823c28771d40/docs/make-docs.R#L67-L69

Maybe something with this:

```r
content_rd <- tools::parse_Rd("man/DataFrame_clone.Rd")
tags <- tools:::RdTags(content_rd)
content_rd[[which(tags == "\\usage")]][[2]][1] <- content_rd[[which(tags == "\\usage")]][[2]][1] |>
  stringr::str_replace("^DataFrame_", "<DataFrame>$")
content_rd |>
  as.character(deparse = TRUE) |>
  readr::write_lines("man/DataFrame_clone.Rd", sep = "")
```

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/issues/560#issuecomment-1837412864_

eitsupi:
Close #698


eitsupi:
This works well, but has the problem that every time `devtool::document()` is run, all the Rd files are updated (and then restored by the replacement script), so the function takes a little longer to run.
However, I think it is acceptable.

eitsupi:
Apparently, such notation is not allowed in R CMD check, so it is not available.

dev/styler_utils.R
```diff
@@ -178,9 +178,90 @@ style_files = function(
...
+              rd_usage_replace(file, pattern, replacement),
+              error = function(e) {
+                cat("Error updating usage section in ", file, "\n", sep = "")
+              }
+            )
+          }
+        )
+    }
+  )
+}
```
eitsupi, line 262: @etiennebacher I think these functions will handle updates to the Usage section better than the current script for altdoc, but it requires storing the Rd file in a separate directory.
If you can change the directory altdoc refers to, you can switch to using these functions.

etiennebacher: What is better handled by this script?

eitsupi: Like this:
https://github.com/pola-rs/r-polars/blob/1f2a4d0494a0e90e49ccb72cc92e10bd82a0a42d/man/pl_options.Rd#L7-L19

image

etiennebacher: Ah yes, indeed, I'll improve the postprocessing script with yours

PR #703, opened 2024-01-14
docs: fix some documents lacking usage section

eitsupi:


eitsupi:
I don't think it's displayed in the altdoc sidebar for now.
We need to think carefully about how you handle DataType.

R/datatype.R
```diff
@@ -152,67 +152,17 @@ DataType_new = function(str) {
 #' @examples
 #' # constructors are finally available via pl$... or pl$dtypes$...
 #' pl$List(pl$List(pl$Int64))
-DataType_constructors = list(
+DataType_constructors = function() {
```
etiennebacher, line 155: do we still need this function if we make one separate function per datatype?

etiennebacher:
> I don't think it's displayed in the altdoc sidebar for now.

You need to add `DataType` in the list of classes in altdoc preprocessing and postprocessing scripts

R/datatype.R
eitsupi: These functions need to remain as they are because of their special processing, such as being stored in two locations, like `pl$List` and `pl$dtypes$List`.

eitsupi:
> You need to add `DataType` in the list of classes in altdoc preprocessing and postprocessing scripts

I'm sorry, but I don't know where to put it in the sidebar, so can you merge that as it is now as a follow-up task?

etiennebacher (approved):

PR #705, opened 2024-01-14
chore: use roxygen2 7.3.0

eitsupi:
Update roxygen2 to the latest version.

- Export S3 methods reported by roxygen2
- `"_PACKAGE"` and `@docType "package"` work like same now, so if there are two of them, one of them must be deleted.
  The extendr-wrapper.R file cannot be updated here, so as a workaround, the polars-package file was deleted.
  If this issue (extendr/extendr#684) is resolved upstream, the polars-package file needs to be reinstated.


etiennebacher (commented):
Why do you want to export all these methods? Is it recommended by `roxygen2`?

etiennebacher:
Hum I see, I think this is going to clutter the namespace with useless methods (for an external user). But at the same time it's weird to make an S3 method just for internal use, no? Maybe we need to refactor those instead?

etiennebacher:
Or maybe we can use `@exportS3Method NULL` as said in the blogpost?

eitsupi:
> I think this is going to clutter the namespace with useless methods (for an external user).

Does this affect namespace? I've compared the two versions, the release version and this PR version, and the behavior doesn't seem to have changed.

etiennebacher:
Ok, I haven't tested locally, my bad

etiennebacher (approved):

eitsupi:
https://github.com/r-lib/devtools/issues/2293#issuecomment-733612670

PR #706, opened 2024-01-16
Better handling of int64

### LINKED ISSUE #465: Improve handling of Polars Int64 to R
Discussed in #455

Since R doesn't have the Int64 type and the `bit64` package is not a hard dependency, it would be great if they could improve the process of converting Int64 to the R type.

etiennebacher:
Close #465


src/rust/src/conversion_s_to_r.rs
```diff
@@ -24,34 +24,52 @@ use pl::PolarsError as pl_error;
...
+                                let x = f64::from_bits(x as u64);
+                                Some(x)
+                            }
+                        })
+                        .collect_robj()
+                        .set_class(&["integer64"])
+                        .expect("internal error could not set class label 'integer64'")
+                }).ok(),
+                _ => panic!("foo"),              
+            },
```
etiennebacher, line 72: this part won't compile

etiennebacher:
@eitsupi @sorhawell I'm not sure this is the good way to address the int64 issue but it works, what do you think?

Note that I took `bigint_conversion` mostly as a temporary name, no problem to change it (I also picked it because it will be easy to search and replace).

R/options.R
```diff
@@ -62,9 +67,15 @@ polars_optreq$rpool_cap = list() # rust-side options already check args
 #' @param no_messages Hide messages.
 #' @param rpool_cap The maximum number of R sessions that can be used to process
 #' R code in the background. See Details.
+#' @param bigint_conversion How should Int64 values be handled when converting a
+#' polars object to R?
+#'
+#' * `"float"` (default) converts the values to Float64.
```
eitsupi, line 73: I am confused. R doesn't have a float64 type, right?

I think it should be `numeric` instead of `float`.
For example, RPostgres' `bigint` option.
https://github.com/r-dbi/RPostgres/blob/f15f8f445def9e24ff98e4bfc910efcf705c85aa/R/dbConnect_RedshiftDriver.R#L5

eitsupi (changes requested):
Thank you for working on this.
But I think we should not set options here with `pl$set_options()`, but use R's standard `options()` function.
The `options` function is widely used, and its usage such as temporary switching with something like `withr` is popular.

- `arrow` uses `arrow.int64_downcast`
  https://github.com/apache/arrow/blob/55afcf0450aa2b611e78335bdbfd77e55ae3bc9f/r/src/array_to_vector.cpp#L1307-L1314

R/options.R
etiennebacher: Actually "numeric" is misleading since an integer is also numeric in R. How about "double"?

``` r
is.numeric(1L)
#> [1] TRUE
is.double(1L)
#> [1] FALSE
```

etiennebacher:
>  But I think we should not set options here with pl$set_options(), but use R's standard options() function.

I'd like to continue this discussion but this is out of scope for this PR. Can you open an issue for this?

> arrow uses arrow.int64_downcast

This name is associated with TRUE/FALSE (do we downcast int64? yes or no). Here we have more choices so I don't think this name is suitable.

R/options.R
eitsupi: That sounds good, thanks.

etiennebacher:
>  bigint_conversion is not bad for me, but int64_conversion might be better as I think Polars terminology is int64 not bigint.

Fine for me, I changed both the value "float" and the arg name

etiennebacher:
I would like to have @sorhawell's opinion before merging

R/expr__expr.R
```diff
@@ -3177,17 +3177,18 @@ Expr_rep_extend = function(expr, n, rechunk = TRUE, upcast = TRUE) {
 #' Otherwise, provide a DataFrame that the Expr should be evaluated in.
 #' @param i Numeric column to extract. Default is zero (which gives the first
 #' column).
+#' @inheritParams pl_set_options
 #' @return R object
 #' @examples
 #' pl$lit(1:3)$to_r()
-Expr_to_r = function(df = NULL, i = 0) {
+Expr_to_r = function(df = NULL, i = 0, int64_conversion  = pl$options$int64_conversion ) {
```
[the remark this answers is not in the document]
etiennebacher: done

src/rust/src/conversion_s_to_r.rs
```diff
@@ -24,39 +24,49 @@ use pl::PolarsError as pl_error;
...
+                                let x = crate::utils::BIT64_NA_ECODING;
+                                let x = f64::from_bits(x as u64);
+                                Some(x)
+                            }
+                        })
+                        .collect_robj()
+                        .set_class(&["integer64"])
+                        .expect("internal error could not set class label 'integer64'")
+                }),
+                _ => panic!("`int64_conversion ` must be one of 'float', 'string', 'bit64'"),
```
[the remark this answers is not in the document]
etiennebacher: done

eitsupi (approved):

R/options.R
```diff
@@ -152,14 +164,17 @@ pl_reset_options = function() {
...
 
 
 translate_failures = \(x) {
   lookups = c(
     "must_be_scalar" = "Input must be of length one.",
     "must_be_integer" = "Input must be an integer.",
-    "must_be_bool" = "Input must be TRUE or FALSE"
+    "must_be_bool" = "Input must be TRUE or FALSE.",
+    "acceptable_choices" = "`int64_conversion ` must be one of \"float\", \"string\", \"bit64\".",
+    "bit64_is_attached" = "Package `bit64` must be attached to use `int64_conversion = \"bit64\"`."
```
eitsupi, line 177: How about use raw strings here?

etiennebacher: I don't think it would add much

PR #709, opened 2024-01-17
Implement `$list$any()` and `$list$all()`

etiennebacher:
The last row in the [python example](https://docs.pola.rs/docs/python/dev/reference/expressions/api/polars.Expr.list.all.html#polars.Expr.list.all) is only `null` but I don't know how to do it from a list


eitsupi (approved):
Thanks!

PR #714, opened 2024-01-19
refactor!: remove deprecated functions for version 0.13.0

eitsupi:


NEWS.md
```diff
@@ -2,6 +2,12 @@
 
 ## polars (development version)
 
+### Breaking changes
+
+-   Depricated functions from 0.12.x are removed (#714).
```
etiennebacher, line 7:
```suggestion
-   Deprecated functions from 0.12.x are removed (#714).
```

etiennebacher (commented):
Thanks, LGTM but I think we can keep that aside and merge it just before 0.13.0 so that we can release 0.12.3 in the meantime if we need to

eitsupi:
> but I think we can keep that aside and merge it just before 0.13.0 so that we can release 0.12.3 in the meantime if we need to

Good point. But I think it would be better to make the next version 0.13.0, since it already pulls in breaking changes.

etiennebacher (approved):

PR #718, opened 2024-01-20
refactor!: remove `<Expr>$where`

eitsupi:
Same as pola-rs/polars#13440

Perhaps this is not used as often as it should be, and I would like to remove it rather than deprecate it.


etiennebacher:
I'd rather deprecate it than removing it at once. At least people can have a helpful message about how to fix their code during a few weeks and it doesn't cost much to us

eitsupi:
> I'd rather deprecate it than removing it at once. At least people can have a helpful message about how to fix their code during a few weeks and it doesn't cost much to us

Ideally, yes, but our development resources are limited.
I also imagine that there would be no problem removing this function immediately, given that there would be few users.

eitsupi:
Personally, the deprecation of Python Polars is of little use to me because the release rate is too fast.
Often the code I run after several months is no longer working without issuing a warning. I end up having to search GitHub to fix the code, so it makes no difference to me whether I go through the deprecation or not.

I find the deprecation process to be of little value unless it is done over a sufficiently long period of time. (For example, the Tidyverse packages take years to remove deprecated functions, right?)

I do not want to go through deprecation on this, as I believe the priority now is development speed. (Of course I would prefer to go through deprecation for very commonly used functions, but as for where, I am sure almost no one uses it.)

etiennebacher:
I'm fine with that, thanks for the explanations (I didn't even know `$where()` before actually)

etiennebacher (approved):

eitsupi:
(It appears that some whitespace has been removed as a result of auto-formatting.)

PR #719, opened 2024-01-20
refactor!: remove the `eager` option from `<SQLContext>$execute()`

eitsupi:
I think it is not very desirable for the return type to change depending on the value of the argument.
It is safe to always return a LazyFrame, since it costs very little to build one.


etiennebacher (commented):
I don't understand why this change is made: it's already `eager = FALSE` by default so we don't have the risk of accidentally collecting, and it would divert from the py-polars API (which is fine sometimes but here I just don't see why).

Do you have some usecases where this argument was a problem?

eitsupi:
> I don't understand why this change is made: it's already `eager = FALSE` by default so we don't have the risk of accidentally collecting, and it would divert from the py-polars API (which is fine to do sometimes but here I just don't see why).
> 
> Do you have some usecases where this argument was a problem?

- I think the act of changing the return type depending on the argument is generally not a good thing (it should separate the function).
- This option merely means that `$collect()` can be omitted, and by complicating the function, it increases the number of test cases and makes it easier to make bugs. (In fact, there was no test case here.)

etiennebacher (approved):
Alright, fine for me, thanks

eitsupi:
Thanks for your review!

Issue #713, opened 2024-01-19
Set options via the `options` function


eitsupi:
Perhaps something like `pl$set_options` should work as a shortcut for the options function.

etiennebacher:
For options and envvars, maybe it would indeed make more sense to diverge from polars since R users are used to `options()`, `Sys.setenv()`, and the `withr` functions for advanced users.

What I really dislike about `options()` and `Sys.setenv()` is that I find it hard to easily get a list of options / envvars available for a package. If we add support for those, we should make sure those are easy to find. One way would be to add `polars_options()` and `polars_envvars()` that would complement `polars_info()`. Those would just print a list of envvars/options with their description (like `polars_info()`).

In this case maybe it wouldn't make sense to keep `pl$set_options()` and `pl$options` in the long run, and those could be deprecated
---

Note: polars envvars: https://github.com/pola-rs/polars/blob/62ecdd3e14e485baa84750b92a6615bc8d680b54/py-polars/polars/config.py#L45

PR #720, opened 2024-01-20
feat: add `disable_limit_max_threads` feature and limit the max threads to 2 by default for CRAN compatibility

### LINKED ISSUE #241: how to config max number of threads?

eitsupi:
Close #241


R/zzz.R
```diff
@@ -116,6 +116,17 @@ move_env_elements(RPolarsExpr, pl, c("lit"), remove = FALSE)
 
 
 .onLoad = function(libname, pkgname) {
+  # Auto limit the max number of threads used by polars
+  if (
+    !(cargo_rpolars_feature_info()[["disable_auto_limit_max_threads"]] ||
+      isTRUE(getOption("polars.disable_auto_limit_max_threads"))) &&
```
eitsupi, line 122: Related to #713, here `disable_auto_limit_max_threads` must be set via the `options` function.
Functions in the package `polars` are not useful here, since the value must be set when Polars is invoked.

eitsupi:
@etiennebacher Could you take a look at this?

NEWS.md
```diff
@@ -27,6 +27,12 @@
 
 ### What's changed
 
+-   For compatibility with CRAN, the number of threads used by Polars may be automatically set to 2 (#720).
```
etiennebacher, line 30:
```suggestion
-   For compatibility with CRAN, the number of threads used by Polars is automatically set to 2 (#720).
```

NEWS.md
```diff
@@ -27,6 +27,12 @@
...
+    To disable this behavior and have the maximum number of threads used automatically,
+    one of the following ways can be used:
+    -   Build the Rust library with the `disable_auto_limit_max_threads` feature.
+    -   Set the `polars.disable_auto_limit_max_threads` option to `TRUE` with the `options()` function
+        before loading the package.
```
etiennebacher, line 35: We could mention `Sys.setenv(POLARS_MAX_THREADS =)` so that users know that they don't have only the choice of 2 threads or all of them, no?

NEWS.md
```diff
@@ -27,6 +27,12 @@
 
 ### What's changed
 
+-   For compatibility with CRAN, the number of threads used by Polars may be automatically set to 2 (#720).
+    To disable this behavior and have the maximum number of threads used automatically,
+    one of the following ways can be used:
+    -   Build the Rust library with the `disable_auto_limit_max_threads` feature.
+    -   Set the `polars.disable_auto_limit_max_threads` option to `TRUE` with the `options()` function
```
etiennebacher, line 34: I find it hard to understand what `disable_auto_limit_max_threads = TRUE` does, can we rename if `use_max_threads`? It's clearer to set `polars.use_max_threads = TRUE`

R/info.R
```diff
@@ -77,10 +81,21 @@ check_feature = function(feature_name, context = NULL, call = sys.call(1L)) {
...
+#' For compatibility with CRAN, the threadpool size is set to 2 by default.
+#' To disable this behavior and let the engine determine the threadpool size,
+#' one of the following ways can be used:
+#'
+#' - Enable the `disable_auto_limit_max_threads` feature of the library.
+#'   This can be done by setting the feature flag when installing the package.
+#'   See the installation vignette (`vignette("install", "polars")`)
+#'   for details.
+#' - Set the `polars.disable_auto_limit_max_threads` option to `TRUE` with
+#'   the [options()] function.
```
etiennebacher, line 97: this must be done before loading the package, right? can you mention it here?

vignettes/install.Rmd
```diff
@@ -125,7 +125,17 @@ At this time, the following environment variables can be used to change the Rust
 The feature flags can be set by the `LIBR_POLARS_FEATURES` environment variable.
 For example, to enable the `full_features` feature, set the environment variable as like `LIBR_POLARS_FEATURES="full_features"`.
 
-Note that these features requires the Rust nightly toolchain `r rust_toolchain_version`.
+Currently `full_features` would work as a combination of the following features.
+
+- Features for CRAN compatibility
+  - `disable_auto_limit_max_threads`, this feature disables the automatic limit of the maximum number of threads to 2 for CRAN compatibility,
+    and the maximum number of threads is used by default.
+    In other words, if this feature is enabled, it is not compatible with CRAN.
```
etiennebacher, line 133: I think this warning is only for people that want to use `polars` in other packages and publish them to CRAN. This sentence could be less "scary" for standard R users

etiennebacher (changes requested):

NEWS.md
eitsupi: Good point.
I was wondering if it would be worth aligning the name with feature, but it would definitely be easier to understand if the option name was changed.

NEWS.md
eitsupi: How about 87d5eebaa8ba8e04cedab5827c6c3277bdc006fc?

NEWS.md
eitsupi: I changed it to `polars.limit_max_threads`.
Given that the environment variable is `POLARS_MAX_THREADS`, I think the term `max` confusing.

tests/testthat/_snaps/options.md
```diff
@@ -23,5 +24,5 @@
     - strictly_immutable: input must be TRUE or FALSE.
     - debug_polars: input must be TRUE or FALSE.
     
-    More info at `?polars_options`.
+    More info at `?polars::polars_options`.
```
eitsupi, line 27: Note: This is not directly related to this PR, but it is necessary to be able to display the help page without loading the package because the package cannot be loaded if the option validation fails.

etiennebacher:
Just fyi I'll try to do another review tomorrow or friday, very busy week

R/info.R
```diff
@@ -77,10 +81,22 @@ check_feature = function(feature_name, context = NULL, call = sys.call(1L)) {
...
-#' (The thread pool is not behind a lock, so it cannot be modified once set).
+#' It cannot be modified once `polars` is loaded.
 #' It is strongly recommended not to override this value as it will be
 #' set automatically by the engine.
 #'
+#' For compatibility with CRAN, the threadpool size is set to 2 by default.
+#' To disable this behavior and let the engine determine the threadpool size,
+#' one of the following ways can be used:
+#'
+#' - Enable the `disable_auto_limit_max_threads` feature of the library.
```
etiennebacher, line 92: for consistency I think this feature should also be renamed `limit_max_threads` on the Rust side

R/options.R
```diff
@@ -179,6 +175,7 @@ polars_options = function() {
     df_knitr_print = getOption("polars.df_knitr_print"),
     do_not_repeat_call = getOption("polars.do_not_repeat_call"),
     int64_conversion = getOption("polars.int64_conversion"),
+    limit_max_threads = getOption("polars.limit_max_threads") %||% "Not set (NULL)",
```
etiennebacher, line 178: if the default is to limit the number of threads to 2, shouldn't it be `%||% TRUE`?

R/options.R
```diff
@@ -202,7 +199,8 @@ polars_options_reset = function() {
       polars.no_messages = FALSE,
       polars.rpool_active = 0,
       polars.rpool_cap = 4,
-      polars.strictly_immutable = TRUE
+      polars.strictly_immutable = TRUE,
+      polars.limit_max_threads = NULL
```
etiennebacher, line 203: can you put options in alphabetical order?

R/utils.R
```diff
@@ -20,7 +20,8 @@ options(
   polars.int64_conversion = "double",
   polars.maintain_order = FALSE,
   polars.no_messages = FALSE,
-  polars.strictly_immutable = TRUE
+  polars.strictly_immutable = TRUE,
+  polars.limit_max_threads = NULL
```
etiennebacher, line 24: ditto

R/zzz.R
```diff
@@ -116,6 +116,17 @@ move_env_elements(RPolarsExpr, pl, c("lit"), remove = FALSE)
 
 
 .onLoad = function(libname, pkgname) {
+  # Auto limit the max number of threads used by polars
+  if (
+    isFALSE(cargo_rpolars_feature_info()[["disable_auto_limit_max_threads"]]) &&
+      !isFALSE(getOption("polars.limit_max_threads")) &&
```
etiennebacher, line 122:
```suggestion
      isTRUE(getOption("polars.limit_max_threads")) &&
```

src/rust/Cargo.toml
```diff
@@ -10,7 +10,8 @@ crate-type = ['staticlib']
 
 [features]
 default = []
-full_features = ["simd", "sql"]
+full_features = ["simd", "sql", "disable_auto_limit_max_threads"]
```
etiennebacher, line 13: cf comment above about renaming this feature

etiennebacher (changes requested):
The changes look good but I still have comments about the feature names and a few minor ones

R/info.R
eitsupi: You mean "disable_limit_max_threads"?
Note that this feature must be turned off by default (must not be included in the default feature)

R/zzz.R
eitsupi: This option is intended to only work when explicitly specified as `FALSE`

R/info.R
etiennebacher: Oh yes I have inverted it in my mind. I think the choices should be:

- enable the feature `disable_limit_max_threads` (or even `no_limit_max_threads`?)
- `options(polars.limit_max_threads = FALSE)`

R/options.R
eitsupi: TRUE doesn't make sense. This is because there is a possibility that the behavior will be changed on the lib side. Only an explicit FALSE is meaningful

etiennebacher: Sorry I don't understand that. Suppose I install the package from CRAN and I don't set any option anywhere. This means that there is an automatic limit to use 2 threads. If I do `polars_options()`, what would be printed?

`limit_max_threads    TRUE` or `limit_max_threads    Not set (NULL)`?

If it's `TRUE`, I think it's fine, but if it's `Not set (NULL)` then that's confusing because I would understand it as "there's no limit in the number of threads", while actually there is one.

R/info.R
eitsupi: Renamed to `disable_limit_max_threads`

R/options.R
eitsupi: The default is changed to `!polars_info()$features$disable_limit_max_threads`

etiennebacher (approved):
Thanks, this looks good to me, sorry for the delay

eitsupi:
Thanks for your detailed review!

R/info.R
```diff
@@ -77,10 +81,22 @@ check_feature = function(feature_name, context = NULL, call = sys.call(1L)) {
 #'
 #' The threadpool size can be overridden by setting the
 #' `POLARS_MAX_THREADS` environment variable before process start.
-#' (The thread pool is not behind a lock, so it cannot be modified once set).
+#' It cannot be modified once `polars` is loaded.
```
tdhock, line 84: thank you for documenting this

PR #724, opened 2024-01-20
docs: clean docs for `$list` methods

etiennebacher:
Style:

* remove useless `@format` and `@keywords`
* harmonize examples:
  * reuse same data when possible
  * use `with_columns()` instead of `select()` to show input and output side-by-side
  * when applicable, have an example with Expr and one with literal
* fix typos and weird sentences
* better titles

Other:

* deprecate `$list$lengths()`, use `$list$len()` instead
* remove the S3 `[` that was an alias for `$list$get()`, it just seemed weird to me to have this single method in there and I can't imagine it's easier to use. Ok to revert if necessary though


eitsupi (approved):
Great!

Could you update NEWS as a breaking change?

etiennebacher:
Thanks, done

PR #726, opened 2024-01-20
Add support for `options()` and deprecate `pl$set_options()`

### LINKED ISSUE #95: Refactor options.R into Config.R based on py-polars

### LINKED ISSUE #713: Set options via the `options` function

etiennebacher:
Close #713, close #95

This PR adds support for R `options()` that is generally used across packages, and deprecates `pl$set_options()` and `pl$reset_options()`.

### What are the main changes?

1. Option checks are no longer ran at "setting time" because we don't have control over `options()` so we can't throw an error if the option has a wrong value. Instead, options are now checked at run time, i.e when calling `polars_options()`. While this is not optimal, error should be raised early enough in a workflow because a lot of low-level functions call `polars_options()`.
1. options must be prefixed with `polars.`, e.g `pl$set_options(strictly_immutable = TRUE)` now is `options(polars.strictly_immutable = TRUE)`
1. `pl$set_options()` now throws a deprecation message, `pl$options` is removed and simply throws an error. I don't have a replacement for `pl$reset_options()` for now (it could simply be `polars_reset_options()`, but I don't think it should stay in the `pl` object). Edit: I've added `polars_options_reset()` (and not `polars_reset_options()` so that users can see the suggestion "polars_options_reset" when they search "polars_options")

A few minor things:

- it's important to indicate to users as much as possible that options description is available at `?polars_options()`. Currently this is displayed when printing `polars_info()` and in the error message for bad inputs.

### TODO

- [x] move option description to `polars_info()`
- [x] add deprecation message in `pl$set_options()`
- [x] ~write a vignette about options (including envvars when those are properly documented). Options are one of the few things that differ between R and python polars API so it seems important to me to have a dedicated document~ I'd like to address this in a followup PR, this one is quite big already
- [x] add snapshot test to ensure error is correctly displayed when several options are incorrect
- [x] should I replace the `getOption()` calls with simply `polars_options()$OP`? That way the option checks are run more often
- [x] add test that this works with `withr`
- [x] rename the currently used polars.df_print to something like polars.df_knitr_print and display this within polars_options as well.
- [x] use `withr` in all tests where options are modified

@eitsupi any suggestions?


eitsupi (commented):
Thanks for working on this. I have not looked at it in detail but it looks good.

One thing we should do together is to rename the currently used `polars.df_print` to something like `polars.df_knitr_print` and display this within `polars_options` as well.

https://github.com/pola-rs/r-polars/blob/fd7e03c94d2636fce23207d2a2e525d6fe12765d/R/pkg-knitr.R#L21

R/zzz.R
```diff
@@ -116,6 +116,20 @@ move_env_elements(RPolarsExpr, pl, c("lit"), remove = FALSE)
...
+    polars.debug_polars = FALSE,
+    polars.df_knitr_print = "auto",
+    polars.do_not_repeat_call = FALSE,
+    polars.int64_conversion = "double",
+    polars.maintain_order = FALSE,
+    polars.no_messages = FALSE,
+    polars.rpool_active = unwrap(get_global_rpool_cap())$active,
+    polars.rpool_cap = unwrap(get_global_rpool_cap())$capacity,
+    polars.strictly_immutable = TRUE
+  )
```
eitsupi, line 131: Wouldn't doing this here overwrite any options the user has set in Rprofile, etc.?

eitsupi (commented):
Awesome work, thanks!

However, I have a question about the behavior.

R/zzz.R
etiennebacher: Good point, I didn't think about that (I barely use Rprofile, and I think it's a bad idea to store options that modify so much the behavior there).

Do you know how other packages handle this?

etiennebacher: In any case, the value for `rpool_cap` is determined on the Rust side, so we shouldn't take into account what the user put in their Rprofile. For the rest, I'm exploring a bit more

eitsupi: > Do you know how other packages handle this?

I've seen the readr package have functions that wrap options.
Here. https://github.com/tidyverse/readr/blob/e529cb2775f1b52a0dfa30dabc9f8e0014aa77e6/R/utils.R#L24-L32

etiennebacher: I think [7b7d7f3](https://github.com/pola-rs/r-polars/pull/726/commits/7b7d7f3c6fdfc8b184664061ecb96f12baf151f3) fixed it. I tried to set options in .Rprofile in another project and it correctly got their values on startup. I don't really know how to properly test this though, there's no `withr::with_rprofile()`...

R/utils.R
```diff
@@ -1,3 +1,29 @@
...
+#' @noRd
+options(
+  polars.debug_polars = FALSE,
+  polars.df_knitr_print = "auto",
+  polars.do_not_repeat_call = FALSE,
+  polars.int64_conversion = "double",
+  polars.maintain_order = FALSE,
+  polars.no_messages = FALSE,
+  polars.strictly_immutable = TRUE
+)
```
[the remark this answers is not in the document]
etiennebacher: Yes this is used when running `devtools::load_all()` or when building the package. The reason is that `polars_info()` is called in several functions, so when we load those functions we need to have some initial values for options. As said in the comments, those are not necessarily the same as default options, which are set in `.onLoad()`

man/S3_knit_print.Rd
```diff
@@ -4,7 +4,7 @@
 \alias{knit_print.RPolarsDataFrame}
 \title{knit print polars DataFrame}
 \usage{
-\method{knit_print}{RPolarsDataFrame}(x, ...)
+knit_print.RPolarsDataFrame(x, ...)
```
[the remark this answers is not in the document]
etiennebacher: Right, it got included by mistake

tests/testthat/helper.R
```diff
@@ -25,7 +25,7 @@ expect_different = function(x, y) {
 expect_grepl_error = function(expr, expected_err = NULL, do_not_repeat_call = TRUE, ...) {
   # turn of including call in err msg
   if (do_not_repeat_call) {
-    pl$set_options(do_not_repeat_call = TRUE)
+    options(polars.do_not_repeat_call = TRUE)
```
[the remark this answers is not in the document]
etiennebacher: Indeed, but I think this helper could be removed in another PR, I don't think it adds something compared to `expect_error()`

R/utils.R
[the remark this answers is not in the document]
eitsupi: Ok, thanks for details.

eitsupi (approved):
Thanks. Let's try merging this.

man/nanoarrow.Rd
```diff
@@ -16,13 +16,13 @@
 \alias{as_record_batch_reader.RPolarsDataFrame}
 \title{polars to nanoarrow and arrow}
 \usage{
-\method{as_nanoarrow_array_stream}{RPolarsDataFrame}(x, ..., schema = NULL)
+as_nanoarrow_array_stream.RPolarsDataFrame(x, ..., schema = NULL)
```
eitsupi, line 19: Unnecessary changes...

etiennebacher: That's automatically made by devtools::document, I don't know why it changes that, I have R and packages up to date

eitsupi: Oh, I didn't know you didn't know this.
We have to load some packages before updating the Rd file or it will not be recognized some funtions as S3 methods.

https://github.com/pola-rs/r-polars/blob/870cb3ac7b7aa144b5d67f2a184eb6e61de7e3be/R/zzz.R#L148-L153

This is one of the reasons why we have to use Taskfile for development.

https://github.com/pola-rs/r-polars/blob/870cb3ac7b7aa144b5d67f2a184eb6e61de7e3be/Taskfile.yml#L10
https://github.com/pola-rs/r-polars/blob/870cb3ac7b7aa144b5d67f2a184eb6e61de7e3be/Taskfile.yml#L21
https://github.com/pola-rs/r-polars/blob/870cb3ac7b7aa144b5d67f2a184eb6e61de7e3be/Taskfile.yml#L177-L180

etiennebacher: thanks, I'll load those in the Rprofile then, that should solve it

PR #727, opened 2024-01-21
feat: rewrite `arrow_to_rdf` and `arrow_to_rseries_result`

eitsupi:
Fix #725, part of #677

After rewriting the contents of the functions with reference to Python Polars, the bug #725 seems to have been fixed.
(I'm not sure where the bug was occurring after all)

Other improvements:

- `as_polars_df` for `RecordBatch` now works.
- `as_polars_series` for `Array` and `ChunkedArray` are added.

These changes have completely ported `pl$from_arrow`'s functionality to `as_polars_*`, so we can deprecate `pl$from_arrow` in a follow-up PR.

PR #728, opened 2024-01-21
refactor!: deprecate `pl$from_arrow` in favor of `as_polars_df` and `as_polars_series`

### LINKED ISSUE #677: `pl$from_arrow()` v.s. `as_polars_df()`
I suspect that things like the `pl$from_arrow()` function are due to Python syntax/conventions and should not be brought into R.
In R it is common to handle such things by dispatching using the S3 (or S7 in the future) method, and I think we can eliminate `pl$from_arrow` in favor of `as_polars_df.Table`.

Related to #497

eitsupi:
Close #677

A follow up for #727

Python Polars' `polars.from_arrow` takes a `pyarrow.Table` or `pyarrow.Array` (or something) as an argument and returns a DataFrame or Series.
In R, it is more natural to use the S3 method dispatch than to do conditional branching with if depending on the argument type.


R/convert.R
```diff
@@ -30,6 +32,8 @@ pl_from_arrow = function(
     rechunk = TRUE,
     schema = NULL,
     schema_overrides = NULL) {
+  warning("`pl$from_arrow()` is deprecated and will be removed in 0.14.0. Use `as_polars_df()` or `as_polars_series` insead.")
```
etiennebacher, line 35:
```suggestion
  warning("`pl$from_arrow()` is deprecated and will be removed in 0.14.0. Use `as_polars_df()` or `as_polars_series()` insead.")
```

etiennebacher (approved):
thanks!

R/convert.R
eitsupi, line 35:
```suggestion
  warning("`pl$from_arrow()` is deprecated and will be removed in 0.14.0. Use `as_polars_df()` or `as_polars_series()` insead.")
```

eitsupi: Oops, looks like you guys were quick to find it (but I was committed faster than I realized it)

PR #730, opened 2024-01-22
feat: support `as_polars_*` for nanoarrow package objects

eitsupi:
Part of #497


eitsupi:
@paleolimbot Could you take a look at this?
`as_polars_series.nanoarrow_array` seems to work in a simple example, but is there anything else I should test?

src/rust/src/arrow_interop/to_rust.rs
```diff
@@ -8,11 +8,66 @@ use polars_core::utils::arrow::ffi;
...
+    }
+}
+
+pub trait RPackage {
+    fn export_array_func(&self) -> Result<Function, String>;
+}
+
+impl RPackage for ArrowRPackage {
+    fn export_array_func(&self) -> Result<Function, String> {
+        R!("function(array, exportable_array, exportable_schema) arrow:::ExportArray(array, exportable_array, exportable_schema)")?
```
paleolimbot, line 49: I think that `array$export_to_c()` is the intended usage...I don't think there are any plans to change `ExportArray` but it's an internal/undocumented function and there's no guarantee it will always be there.

paleolimbot (commented):
Very cool!

No need to do so right now, but if I remember correctly, a `Series` comes in chunks and you could in theory convert a `nanoarrow_array_stream()` into one. The nanoarrow that's about to be released has a method for `as_nanoarrow_array()` for a `ChunkedArray` for exactly that purpose.

eitsupi:
Thanks for the review!
Currently the function for converting chunked arrays to Series seems to be written entirely for the `arrow` package, so it looks like it needs to be extended for nanoarrow_array_stream.

https://github.com/pola-rs/r-polars/blob/1a6521c5a9622d432b8676bf45782515b2059863/R/construction.R#L143-L189

By the way, is there any way to get the nanoarrow type as a string representation?
Something like the following:

```r
> arrow::int16()$ToString()
[1] "int16"
```

I imagine that such functionality in the following areas would simplify the process when a package does not exist.

https://github.com/pola-rs/r-polars/blob/1a6521c5a9622d432b8676bf45782515b2059863/R/construction.R#L130-L133

R/construction.R
```diff
@@ -127,11 +127,8 @@ is_arrow_struct = \(x) {
 coerce_arrow = function(arr, rechunk = TRUE) {
   if (!is.null(arr$num_chunks) && is_arrow_dictonary(arr)) {
     # recast non ideal index types
-    non_ideal_idx_types = list(
-      arrow::int8(), arrow::uint8(), arrow::int16(),
-      arrow::uint16(), arrow::int32()
-    )
-    if (arr$type$index_type %in_list% non_ideal_idx_types) {
+    non_ideal_idx_types = c("int8", "uint8", "int16", "uint16", "int32")
```
paleolimbot, line 130: Should int64 and uint64 be in here, too?

eitsupi: This is a copy of https://github.com/pola-rs/polars/blob/25537c2d3f83422790fde50c6c7971e906f238e4/py-polars/polars/utils/_construction.py#L1790-L1806, and it says:

> small integer keys can often not be combined, so let's already cast to the uint32 used by polars

So, it seems intended.

paleolimbot: I definitely may be misunderstanding the code, but in Arrow you could theoretically have an int64 or uint64 index also (i.e., if you're headed from arrow -> Polars, you might want to cast those too)

eitsupi:
Hmmm, I did a little research and it seems that Polars does not have an API for creating Series from array streams.

eitsupi:
Thanks for your response.
Yeah, copying your `as_chunked_array.nanoarrow_array_stream` seems like a good idea for now.

R/as_polars.R
```diff
@@ -329,3 +329,24 @@ as_polars_series.nanoarrow_array = function(x, name = NULL, ...) {
...
+
+#' @rdname as_polars_series
+#' @export
+as_polars_series.nanoarrow_array_stream = function(x, name = NULL, ...) {
+  on.exit(x$release())
+
+  list_of_arrays = nanoarrow::collect_array_stream(x, validate = FALSE)
+
+  if (length(list_of_arrays) < 1L) {
+    out = pl$Series(NULL, name = name)
```
paleolimbot, line 341: A zero-size list of arrays should probably keep the type? You can get the type before `collect_array_stream()` with `x$get_schema()`.

R/extendr-wrappers.R
```diff
@@ -1252,7 +1252,9 @@ RPolarsSeries$to_frame <- function() .Call(wrap__RPolarsSeries__to_frame, self)
 
 RPolarsSeries$set_sorted_mut <- function(descending) invisible(.Call(wrap__RPolarsSeries__set_sorted_mut, self, descending))
 
-RPolarsSeries$from_arrow <- function(name, array) .Call(wrap__RPolarsSeries__from_arrow, name, array)
+RPolarsSeries$from_arrow_array_stream_str <- function(name, robj_str) .Call(wrap__RPolarsSeries__from_arrow_array_stream_str, name, robj_str)
```
eitsupi, line 1255: I was going to use this as the inside of `as_polars_series`, but now I can't because it causes a segmentation fault.
See https://github.com/pola-rs/r-polars/issues/732#issuecomment-1916599082.

R/as_polars.R
```diff
@@ -206,7 +206,43 @@ as_polars_df.ArrowTabular = function(
...
+      }
+    }
+
+    out = do.call(pl$DataFrame, data_cols)
+  } else {
+    out = pl$DataFrame() # TODO: support creating 0-row DataFrame
+  }
+
+  out
+}
```
eitsupi, line 245: I think we should add `as_polars_df.nanoarrow_array` and use that inside of `as_polars_df.nanoarrow_array_stream`. (concat all DataFrames)

May require the `vstack` method.
https://stackoverflow.com/questions/71654966/how-can-i-append-or-concatenate-two-dataframes-in-python-polars

eitsupi:
I feel that no further work is needed on Rust's side, so I will merge this in as soon as possible and address the comments with follow-up PRs.
(NEWS file has not been updated yet, but will be updated once the method is fully implemented in a follow-up PRs)

etiennebacher (approved):
I'm not familiar with `nanoarrow` so can't see if this is correct or not, but it is tested so I suppose it's fine.

R/construction.R
eitsupi: I have created an issue for this #752

R/as_polars.R
eitsupi: > I am not sure that you want to force a concatenation until absolutely necessary!

Sorry, I didn't understand what you meant.

I am currently concatenating each column into a DataFrame after each column is concatenated, but just wondering if the order could be changed to concatenate the data frames after each chunk is concatenated into a DataFrame.

eitsupi:
At some stage I would like to write an article on exchanging data between packages via Arrow.
Python Polars has ADBC integration (via pyarrow), but we can do it here via nanoarrow. The R arrow package is not required.

R/as_polars.R
paleolimbot: I see...I think that's fine! I don't know the Polars details well enough. It would be optimal (but not required) if (for example) when importing a ChunkedArray that composed of 1000 chunks, the Polars representation would be a Series that also had 1000 chunks (I think that's a thing).

eitsupi: Yes, I will modify the function to control such behavior with the rechunk option like this.
https://github.com/pola-rs/r-polars/blob/eb4fbcc553ca80f65efbba9031ef41f8b2227722/R/construction.R#L14

PR #731, opened 2024-01-23
docs: regen Rd file and add a note about S3 method documentation

eitsupi:
A follow up for https://github.com/pola-rs/r-polars/pull/726#discussion_r1462740637


etiennebacher (approved):
Good clarification, thanks

PR #735, opened 2024-01-26
Implement `polars_envvars()`

etiennebacher:
Used this PR to also rename the R files for `polars_info()` and `polars_options()`.

For some envvars, that is trickier to do because `polars` calls a Rust function to set them:
https://github.com/pola-rs/polars/blob/c088edbe2ea576fb82985361425cf4a5591c91a1/py-polars/polars/config.py#L458-L462

However that's not something we can do with `Sys.setenv()`. I think we have two choices:

1. add a `pl$set_env()` or equivalent. I'm not a big fan of this because it breaks the consistency of `polars_envvars()` / `polars_info()` / `polars_options()`
2. ignore those envvars

@eitsupi what do you think?


eitsupi:
Thanks for working on this, it would be really great to have documentation and testing of this.

I think it is sufficient to mention in the documentation those things that are not meant to be set via Sys.setenv.

eitsupi (approved):

R/polars_envvars.R
```diff
@@ -0,0 +1,134 @@
...
+    c("POLARS_FMT_TABLE_CELL_NUMERIC_ALIGNMENT", "LEFT"),
+    c("POLARS_FMT_TABLE_DATAFRAME_SHAPE_BELOW", "0"),
+    c("POLARS_FMT_TABLE_FORMATTING", "UTF8_FULL_CONDENSED"),
+    c("POLARS_FMT_TABLE_HIDE_COLUMN_DATA_TYPES", "0"),
+    c("POLARS_FMT_TABLE_HIDE_COLUMN_NAMES", "0"),
+    c("POLARS_FMT_TABLE_HIDE_COLUMN_SEPARATOR", "0"),
+    c("POLARS_FMT_TABLE_HIDE_DATAFRAME_SHAPE_INFORMATION", "0"),
+    c("POLARS_FMT_TABLE_INLINE_COLUMN_DATA_TYPE", "0"),
+    c("POLARS_FMT_TABLE_ROUNDED_CORNERS", "0"),
+    c("POLARS_MAX_THREADS", unwrap(get_global_rpool_cap())$capacity),
```
eitsupi, line 105: @etiennebacher Isn't this `threadpool_size`?

etiennebacher: Ah right I'll fix this

Issue #732, opened 2024-01-24
There are multiple functions that do the same thing (for arrow)


eitsupi:
Some context can be found in #64

While looking into the Arrow implementation, I noticed that this package has several functions related to the Arrow C interface.

The current publicly available function for creating DataFrames from arrow Tables is `polars:::arrow_to_rdf` (ported from Python Polars) written in R inside `polars::as_polars_df` (formerly `polars::pl$from_arrow`), but here is a function written in Rust with almost the same functionality.

https://github.com/pola-rs/r-polars/blob/1a6521c5a9622d432b8676bf45782515b2059863/R/extendr-wrappers.R#L41

This function does not appear to be used anywhere.

Presumably this is a function created for experimental purposes and used in the benchmarks stored below.

https://github.com/pola-rs/r-polars/blob/1a6521c5a9622d432b8676bf45782515b2059863/inst/misc/examples/from_arrow.R

I ran this script with the current development version (on #730) and found that `pl$from_arrow` and `rb_list_to_df` are about the same speed, so perhaps `rb_list_to_df` can be removed.

```r
> source("/workspaces/r-polars/inst/misc/examples/from_arrow.R", encoding = "UTF-8")

Attaching package: ‘arrow’

The following object is masked from ‘package:utils’:

    timestamp

total n chunks: 475# A tibble: 4 × 13
  expression              min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result
  <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>
1 to_arrow           655.57ms    1.44s     0.518  915.77KB     1.45    10    28     19.31s <NULL>
2 from_arrow            827ms    1.74s     0.547    3.15MB     1.75    10    32     18.27s <NULL>
3 from_arrow_no_rec… 654.13ms 746.89ms     1.31   959.62KB     4.33    10    33      7.62s <NULL>
4 from_arrow_all_se…    1.08s    1.13s     0.827 1011.25KB     3.97    10    48     12.09s <NULL>
# ℹ 3 more variables: memory <list>, time <list>, gc <list>
# A tibble: 4 × 13
  expression   min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory    
  <bch:expr> <bch> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>    
[...]

PR #753, opened 2024-01-31
refactor: remove unused functions related to arrow

eitsupi:
Part of #732

- Remove unused internal R function `rb_list_to_df`
- Move some codes from `to_rust.rs` to `mod.rs` for future use.
- Remove some unused Rust functions and comments.

### DIFF
--- a/src/rust/src/arrow_interop/mod.rs
+++ b/src/rust/src/arrow_interop/mod.rs
@@ -1,1 +1,63 @@
 pub mod to_rust;
+
+use extendr_api::prelude::*;
+use std::result::Result;
+
+#[derive(Debug)]
+pub enum RArrowArrayClass {
+    ArrowArray,
+    NanoArrowArray,
+}
+
+impl<'a> FromRobj<'a> for RArrowArrayClass {
+    fn from_robj(robj: &Robj) -> std::result::Result<Self, &'static str> {
+        if robj.inherits("nanoarrow_array") {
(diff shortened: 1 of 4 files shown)


src/rust/src/arrow_interop/mod.rs
```diff
@@ -1 +1,63 @@
...
+        function(array, exportable_array, exportable_schema) {
+            nanoarrow::nanoarrow_pointer_export(
+                nanoarrow::infer_nanoarrow_schema(array),
+                exportable_schema
+            )
+            nanoarrow::nanoarrow_pointer_export(array, exportable_array)
+        }
+        "#)
+    }
+}
```
etiennebacher, line 63: Sorry if I miss sth obvious but what is the point of calling Rust functions that themselves only run R code? Can't we just run the whole in R in this case?

eitsupi: I didn't write this, but it has been like this for a long time. I've just rewritten it so that it can receive nanoarrow_array as well as arrow Array.

Another reason is that the C interface is not fully implemented in upstream Polars

etiennebacher: Ah ok this is called in another Rust function so I guess it's useful to have this part in Rust

etiennebacher (approved):

eitsupi:
Thanks for the review.

PR #5, opened 2023-01-03
Export ArrowArrayStream from polars data frame

paleolimbot:
Building on the excellent experiments of @sorhawell in https://github.com/rpolars/rpolars/issues/4 and https://github.com/rpolars/rpolars/compare/main...nanoarrow , this is an attempt to export data frames to the Arrow C Stream interface.

This doesn't compile yet, of course, but hopefully somebody who actually does Rust can help here! The error really does hit at the crux of the matter, which is that the `DataFrame` has to outlive the stream. In C++ one could do something like this using shared pointers and a virtual deleter...I'm lost as to how this should be done here but am totally willing to learn!

PR #754, opened 2024-01-31
refactor: rewrite S3 methods of `arrow` and `nanoarrow` package functions

eitsupi:
- `as_arrow_table` and `as_record_batch_reader` for RPolarsDataFrame no longer need `nanoarrow`
- Improve Rd files
- Split the test file


R/pkg-nanoarrow.R
```diff
@@ -31,31 +17,12 @@ as_nanoarrow_array_stream.RPolarsDataFrame = function(x, ..., schema = NULL) {
...
-#' @examples
-#' nanoarrow_array_schema = infer_nanoarrow_schema(df)
-#' print(nanoarrow_array_schema)
+
+#' Infer nanoarrow schema from a Polars object
+#'
+#' @rdname S3_infer_nanoarrow_schema
+# exported in zzz.R
 infer_nanoarrow_schema.RPolarsDataFrame = function(x, ...) {
+  # TODO: we may not use `as_nanoarrow_array_stream` here
```
eitsupi, line 26: See https://github.com/pola-rs/r-polars/pull/5#discussion_r1472918099

R/pkg-arrow.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' @param x [A Polars DataFrame][DataFrame_class]
+#' @param ... Ignored
+#' @rdname S3_as_arrow_table
+#' @examples
+#' library(arrow)
+#'
+#' pl_df = as_polars_df(mtcars)
+#' as_arrow_table(pl_df)
+# exported in zzz.R
```
etiennebacher, line 11:
```suggestion
# exported in zzz.R: this generic belongs to `arrow` and we only export the method if 
# `arrow` is already loaded when we load `polars`
```
just a note for future us

etiennebacher (approved):

R/pkg-arrow.R
eitsupi: There are similar things elsewhere, so if you are interested, please change them all by another PR.

PR #760, opened 2024-02-01
rustfmt.toml take precedence over ~/.rustfmt.toml

ssh352:
so that ~/rustfmt.toml doesn't affect r-polars formatting.


eitsupi (changes requested):
Thanks, could you add that file to the .Rbuildignore?

rustfmt.toml
eitsupi, line 1: Is this file name correct? Isn't it correct that it starts with a dot?

eitsupi: I see, thanks.

eitsupi (approved):

PR #762, opened 2024-02-01
feat: more arguments for `as.data.frame`

eitsupi:


R/s3_methods.R
```diff
@@ -229,17 +229,55 @@ dimnames.RPolarsLazyFrame = function(x) list(NULL, names(x))
 
 #' Convert to a data.frame
 #'
-#' @param x A [DataFrame][DataFrame_class] or [LazyFrame][LazyFrame_class]
-#' @param ... Any arguments passed to [data.frame()].
-#'
+#' Equivalent to `as_polars_df(x, ...)$to_data_frame(...)`.
+#' @param x A object to convert to a [data.frame].
```
etiennebacher, line 233:
```suggestion
#' @param x An object to convert to a [data.frame].
```

R/s3_methods.R
```diff
@@ -229,17 +229,55 @@ dimnames.RPolarsLazyFrame = function(x) list(NULL, names(x))
...
+    projection_pushdown = TRUE,
+    simplify_expression = TRUE,
+    slice_pushdown = TRUE,
+    comm_subplan_elim = TRUE,
+    comm_subexpr_elim = TRUE,
+    streaming = FALSE,
+    no_optimization = FALSE,
+    inherit_optimization = FALSE,
+    collect_in_background = FALSE) {
+  as_polars_df.RPolarsLazyFrame(
```
etiennebacher, line 264:
```suggestion
  as_polars_df(
```
Not necessary since the class of `x` has to be `RPolarsLazyFrame`

R/s3_methods.R
```diff
@@ -229,17 +229,55 @@ dimnames.RPolarsLazyFrame = function(x) list(NULL, names(x))
...
+    simplify_expression = simplify_expression,
+    slice_pushdown = slice_pushdown,
+    comm_subplan_elim = comm_subplan_elim,
+    comm_subexpr_elim = comm_subexpr_elim,
+    streaming = streaming,
+    no_optimization = no_optimization,
+    inherit_optimization = inherit_optimization,
+    collect_in_background = collect_in_background
+  ) |>
+    as.data.frame.RPolarsDataFrame(...)
```
etiennebacher, line 279:
```suggestion
    as.data.frame(...)
```
same

etiennebacher (approved):
Thanks, some minor comments

R/s3_methods.R
etiennebacher: I don't understand your argument. If the arg `x` doesn't have the class `RPolarsLazyFrame` then calling `as.data.frame()` on it won't dispatch it to `as.data.frame.RPolarsLazyFrame()`. So putting `.RPolarsLazyFrame` inside the function is redundant. This is not very important though and I already approved the PR

eitsupi: Indeed, you are right about this place, it may be redundant since there is no possibility that anything other than RPolarsLazyFrame will be entered.

PR #763, opened 2024-02-01
ci: add `macos-14` (arm64 macOS) to GitHub Actions matrixes

eitsupi:
Incorporate the newly available arm64 macOS runner into the CI.

It has been announced that `macos-latest` will be switched from `macos-12` (amd64) to `macos-14` (arm64) in a few months, so it has been changed to explicitly specify `macos-12` or `macos-14`.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -48,14 +48,14 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - { os: macos-latest, r: "release" }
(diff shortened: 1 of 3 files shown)


etiennebacher (approved):
Thanks! The milestone should be 0.14.0 though (or 0.13.1 ?)

eitsupi:
Ah, and for now the milestone is 0.13.
I don't have milestones for each patch version.

PR #766, opened 2024-02-02
empty .lintr take precedence over ~/.lintr

ssh352:
For users who have .lintr in their home dir. r-polars code may show many unwanted lints. This PR is to fix it.

CRAN package rextendr  has .lintr in its project root so I think it is ok to add it in r-polars too.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -31,3 +31,4 @@
 ^Taskfile\.yml$
 ^\.editorconfig$
 ^rustfmt\.toml$
+^\.lintr$


ssh352:
```
+linters: linters_with_defaults() # see vignette("lintr")
+encoding: "UTF-8"
```

not sure if I should turn on default linters, this generates a lot of lints. for example = is used a lot for assignment.

etiennebacher:
Related to https://github.com/pola-rs/r-polars/issues/214

ssh352:
> Related to #214. @eitsupi can you address this one since you've set up most of development tools?

Thanks, we probably should have a project .lintr line just like CRAN package rextendr, to ensure coding styles are consistent. 

this is what is in rextendr/.lintr

```
❯ cat .lintr
linters: linters_with_defaults(
    line_length_linter(120),
    object_name_linter = NULL,
    commented_code_linter = NULL
    )
encoding: "UTF-8"
```

eitsupi:
See #770

PR #768, opened 2024-02-02
docs(website): show DynamicGroupBy in a different section

etiennebacher:
We already had a section for `RollingGroupBy`, I think it makes sense to have one for `RollingGroupBy` too.

Before:

image

After:

image


eitsupi (approved):
Thanks!

PR #770, opened 2024-02-02
refactor: add the lintr config file and some refactoring

eitsupi:
Close #766
Part of #214

Note that this PR does not correct all warnings because there are too many to correct.

Also, because `lint_package` takes several minutes to run, the Taskfile does not run the function.


etiennebacher (approved):
Thanks!

PR #773, opened 2024-02-03
chore: support lib v0.36.2 binaries

eitsupi:


etiennebacher (approved):

eitsupi:
Wow, what a quick review!
(This is the first binary library release support since switching to the arm64 macOS runner, so I'll wait until CI passes before merging just to be safe)

PR #774, opened 2024-02-03
docs(vignette): update the introduction vignette to use the new functions

eitsupi:
I saw a user [on SNS](https://twitter.com/LegoAnand/status/1751190842641957042) who could not figure out how to convert RPolarsDataFrame to data.frame.
I would like to emphasize that it is easy to convert between objects by actively using S3 methods.

- Use `as_polars_*` functions instead of `pl$*` function to convert base objects to polars objects.
- Use `as.data.frame` instead of `$to_data_frame()`.
- Use `<DataFrame>$write_parquet` instead of `arrow::write_parquet`.

### DIFF
--- a/vignettes/polars.Rmd
+++ b/vignettes/polars.Rmd
@@ -471,8 +471,11 @@ on disk. Let's see an example.
 ```{r}
 as_polars_df(airquality)$write_parquet("airquality.parquet")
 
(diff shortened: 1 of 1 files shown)


grantmcdermott (approved):
Lgtm

Issue #756, opened 2024-01-31
Vignette for Apache Arrow interface


eitsupi:
Related to #497, #732

Currently, several public and private functions can exchange Arrow data, but they are not well documented.

PR #785, opened 2024-02-07
Ensure that R CMD check succeeds if `arrow` not available

### LINKED ISSUE #779: The `arrow` package will be removed from CRAN
See apache/arrow#39806.

We may have to update the CI.

etiennebacher:
Close #779 

In tests, I wrapped parameterized cases in `if(requireNamespace("arrow"))` because putting `skip_if_not_installed()` in `with_parameters_test_that()` doesn't work. Basically in `with_parameters_test_that()`, `make_as_polars_df_cases()` is called *before* `skip_if_not_installed()` and it requires `arrow`. 

TODO: `roxygen2` fails because it can't link to `arrow` docs if `arrow` is not installed.


inst/misc/examples/from_arrow.R
eitsupi, line 1: Since this file doesn't actually do anything, why not leave it unedited? I would like to delete it at some point. (Related to #756)

eitsupi (commented):
Thanks for doing this!

inst/misc/examples/from_arrow.R
etiennebacher: Sure, I'll revert that, I forgot it was included

eitsupi (approved):
Thanks. To be sure, this does not close #779. (PR description updated)

PR #786, opened 2024-02-07
fix: simplify print output of LazyFrame

eitsupi:


etiennebacher (commented):
I actually like having this advice for showing the optimized plan. The message could be shortened to simply ```run `$describe_optimized_plan()` to see the optimized plan``` but I think it's useful

eitsupi:
> I actually like having this advice for showing the optimized plan. The message could be shortened to simply `` run `$describe_optimized_plan()` to see the optimized plan `` but I think it's useful

I honestly don't see the value in this being the only one of the many methods shown here, but I changed it to show it.

etiennebacher (approved):
Thanks!

PR #787, opened 2024-02-08
Render vignettes even if `arrow` is absent

### LINKED ISSUE #779: The `arrow` package will be removed from CRAN

etiennebacher:
Related to #779. I tested this locally and the vignette knits correctly if `arrow` is not installed


eitsupi (approved):
Thanks!

PR #790, opened 2024-02-08
Implement the `arr` subnamespace

### LINKED ISSUE #710: Implement the `arr` subnamespace
This was renamed to `list` in earlier versions (following upstream polars), but it looks like it was reintroduced later on. Basically an array column is a list column where each sublist must contain the same number of values.

More details: https://docs.pola.rs/user-guide/expressions/lists/

etiennebacher:
Close #710 

Still WIP:
- [x] add tests
- [x] double-check when strings are parsed as column names and when they're not as there are some inconsistencies with the `list` subnamespace
- [x] cleanup docs (mostly for datatype)


etiennebacher:
@eitsupi R CMD check fails for ubuntu/macos but not for Windows. I'm on windows for now, can you reproduce these panics on either ubuntu or macos?

eitsupi:
It looks like a bug in Rust itself that has not yet been fixed in the latest release. rust-lang/rust#119352

etiennebacher:
Given this leads to panicks (and segfaults if it runs twice), we should probably remove `arr$min` and `arr$max` for now and reimplement them when rust-polars 0.38.0 is released. Do you agree?

eitsupi:
> Do you agree?

Or I think it could be classified as nightly functions instead of doing a deletion.

https://github.com/pola-rs/r-polars/blob/025e3be7dd8acf8e04af207c5f58aff074c899e5/src/rust/Cargo.toml#L11-L17

This actually has nothing to do with `simd`, but could you make sure to check the `simd` feature anyway?
I would like to organize the features later (#300).

etiennebacher:
Not sure I understood correctly, is [097da1a](https://github.com/pola-rs/r-polars/pull/790/commits/097da1a6a3f5549512e3290989cc7092533da115)  what you had in mind?

eitsupi:
It is a little different; it does not need to be handled on the Rust side.
Please wait a bit while I fix it.

etiennebacher:
Thanks for the fix, I'll try to finish this PR during the weekend

PR #796, opened 2024-02-08
Make it easier to create a DataFrame with a single list-column

### LINKED ISSUE #794: Make it easier to create a `DataFrame` with a single list-column
To create a DataFrame with a single list column, we need to wrap the input in `list()` which is annoying:

``` r
library(polars)

pl$DataFrame(list(x = list(1:2, 3:10)))
#> shape: (2, 1)
#> ┌──────────────┐
#> │ x            │
#> │ ---          │
#> │ list[i32]    │
#> ╞══════════════╡
#> │ [1, 2]       │
#> │ [3, 4, … 10] │
#> └──────────────┘

# this should work
pl$DataFrame(x = list(1:2, 3:10))
#>    0: In R: in $DataFrame():
#>          Series length 2 doesn't match the DataFrame height of 8
```
Note that there's no problem to create a DataFrame with >= 2 list-columns:
```r
pl$DataFrame(x = list(1:2, 3:10), y = list(1:2, 3:10))
#> shape: (2, 2)
#> ┌──────────────┬──────────────┐
#> │ x            ┆ y            │
#> │ ---          ┆ ---          │
#> │ list[i32]    ┆ list[i32]    │
#> ╞══════════════╪══════════════╡
#> │ [1, 2]       ┆ [1, 2]       │
#> │ [3, 4, … 10] ┆ [3, 4, … 10] │
#> └──────────────┴──────────────┘
```

etiennebacher:
Close #794

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,7 +40,7 @@
         is changed to `$versions$r_package` (#791).
     -   `$rust_polars`, which indicates the version of the dependent Rust Polars,
         is changed to `$versions$rust_crate` (#791).
--   New behavior when creating a `DataFrame` with a single list-variable. 
+-   New behavior when creating a `DataFrame` with a single list-variable.
     `pl$DataFrame(x = list(1:2, 3:4))` used to create a `DataFrame` with two
     columns named "new_column" and "new_column_1", which was unexpected. It now
     produces a `DataFrame` with a single `list` variable. This also applies to
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,6 +40,10 @@
         is changed to `$versions$r_package` (#791).
     -   `$rust_polars`, which indicates the version of the dependent Rust Polars,
         is changed to `$versions$rust_crate` (#791).
+-   New behavior when creating a `DataFrame` with a single list-variable. 
+    `pl$DataFrame(x = list(1:2, 3:4))` used to create a `DataFrame` with two
+    columns named "new_column" and "new_column_1", which was unexpected. It now
+    produces a `DataFrame` with a single `list` variable (#794).
 
 ### Deprecations
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,7 +43,8 @@
 -   New behavior when creating a `DataFrame` with a single list-variable. 
     `pl$DataFrame(x = list(1:2, 3:4))` used to create a `DataFrame` with two
     columns named "new_column" and "new_column_1", which was unexpected. It now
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -40,6 +40,10 @@
         is changed to `$versions$r_package` (#791).
     -   `$rust_polars`, which indicates the version of the dependent Rust Polars,
         is changed to `$versions$rust_crate` (#791).
+-   New behavior when creating a `DataFrame` with a single list-variable. 
+    `pl$DataFrame(x = list(1:2, 3:4))` used to create a `DataFrame` with two
+    columns named "new_column" and "new_column_1", which was unexpected. It now
+    produces a `DataFrame` with a single `list` variable (#794).
```
eitsupi, line 46: I suppose it affects to `$with_columns` etc. too.
Could you check them?

eitsupi (changes requested):
Good point, thanks!

NEWS.md
etiennebacher: Thanks I added some tests

eitsupi: Thanks, could you please mention in the NEWS the other functions that are affected?

eitsupi (approved):

NEWS.md
```diff
@@ -40,6 +40,10 @@
         is changed to `$versions$r_package` (#791).
     -   `$rust_polars`, which indicates the version of the dependent Rust Polars,
         is changed to `$versions$rust_crate` (#791).
+-   New behavior when creating a `DataFrame` with a single list-variable.
```
eitsupi, line 43:
```suggestion
-   New behavior when creating a `DataFrame` with a single list-variable.
```

etiennebacher: Can we avoid this type of changes? This is something that `task format-r` doesn't fix

etiennebacher:
Note that behavior for creating list-variable in `$agg()` was already broken so I'll open a separate issue for that

NEWS.md
eitsupi: An auto-formatter for Markdown must be installed.
VSCode can easily handle this with extensions, but I am not familiar with the tools available for other editors. Perhaps Pretitter is well known?

NEWS.md
etiennebacher: done

NEWS.md
etiennebacher: I don't know how that works with RStudio

PR #798, opened 2024-02-10
docs: clarify argument `maintain_order`

etiennebacher:
* For `group_by`, clarify that `maintain_order` only influences the order of groups, not the order of rows in each group
* Use a different definition for each type of function. `maintain_order` doesn't mean the same thing for `sink_*()` and for `unique()` for example
* Clarify that `polars_options()$maintain_order` only affects the value of `maintain_order` in `group_by` functions. We can discuss if this should be extended to all `maintain_order` args, but that's how it has been in the code so far.

Easier to review by looking at the content of the Rd files.

### DIFF
--- a/R/lazyframe__lazy.R
+++ b/R/lazyframe__lazy.R
@@ -1043,15 +1043,15 @@ LazyFrame_unique = function(
 
 #' Group a LazyFrame
 #' @description This doesn't modify the data but only stores information about
-#' the group structure. This structure can then be used by several functions
-#' (`$agg()`, `$filter()`, etc.).
+#'   the group structure. This structure can then be used by several functions
+#'   (`$agg()`, `$filter()`, etc.).
 #' @keywords LazyFrame
 #' @param ... Any Expr(s) or string(s) naming a column.
-#' @param maintain_order Keep the same group order as the original `LazyFrame`.
-#' Within each group, the order of rows is always preserved, regardless of this
-#' argument. Setting this to `TRUE` makes it more expensive to compute and
-#' blocks the possibility to run on the streaming engine. The default value can
-#' be changed with `options(polars.maintain_order = TRUE)`.
+#' @param maintain_order Keep the same group order as the original data. Within
+#'   each group, the order of rows is always preserved, regardless of this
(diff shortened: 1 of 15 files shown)


R/lazyframe__lazy.R
```diff
@@ -1038,14 +1043,15 @@ LazyFrame_unique = function(
 
 #' Group a LazyFrame
 #' @description This doesn't modify the data but only stores information about
-#' the group structure. This structure can then be used by several functions
-#' (`$agg()`, `$filter()`, etc.).
+#'   the group structure. This structure can then be used by several functions
+#'   (`$agg()`, `$filter()`, etc.).
```
eitsupi, line 1047: Why this change?

R/lazyframe__lazy.R
```diff
@@ -1038,14 +1043,15 @@ LazyFrame_unique = function(
...
 #' @param ... Any Expr(s) or string(s) naming a column.
-#' @param maintain_order Keep the same order as the original `LazyFrame`. Setting
-#'  this to `TRUE` makes it more expensive to compute and blocks the possibility
-#'  to run on the streaming engine. The default value can be changed with
-#' `options(polars.maintain_order = TRUE)`.
+#' @param maintain_order Keep the same group order as in the original data.
+#'   Within each group, the order of rows is always preserved, regardless of
+#'   this argument. Setting this to `TRUE` makes it more expensive to compute
+#'   and blocks the possibility to run on the streaming engine. The default
+#'   value can be changed with `options(polars.maintain_order = TRUE)`.
```
eitsupi, line 1054:
```suggestion
#' Within each group, the order of rows is always preserved, regardless of
#' this argument. Setting this to `TRUE` makes it more expensive to compute
#' and blocks the possibility to run on the streaming engine. The default
#' value can be changed with `options(polars.maintain_order = TRUE)`.
```

eitsupi (approved):
Thanks!

R/lazyframe__lazy.R
etiennebacher: That's the autoformatting that wraps lines so that they don't go over 80 chars (ctrl+shift+/). Apparently it also adds indents. I can remove that

PR #799, opened 2024-02-10
dev: add aliases for Task

etiennebacher:
I just added a few aliases, not for all tasks (those are the ones I use the most). @eitsupi Do you want to add more?

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -95,6 +95,7 @@ tasks:
       - Rscript dev/generate-lib-sums.R
 
   build-all:
+    aliases: [ba]
     desc: Build the R package, generate documents, run all tests, and update files.
     deps:
       - build-lib-sums
@@ -104,6 +105,7 @@ tasks:
       - task: build-readme
 
   test-all:
+    aliases: [ta]
     desc: Run all tests.
     cmds:
       - task: test-source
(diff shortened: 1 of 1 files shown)


eitsupi:
In fact, I have never wanted such a feature as I use shell completion.
https://taskfile.dev/installation/#setup-completions

Aliases are basically very confusing and difficult to understand what the name indicates.......

etiennebacher:
I use the default Windows terminal in RStudio and I don't have such completion. You don't have to use the aliases and I don't think they'll appear in the suggested completion (but I'm not sure)

eitsupi:
> I use the default Windows terminal in RStudio and I don't have such completion.

It must be set up as described in the documentation.

etiennebacher:
I followed the steps for PowerShell and got this error: `File <path>\Microsoft.PowerShell_profile.ps1 cannot be loaded because running scripts is disabled on this system.` (I don't have admin rights).

Maybe you can just ignore the aliases if you don't want to use them?

eitsupi:
If you are using PowerShell, will suggestions using history with PSReadLine not work?

My point is that such aliases for personal convenience should be kept as personal settings, not added here.
For example, I know many people generally set aliases for Git subcommands, but that's something the person should manage in something like a dotfile, right?

etiennebacher:
Yeah, I could keep that locally and gitignore the Task file, but then that would be problematic if I have other changes to add to the Taskfile. That shouldn't happen too often so I'm gonna keep those aliases locally for now

eitsupi:
I also cannot be certain whether the error you saw in the first place was really due to the configuration or simply a misstep in the procedure.

PR #800, opened 2024-02-11
refactor!: feature flags reorganizization

### LINKED ISSUE #300: Revise the contents of Cargo.toml

eitsupi:
Close #300

Match with py-polars.

- `polars/nightly` includes `polars/simd`, and py-polars does not specify `polars/simd` directly but includes `polras/nightly` in the `nightly` feature.
- The `default` feature includes the `sql` feature (because `sql` no longer requires nightly toolchain).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,8 @@
 
 ### Other breaking changes
 
+-   The `simd` feature of the Rust library is removed in favor of
+    the new `nightly` feature (#800).
 -   The following functions were deprecated in 0.13.0 and are now removed (#783):
     -   `$list$lengths()` -> `$list$len()`
     -   `pl$from_arrow()` -> `as_polars_df()` or `as_polars_series()`
@@ -50,6 +52,10 @@
 
 -   `pl$threadpool_size()` is deprecated and will be removed in 0.15.0. Use
     `pl$thread_pool_size()` instead (#784).
+
+### Other improvements
+
+-   The `sql` feature is included in the default feature (#800).
 
 ## Polars R Package 0.13.1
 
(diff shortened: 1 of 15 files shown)


NEWS.md
```diff
@@ -24,6 +24,8 @@
 
 ### Other breaking changes
 
+-   The `simd` feature of the Rust library is removed in favor of
```
etiennebacher, line 27: can you add details on whether it changes something for the users? Should they change anything?

NEWS.md
```diff
@@ -51,6 +53,10 @@
 -   `pl$threadpool_size()` is deprecated and will be removed in 0.15.0. Use
     `pl$thread_pool_size()` instead (#784).
 
+### Other improvements
+
+-   The `sql` feature is included in the default feature (#800).
```
etiennebacher, line 58: same

NEWS.md
eitsupi: Since we can update the NEWS file later, could you approve it for now?

It contains changes to the manifest file, and given the caching and the number of jobs triggered, I don't want to prolong this PR with only a documentation update.

etiennebacher (approved):

PR #801, opened 2024-02-11
Bump news for `arr` subnamespace

etiennebacher:
Followup of #790, thanks for the reminder @eitsupi

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -50,6 +50,12 @@
 
 -   `pl$threadpool_size()` is deprecated and will be removed in 0.15.0. Use
     `pl$thread_pool_size()` instead (#784).
+    
+### New features
+
+-   Implementation of the subnamespace `arr` for `array`-type columns. An `array`
+    column is similar to a `list` column, but is stricter as each sub-array
+    must have the same number of elements (#790). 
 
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -52,6 +52,12 @@
 
 -   `pl$threadpool_size()` is deprecated and will be removed in 0.15.0. Use
     `pl$thread_pool_size()` instead (#784).
+    
+### New features
+
+-   Implementation of the subnamespace `arr` for `array`-type columns. An `array`
+    column is similar to a `list` column, but is stricter as each sub-array
+    must have the same number of elements (#790).
```
eitsupi, line 60:
```suggestion
-   Implementation of the subnamespace `arr` for `array`-type columns. An `array`
    column is similar to a `list` column, but is stricter as each sub-array
    must have the same number of elements (#790).
```

- Please remove white space.
- I don't think the expression "subnamespace" is clear. Could you express that it is a subnamespace of Expr?

eitsupi (changes requested):

NEWS.md
etiennebacher: done

eitsupi (approved):

Issue #772, opened 2024-02-03
Documents for active bindings are wrong


eitsupi:
It says like `<DataFrame>$shape()` for example, but that should be `<DataFrame>$shape`, not `shape()`.

https://github.com/pola-rs/r-polars/blob/7f1ee4adcd5e0456d7b9abc9c08ab8e938aa0d12/man/DataFrame_shape.Rd#L6-L8

Perhaps it should appear on the Class page rather than on individual pages with respect to attributes.

eitsupi:
Example of documentation: #809

PR #809, opened 2024-02-13
Implement `<DataFrame>$flags` and improve `<Series>$flags`

etiennebacher:
One issue here is that I'm actually not sure what the `"FAST_EXPLODE"` actually means (which is why I didn't explain in the docs)


NEWS.md
```diff
@@ -2,6 +2,12 @@
 
 ## Polars R Package (development version)
 
+### New features
+
+-   New method `flags` for `DataFrame` to show the flags used internally for 
+    each column. The output of `flags` for `Series` was also improved and now
+    contains `FAST_EXPLODE` for `Series` of type `list` and `array`.
```
eitsupi, line 9: I think they are attributes, not methods.

R/dataframe__frame.R
```diff
@@ -2059,3 +2059,29 @@ DataFrame_group_by_dynamic = function(
...
+#'
+#' `SORTED_ASC` is set to `TRUE` when we sort a column in increasing order, so
+#' that we can use this information later on to avoid re-sorting it. `SORTED_DESC`
+#' is similar but applies to sort in decreasing order.
+#'
+#' @return A nested list with column names at the top level and column flags
+#' in each sublist.
+#'
+#' @examples
+#' pl$DataFrame(a = c(2, 1), b = c(3, 4), c = list(c(1, 2), 4))$sort("a")$flags
```
eitsupi, line 2080: I think these documents should not be here.
See #772

eitsupi (changes requested):

R/dataframe__frame.R
etiennebacher: But then should we group man pages for `Series_flags` and `DataFrame_flags`?

eitsupi: I'm sorry but I don't understand what you are trying to say.
In my opinion attributes should be mentioned on the class pages without having a separate documentation.

etiennebacher: I misunderstood what you said but I think I get it, I'll finish this PR later

eitsupi:
Sorry, it is the Python world that calls such things "attributes", and here (in R) "active bindings" was appropriate.

eitsupi (approved):

R/dataframe__frame.R
```diff
@@ -1,34 +1,52 @@
...
+#' varies depending on the column type: columns of type `array` and `list`
+#' have the flags `SORTED_ASC`, `SORTED_DESC`, and `FAST_EXPLODE`, while other
+#' column types only have the former two.
+#'
+#' - `SORTED_ASC` is set to `TRUE` when we sort a column in increasing order, so
+#'   that we can use this information later on to avoid re-sorting it.
+#' - `SORTED_DESC` is similar but applies to sort in decreasing order.
+#'
+#' @details Check out the source code in
+#' [R/dataframe_frame.R](https://github.com/pola-rs/r-polars/blob/main/R/dataframe__frame.R)
```
eitsupi, line 38: I don't think such things should be placed here (they should be in the developer's documentation), but since they are irrelevant to this PR, I'll leave them alone.

R/dataframe__frame.R
```diff
@@ -74,6 +94,15 @@
 NULL
 
 
+DataFrame_flags = method_as_property(function() {
+  out = lapply(self$columns, \(x) {
+    self[x]$to_series()$flags
```
eitsupi, line 99: `to_series()` is needed here?
How about `self[, x]`?

etiennebacher: >  How about self[, x]?

Yes that's simpler, good catch

Issue #255, opened 2023-06-23
String methods doesn't work from Series


ddotta:
I noticed that the string methods didn't work from a `Series`.

Whereas it works fine from a column in a `DataFrame`.

For example with `to_uppercase()` :

```r
> pl$Series(name = "myletters", c("a", "b", "c"))$str$to_uppercase()
Error: syntax error: str is not a method/attribute of the class Series 
 when calling method:
 pl$Series(name = "myletters", c("a", "b", "c"))$str
```

```r
> pl$DataFrame(
+   col1 = c("a","b","c")
+ )$with_columns(
+   pl$col("col1")$str$to_uppercase()
+ )
shape: (3, 1)
┌──────┐
│ col1 │
│ ---  │
│ str  │
╞══════╡
│ A    │
│ B    │
│ C    │
└──────┘
```

eitsupi:
If someone wants to take this, please refer to #810 and work like that.

eitsupi:
https://github.com/pola-rs/r-polars/pull/810#issuecomment-1950958347

PR #810, opened 2024-02-16
Create the string namespace for Series

etiennebacher:
Related to #255.

I don't have the time to fully implement all namespaces for Series (it is also quite tedious work), but hopefully this opens the way to some contributions from other people as it shows how to implement those namespaces and the methods in each of them.


eitsupi (approved):
Makes sense, thanks!

eitsupi:
I noticed that Series already has a list subnamespace, couldn't we use the same implementation method as list?

eitsupi:
I am now convinced that this method of implementation was wrong...

PR #812, opened 2024-02-17
ci: make dependabot to check local action and devcontainer config file

eitsupi:
I did not know that additional configuration was required to have GitHub Actions check local actions.

> Dependabot will ignore actions or reusable workflows referenced locally (for example, `./.github/actions/foo.yml)`.

https://docs.github.com/ja/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#github-actions

Also add settings for updating devcontainer.json.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,9 +7,21 @@ updates:
     commit-message:
       prefix: "ci: "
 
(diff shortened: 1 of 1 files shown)


etiennebacher (approved):
Thanks

Issue #676, opened 2024-01-06
Scan R data.frame


eitsupi:
For now, in `as_polars_lf`, we convert data.frame to RPolarsDataFrame (high cost) and then scan that.
Can we scan data.frame directly?
I think duckdb does that.

etiennebacher:
That could also improve `pl$LazyFrame()`, which creates a DataFrame and then calls `$lazy()`

eitsupi:
It seems that this requires something like this: https://github.com/pola-rs/polars/blob/7749371169d62ba0a170520baa10834073c5d51f/crates/polars-lazy/src/frame/python.rs#L6-L18

eitsupi:
AnonymousScan (pola-rs/polars#3561) seems exist for usages like this.

PR #817, opened 2024-02-17
refactor: rewrite `as_polars_df.data.frame`

### LINKED ISSUE #816: Rewrite `as_polars_df.data.frame`
Related to #676

eitsupi:
Close #816
It used to call `pl$DataFrame` internally, but has been completely rewritten to not call it.

Speeds seem to be comparable, but memory usage may have been reduced.
(The `bench` package does not measure memory usage on the Rust side, but given that the output RPolarsDataFrame is the same, memory usage on the Rust side should not be much different.)

``` r
big_df = do.call(rbind, lapply(1:10, \(x) nycflights13::flights))

library(polars)
library(nanoarrow)

bench::mark(
  pl_DataFrame = {
    pl$DataFrame(big_df)
  },
  as_polars_df = {
    as_polars_df(big_df)
  },
  from_nanoarrow = {
    as_polars_df(as_nanoarrow_array_stream(big_df))
  },
  check = FALSE,
  min_iterations = 10L
)
#> # A tibble: 3 × 6
#>   expression          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 pl_DataFrame   826.41ms 839.49ms     1.16     1.91MB    0.289
#> 2 as_polars_df   738.85ms 864.31ms     1.16   204.05KB    0.129
#> 3 from_nanoarrow    1.15s    1.27s     0.790   77.48MB    0.339
```

<sup>Created on 2024-02-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
     each column. The output of `$flags` for `Series` was also improved and now
     contains `FAST_EXPLODE` for `Series` of type `list` and `array` (#809).
 -   Add string methods `to_lowercase()` and `to_uppercase()` for `Series` (#810).
+-   `as_polars_df()` for `data.frame` is completely rewritten and new arguments
+    `schema` and `schema_overrides` are added (#817).
 
 ## Polars R Package 0.14.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
     each column. The output of `$flags` for `Series` was also improved and now
     contains `FAST_EXPLODE` for `Series` of type `list` and `array` (#809).
 -   Add string methods `to_lowercase()` and `to_uppercase()` for `Series` (#810).
--   `as_polars_df()` for `data.frame` is completely rewritten and new arguments
+-   `as_polars_df()` for `data.frame` is more memory-efficient and new arguments
     `schema` and `schema_overrides` are added (#817).
 
 ## Polars R Package 0.14.0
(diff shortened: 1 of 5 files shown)


eitsupi:
I will update the test and documentation later.

@etiennebacher @sorhawell If you are interested, check the performance in the benchmarks above.

etiennebacher:
I'll review in more details later, but I'm a bit confused about the objective of this PR. If the speed isn't affected and we don't know if the memory used decreases, then why rewrite this? Is this an intermediate step to fix #676?

eitsupi:
Given the simplicity of the process, it is not surprising that memory usage is reduced.

NEWS.md
```diff
@@ -8,6 +8,8 @@
     each column. The output of `$flags` for `Series` was also improved and now
     contains `FAST_EXPLODE` for `Series` of type `list` and `array` (#809).
 -   Add string methods `to_lowercase()` and `to_uppercase()` for `Series` (#810).
+-   `as_polars_df()` for `data.frame` is completely rewritten and new arguments
```
etiennebacher, line 11:
```suggestion
-   `as_polars_df()` for `data.frame` is more memory-efficient and new arguments
```

etiennebacher (approved):

PR #818, opened 2024-02-17
test: `pl$DataFrame` can include vctrs_rcrd class vector

### LINKED ISSUE #575: `Can't assign names to a <vctrs_rcrd>` error in `pl$DataFrame()`
Create a new class based on <https://vctrs.r-lib.org/articles/pillar.html>.

```r
library(vctrs)

latlon <- function(lat, lon) {
  new_rcrd(list(lat = lat, lon = lon), class = "earth_latlon")
}

format.earth_latlon <- function(x, ..., formatter = deg_min) {
  x_valid <- which(!is.na(x))

  lat <- field(x, "lat")[x_valid]
  lon <- field(x, "lon")[x_valid]

  ret <- rep(NA_character_, vec_size(x))
  ret[x_valid] <- paste0(formatter(lat, "lat"), " ", formatter(lon, "lon"))
  # It's important to keep NA in the vector!
  ret
}

deg_min <- function(x, direction) {
  pm <- if (direction == "lat") c("N", "S") else c("E", "W")

  sign <- sign(x)
  x <- abs(x)
  deg <- trunc(x)
  x <- x - deg
  min <- round(x * 60)

  # Ensure the columns are always the same width so they line up nicely
  ret <- sprintf("%d°%.2d'%s", deg, min, ifelse(sign >= 0, pm[[1]], pm[[2]]))
  format(ret, justify = "right")
}
```

Then, create a polars DataFrame.

```r
polars::pl$DataFrame(x = latlon(c(32.71, 2.95), c(-117.17, 1.67)))
```

This error is occur.

```
   0: In R: in $DataFrame():
   0: During function call [polars::pl$DataFrame(x = latlon(c(32.71, 2.95), c(-117.17, 1.67)))]
   1: Can't assign names to a <vctrs_rcrd>.
```

eitsupi:
Close #575

### DIFF
--- a/tests/testthat/test-as_polars.R
+++ b/tests/testthat/test-as_polars.R
@@ -179,7 +179,7 @@ test_that("tests for vctrs_rcrd", {
 
   expect_identical(length(as_polars_series(vec)), 2L)
 
-  pl$DataFrame(foo = vec)
+  expect_snapshot(pl$DataFrame(foo = vec)$dtypes, cran = TRUE)
 
   expect_identical(
     dim(as_polars_df(tibble::tibble(foo = vec))),
--- a/tests/testthat/test-as_polars.R
+++ b/tests/testthat/test-as_polars.R
@@ -119,7 +119,6 @@ if (requireNamespace("arrow", quietly = TRUE) && requireNamespace("nanoarrow", q
   patrick::with_parameters_test_that(
     "as_polars_series S3 methods",
     {
-
       pl_series = as_polars_series(x)
       expect_s3_class(pl_series, "RPolarsSeries")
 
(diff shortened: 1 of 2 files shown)


tests/testthat/test-as_polars.R
```diff
@@ -180,9 +179,7 @@ test_that("tests for vctrs_rcrd", {
 
   expect_identical(length(as_polars_series(vec)), 2L)
 
-  # TODO: this should work
-  # https://github.com/pola-rs/r-polars/issues/575
-  # pl$DataFrame(foo = vec)
+  pl$DataFrame(foo = vec)
```
etiennebacher, line 182: I think you forgot to wrap this in an `expect_` statement

eitsupi: Indeed.
I couldn't think of a good test to confirm this, so I decided on a snapshot test.

etiennebacher (approved):
thanks!

PR #819, opened 2024-02-18
refactor: rewrite active bindings and sub-namespaces

### LINKED ISSUE #255: String methods doesn't work from Series

### LINKED ISSUE #772: Documents for active bindings are wrong

### LINKED ISSUE #813: `?RPolarsDataFrame` does not work
We should `?DataFrame_class` now, but I think `?RPolarsDataFrame` must work.

eitsupi:
Close #255 (Rewrite of #810)
Close #772
Close #813

Rewrite the documents for active bindings (and sub-namespaces) to include them in the docs for each class.


eitsupi:
@etiennebacher Could you take a look at this?

R/dataframe__frame.R
```diff
@@ -55,20 +81,20 @@
 #' # see all private methods (not intended for regular use)
 #' ls(.pr$DataFrame)
 #'
-#'
 #' # make an object
-#' df = pl$DataFrame(iris)
+#' df = as_polars_df(iris)
 #'
-#' # use a public method/property
+#' # call a active binding
```
etiennebacher, line 87:
```suggestion
#' # call an active binding
```

R/expr__expr.R
```diff
@@ -1,27 +1,88 @@
...
-#'  * String
-#'  * Struct
-#'  * Temporal
+#' Expressions are all the functions and methods that are applicable to
+#' a Polars [DataFrame][DataFrame_class] or [LazyFrame][LazyFrame_class] object.
+#' Some methods are under the sub-namespaces.
 #'
+#' @section Sub-namespaces:
+#'
+#' ## arr
```
etiennebacher, line 9: Making a subsection with `##` for each category is not very useful, you could make a list of bullet points instead. Same thing for the active bindings `columns`, `shape`, etc.

R/series__series.R
```diff
@@ -40,6 +45,33 @@
...
+#'
+#' @section Sub-namespaces:
+#'
+#' Some functions are stored under active bindings, so it works like a sub-namespaces.
+#'
+#' ## expr
+#'
+#' `$expr` is a shorthand of writing something like
+#' `pl$select(s)$select(pl$col(<name>)$<Expr>)$to_series(0)`.
+#' This subnamespace is experimental.
```
etiennebacher, line 65: Just to be sure, this is a workaround because `pl$Series(c(1:3))$mean()` doesn't work for example, right? If so it would be useful to add a note that explains the diff with the python syntax (that just needs `.mean()`).

R/series__series.R
```diff
@@ -61,34 +93,104 @@
 #' s_copy = s
 #' .pr$Series$append_mut(s, pl$Series(5:1))
 #' identical(s_copy$to_r(), s$to_r()) # s_copy was modified when s was modified
+#'
+#' # call functions via sub-namespaces
+#' pl$Series(c(1:3))$expr$add(1)
+#'
+#' pl$Series(list(3:1, 1:2, NULL))$list$first()
+#'
+#' pl$Series(c(1, NA, 2))$str$concat("-")
```
etiennebacher, line 102: This should return a Series, not a DataFrame

R/series__series.R
```diff
@@ -61,34 +93,104 @@
...
+  )
 })
 
 
+Series_list = method_as_active_binding(
+  \() {
+    df = pl$select(self)
+    arr = expr_list_make_sub_ns(pl$col(self$name))
+    lapply(arr, \(f) {
+      \(...) df$select(f(...))
```
etiennebacher, line 177: I think adding `$to_series()` here would be enough to address my previous comment

tests/testthat/test-series_string.R
```diff
@@ -1,13 +0,0 @@
-test_that("Series to lower case", {
-  expect_identical(
-    pl$Series(c("A", "B", NA, "D"))$str$to_lowercase()$to_r(),
-    c("a", "b", NA, "d")
-  )
-})
-
```
etiennebacher, line 7: I don't think those should be removed. We probably don't need to test all methods for Series but at least a few of them to ensure that the dispatch from Expr to Series works fine

etiennebacher (changes requested):
Thanks, those are great docs improvements!

I have a couple of comments on the subnamespaces for Series

R/series__series.R
eitsupi: `Series_mean` exists here.
https://github.com/pola-rs/r-polars/blob/c2ef53b16b73bcf25d4af15a44f6fbfb0373374c/R/series__series.R#L670-L672

R/series__series.R
eitsupi: Thanks for catching this!

R/expr__expr.R
eitsupi: Yes, you are right here, but there is already a long explanation of active bindings.
https://github.com/pola-rs/r-polars/blob/c2ef53b16b73bcf25d4af15a44f6fbfb0373374c/R/dataframe__frame.R#L20-L35

Therefore, I thought it would be better to align the format of sub-namespaces with that of active bindings.

etiennebacher: I get your point, it's just that this takes unnecessary space since it's one almost identical sentence for each category:

image

compared to:

image

But that's not a big deal, just a matter of preference I guess. Leave it or change it ;)

R/series__series.R
```diff
@@ -40,6 +45,34 @@
...
+#'
+#' Some functions are stored under active bindings, so it works like a sub-namespaces.
+#'
+#' ## expr
+#'
+#' `$expr` works as a workaround for applying arbitrary [Expr][Expr_class] methods
+#' to the Series class object, which means that this is a shortcut
+#' works like `pl$select(s)$select(pl$col(s$name)$<Expr method>)$to_series(0)`.
+#' This subnamespace is experimental.
+#'
```
etiennebacher, line 67: Sorry if I'm missing something obvious but I don't understand the point of this subnamespace. In the examples, you call `pl$Series(c(1:3))$expr$add(1)`, but that is simply the same as calling `pl$Series(c(1:3))$add(1)`. Why do we need this `$expr`?

etiennebacher (changes requested):

R/series__series.R
eitsupi: I don't think `$expr` was added by me. Why it was added I do not know.

eitsupi: This PR is just a refactoring, and I don't think we should discuss its significance now.
We can delete this later.

etiennebacher: Ah ok, I didn't notice that was already there

etiennebacher (approved):
That's a nice refactoring, thanks @eitsupi

R/series__series.R
eitsupi: Yes, no one even noticed that $list methods were returning a DataFrame until now, so I'm guessing that very few users used such a method in Series

PR #827, opened 2024-02-20
Remove the experimental `$expr` subnamespace for `Series`

etiennebacher:
Follow-up of #819

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,13 @@
 
 ### Bug fixes
 
--   `<Series>$list` sub namespace methods returns `Series` class object correctly (#819).
+-   `<Series>$list` sub namespace methods returns `Series` class object correctly 
+    (#819).
+
+### Other
+
+-   The experimental subnamespace `$expr` on `Series` is removed as it was 
+    barely documented and doesn't add anything (#827).
 
 ## Polars R Package 0.14.0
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -17,7 +17,14 @@
...
 
--   `<Series>$list` sub namespace methods returns `Series` class object correctly (#819).
+-   `<Series>$list` sub namespace methods returns `Series` class object correctly 
+    (#819).
+
+### Other
+
+-   The experimental subnamespace `$expr` on `Series` is removed as it was 
+    barely documented and doesn't have additional value compared to applying
+    methods directly on `Series` (#827).
```
eitsupi, line 27: Sorry I have not answered your question adequately.
In fact, under `$expr`, all the methods of the `Expr` class are stored, so there are functions that cannot be used without it.

In Python, the methods of the Series class are almost the same as what we do with `$expr`, since the methods of Expr are dispatched as follows.
https://github.com/pola-rs/polars/blob/24c0f474108e6dcf095068823ba0925b31c4e3b4/py-polars/polars/series/utils.py#L28-L36

So we should aim to be able to do the same thing without `$expr`, and until that is achieved `$expr` may be worthwhile.
But as I commented elsewhere, I think very few users are using this at the moment.

etiennebacher: Ah ok I see, the example I mentioned in another PR only worked because `$add()` was already implemented for `Series`. I'll think more about how we can automatically dispatch the other methods without using `$expr`. Thanks for the explanation

eitsupi: > So we should aim to be able to do the same thing without `$expr`

Done in #831

PR #830, opened 2024-02-22
fix: Update flake to fix failing build

detroyejr:
I've been meaning to try out polars for a while and was pleased to see a flake available. Things didn't build on first run, so I made a few updates to get this working again.

1. `nix flake update`.
2. `cargo update -p ahash` - This was the main reason for the build failure. Some recent changes to ahash seems to have broken this and several other libraries. I left all other dependencies alone.
3. Add a patchPhase with `patchShebangs ./configure`.

One question: in the current main branch, polars is not included as a package dependency in the custom `renv`. A downstream version of this repo *did* include it but it was removed at some point. I'm wondering if that was omitted on purpose or if that should be added back like so:

```
          packages = with pkgs.rPackages; [ devtools languageserver renv rextendr rpolars ];
```

Without it, I cannot run `nix develop github:pola-rs/r-polars` and get a working version of R with polars available.

Tested with `nix develop github:detroyejr/r-polars`. Once everything builds `R -e 'packageVersion("polars")'` returns 0.14.0.9000. The example from the README also works.

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -26,6 +26,14 @@ runs:
           version: 3.x
           repo-token: ${{ inputs.token }}
 
+      - name: Install R
+        # Workaround for https://github.com/actions/runner-images/issues/9394
+        if: runner.os == 'macOS'
+        uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+          Ncpus: 2
+
       - name: Set up Rust nightly toolchain
         if: inputs.rust-nightly == 'true' || env.LIBR_POLARS_FEATURES == 'full_features'
         shell: bash
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -18,6 +18,7 @@ on:
     branches:
       - main
     paths:
+      - .github/actions/setup/action.yaml
       - .github/workflows/check.yaml
       - .Rbuildignore
       - R/**
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -24,6 +24,7 @@ on:
     branches:
       - main
     paths:
+      - .github/actions/setup/action.yaml
       - .github/workflows/docs.yaml
       - altdoc/**
       - man/**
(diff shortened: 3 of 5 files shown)


Sicheng-Pan:
Thanks a lot for the patch! The reason I did not add `rpolars` as a dependency in the devshell is because (from my understanding) Nix does not cache the intermediate build cache for Rust, so if made changes to the code I need to re-compile from scratch, which is really painful for me when I develop. So for simplicity I choose to use `devtools` in R to compile the code in the devshell, and it will automatically load the library when it finishes. `devtools` also helps to build the docs and perform sanity checks, and I'm unsure if it can be replaced by some Nix script. If `devtools` are no longer needed then I believe that the "proper" approach could be using something like [`crane`](https://crane.dev/) to speed up the compilation a bit, and in that case we could put `rpolars` as a dependency in the devshell.

Sicheng-Pan:
@detroyejr Also it's been a while since I last updated the flake lock, so feel free to update it there is no conflict!

detroyejr:
That makes sense. I've removed that from the renv.

The flake lock is updated and I didn't see any issues.

eitsupi:
The CI failure of macos-14 is came from the image update actions/runner-images#9394

Sicheng-Pan:
@eitsupi Just to confirm the `renv` R package is no longer a dependency for the project right? If that is the case it could also be removed from the flake.

eitsupi:
> Just to confirm the `renv` R package is no longer a dependency for the project right?

Yes, see #586

detroyejr:
Got it. Removed the renv package from the flake as well.

Sicheng-Pan (commented):
Please help me to remove `renv` here. Thanks in advance!

eitsupi:
Could you please pull in the main branch?
#832 should fix the CI failure.

Sicheng-Pan (approved):
Thanks for the help!

eitsupi:
Thanks, we are freezing the Rust side in preparation for the 0.14.1 release and will merge it after the 0.14.1 release.

PR #831, opened 2024-02-22
feat: add all methods of RPolarsExpr to RPolarsSeries

eitsupi:
Instead of the experimental `<Series>$expr$<method>`, we can call methods that exist only in Expr with `<Series>$<method>`.


eitsupi:
@etiennebacher I would like to merge this for a new release. Could you help me merge it?

etiennebacher:
I can review in a couple of hours

etiennebacher (changes requested):
That looks good to me, thanks!

Now that the method dispatch is made automatically, can you remove the old implementation of the methods for `Series`? For example it's no longer necessary to implement `Series_add` ourselves. Note however that some methods cannot be removed because they should return a scalar and not a `Series` (e.g `$sum()`, `$mean()`...)

eitsupi:
> Now that the method dispatch is made automatically, can you remove the old implementation of the methods for `Series`?

Of course it is possible, but most of the existing methods look like thin wrappers for Rust Polars, so I think it might as well be left as they are.

etiennebacher (approved):
I get that some methods must be implemented in Rust because either they don't have equivalent in Expr or the Expr method would return a Series instead of a scalar. But other Series methods could simply be removed from the Rust and R sides and nothing would change, e.g `$floor()`, `$ceiling()`.

I don't want this to be a blocker for this PR, but I think we should get rid of them to reduce the codebase.

etiennebacher:
Apart from that, thank you for this PR, it's super useful to dispatch methods like this

eitsupi:
Sorry, the commit comment for the merge commit was broken due to browser behavior.

PR #832, opened 2024-02-23
ci: ensure install R on macOS runner

eitsupi:
Workflow started failing because R was not installed on the new macOS runner (actions/runner-images#9394).
I don't know if this is a temporary or permanent problem, but this PR will correct for the possibility of a temporary problem.


eitsupi:
I think this should be merged immediately.

Sicheng-Pan (approved):
Looks good to me

eitsupi:
Thanks for your review!

PR #834, opened 2024-02-23
ci: add mega-linter for linting

eitsupi:


etiennebacher (commented):
Why do we need this? From what I can see it will `lintr` and `cargo clippy` for linting so what is the advantage of using this compared to our existing formatting tools in the taskfile? Does this push changes automatically?

etiennebacher:
Looks like a very big tool for not many linting usages (the workflow takes about 2min to lint yaml and markdown files and we don't have many of them). If I remember correctly, it was Prettier that fixed the Markdown spaces for you locally. Maybe we could add a `format-other` task that calls the Prettier CLI locally?

I see they also have a [Github action](https://github.com/marketplace/actions/prettier-action), is it worth trying it as well?

eitsupi:
What I really want to run is editorconfg, but I can't find the proper action in GitHub Actions.

Yes, the execution time is relatively long, but it is far better than no check.

eitsupi:
@etiennebacher I want to merge this so can you please take a look?

etiennebacher:
I really think this is overkill for spaces in Markdown files. I merge it so we can move on but I'll open an issue to see if Prettier has a github action that is less massive

etiennebacher (approved):

PR #836, opened 2024-02-23
feat: make `$otherwise` optional in when-then-otherwise expressions

### LINKED ISSUE #599: `otherwise` is not optional in r-polars's when-then
```r
> library(polars)
> pl$when(pl$col("foo") > 2)$then(1)$alias("val")
Error: syntax error: alias is not a method/attribute of the class Then 
 when calling method:
 pl$when(pl$col("foo") > 2)$then(1)$alias
```

In py-polars

```python
>>> import polars as pl
>>> pl.when(pl.col("foo") > 2).then(1).alias("val")
<polars.expr.expr.Expr object at 0x7fa5ee7cc0a0>
```

eitsupi:
Close #599


NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 ## Polars R Package (development version)
 
+### New features
+
+-   In the when-then-otherwise expressions, the last `$otherwise()` is now optional,
+    as in Python Polars (#836).
```
etiennebacher, line 8:
```suggestion
-   In the when-then-otherwise expressions, the last `$otherwise()` is now optional,
    as in Python Polars. If `$otherwise()` is not specified, rows that don't respect the 
    condition set in `$when()` will be filled with `null` (#836).
```

R/functions__whenthen.R
```diff
@@ -1,31 +1,35 @@
...
 #'
-#' * `pl$when()`returns a `When` object,
-#' * `pl$then()`returns a `Then` object,
-#' * `<Then>$otherwise()`returns an [Expression][Expr_class] object,
-#' * `<Then>$when()`returns a `ChainedWhen` object,
-#' * `<ChainedWhen>$then()`returns a `ChainedThen` object,
-#' * `<ChainedThen>$otherwise()`returns an [Expression][Expr_class] object.
+#' If none of the conditions are `true`, an optional
+#' `$otherwise(<value if all statements are false>)` can be appended at the end.
+#' If not appended, and none of the conditions are `true`, `null` will be returned.
```
etiennebacher, line 14:
```suggestion
#' If not appended, and none of the conditions are `true`, it will return `null`.
```

R/functions__whenthen.R
```diff
@@ -1,31 +1,35 @@
...
 #'
+#' @section Expression methods:
+#' `RPolarsThen` objects and `RPolarsChainedThen` objects (returned by `$then`)
+#' stores the same methods as [Expr][Expr_class].
```
etiennebacher, line 18:
```suggestion
#' @details
#' `RPolarsThen` objects and `RPolarsChainedThen` objects (returned by `$then()`)
#' stores the same methods as [Expr][Expr_class].
```

R/functions__whenthen.R
```diff
@@ -1,31 +1,35 @@
...
 #' @name Expr_when_then_otherwise
-#' @param ... Expr or something coercible to an Expr into a boolean mask to
-#'   branch by.
-#' @param statement Expr or something coercible to an Expr value to insert in
-#'   when() or otherwise(). Strings interpreted as column.
-#' @return an polars object, see details.
+#' @param ... [Expr][Expr_class] or something coercible to an Expr into a boolean mask to
+#' branch by.
```
etiennebacher, line 21:
```suggestion
#' @param ... [Expr][Expr_class] or something coercible to an Expr that returns a 
#' boolean for each row.
```

R/functions__whenthen.R
```diff
@@ -1,31 +1,35 @@
...
+#' @param statement [Expr][Expr_class] or something coercible to
+#' an [Expr][Expr_class] value to insert in `then()` or `otherwise()`.
+#' Strings interpreted as column.
```
etiennebacher, line 24:
```suggestion
#' Strings are parsed as columns.
```

R/functions__whenthen.R
```diff
@@ -44,6 +48,13 @@
 #'   $otherwise(-1)
 #' )
 #'
+#' # The `$otherwise` at the end is optional.
+#' # If left out, any rows where none of the `$when` expressions are evaluate to `true`,
```
etiennebacher, line 52:
```suggestion
#' # If left out, any rows where none of the `$when()` expressions are evaluated to `true`,
```

etiennebacher (changes requested):
That looks great, thanks! Just some minor comments on docs

R/functions__whenthen.R
```diff
@@ -1,31 +1,35 @@
...
 #'
-#' `when-then-otherwise` is similar to R [`ifelse()`]. This has to start with
-#' `pl$when(<condition>)$then(<value if condition>)`. From there, it can:
-#' * be chained to an `$otherwise()` statement that specifies the Expr to apply
-#'   to the rows where the condition is `FALSE`;
-#' * or be chained to other `$when()$then()` to specify more cases, and then use
-#'   `$otherwise()` when you arrive at the end of your chain.
-#' Note that one difference with the Python implementation is that we *must*
-#' end the chain with an `$otherwise()` statement.
+#' `when-then-otherwise` is similar to R [`ifelse()`]. Always initiated by a
```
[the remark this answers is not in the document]
eitsupi: In fact, this text is copied from the Python package, so I would like to leave it as is.

R/functions__whenthen.R
eitsupi: Ditto.

etiennebacher (approved):

PR #840, opened 2024-02-24
feat: rewrite $head, $limit, and $tail to be equivalent to Python Polars

eitsupi:
I noticed that in Python, `polars.LazyFrame.head` calls the `slice` method internally.
https://github.com/pola-rs/polars/blob/740e740d9ce3678ea061d5cb4c2bc94892838383/py-polars/polars/lazyframe/frame.py#L4676-L4723

Here these functions are rewritten to match those in Python.
As a result, `<DataFrame>$head()` and `<DataFrame>$tail()` now support negative values.


R/lazyframe__lazy.R
```diff
@@ -844,24 +844,31 @@ LazyFrame_sink_ndjson = function(
...
-  unwrap(.pr$LazyFrame$limit(self, n), "in $limit():")
-}
-
-#' @title Head of a LazyFrame
-#' @inherit DataFrame_head description params details
+#' Get the first `n` rows.
 #'
-#' @examples pl$LazyFrame(mtcars)$head(4)$collect()
+#' A shortcut for [`$slice(0, n)`][LazyFrame_slice].
+#' Conider using the [`$fetch()`][LazyFrame_fetch] method if you want to test your query.
```
etiennebacher, line 850:
```suggestion
#' Consider using the [`$fetch()`][LazyFrame_fetch] method if you want to test your query.
```

NEWS.md
```diff
@@ -7,6 +7,8 @@
 -   In the when-then-otherwise expressions, the last `$otherwise()` is now optional,
     as in Python Polars. If `$otherwise()` is not specified, rows that don't respect
     the condition set in `$when()` will be filled with `null` (#836).
+-   `<DataFrame>$head()` and `<DataFrame>$tail()` methods now support negative
```
etiennebacher, line 10: also the case for `LazyFrame`, right?

R/lazyframe__lazy.R
```diff
@@ -844,24 +844,31 @@ LazyFrame_sink_ndjson = function(
...
+#' The [`$fetch()`][LazyFrame_fetch] operation will load the first `n` rows at
+#' the scan level, whereas `$head()` are applied at the end.
```
etiennebacher, line 852:
```suggestion
#' the scan level, whereas `$head()` is applied at the end.
```

etiennebacher (approved):
Thanks, LGTM, just a couple of typos.

I wonder why they keep both `limit()` and `head()` in the upstream polars if they are identical

NEWS.md
eitsupi: > also the case for `LazyFrame`, right?

No, LazyFrame does not know the total number of rows, so negative values are an error.

etiennebacher: Right

etiennebacher (approved):

eitsupi:
Thanks for your review.

> I wonder why they keep both `limit()` and `head()` in the upstream polars if they are identical

It is also strange that LazyFrame in Rust seems to have `limit` and `tail` but not `head`.

Perhaps the reason Python has `limit` as well as `head` is to match Rust.

PR #841, opened 2024-02-24
docs(vignette): tweak vignettes

eitsupi:
- Fix title
- Use `$method_name()` to method names
- Use `$fetch()` instead of `$limit()` (related to #840)
- Use `as.vector()` instead of `$to_vector()`

### DIFF
--- a/vignettes/polars.Rmd
+++ b/vignettes/polars.Rmd
@@ -191,7 +191,7 @@ instance,
 - Etc.
 
 In **polars** our method chaining syntax takes the form `object$m1()$m2()`,
-where `object` is our data object, and `m1()` and `m2()` are appropriate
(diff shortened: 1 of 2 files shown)


etiennebacher (approved):
Thanks!

vignettes/polars.Rmd
```diff
@@ -306,41 +306,44 @@ multiple variables and aggregation types.
 ```{r}
 dat$group_by(
   "cyl",
-  pl$col("am")$cast(pl$Boolean)$alias("manual")
+  manual = pl$col("am")$cast(pl$Boolean)
```
eitsupi, line 309: I noticed that the column renaming by named Expr is not working here. It needs to be fixed.

PR #842, opened 2024-02-24
chore: fix the example of `polars_info()` and change the type of `polars_info()$versions$r_package` to character

eitsupi:
I forgot to update the example in my last update.

In addition, minor modifications to the test are also made.


etiennebacher (commented):
While you're on this, should both version numbers return `package_version` objects?

``` r
x <- polars::polars_info()$versions
x
#> $r_package
#> [1] '0.14.1'
#> 
#> $rust_crate
#> [1] "0.37.0"

lapply(x, class)
#> $r_package
#> [1] "package_version" "numeric_version"
#> 
#> $rust_crate
#> [1] "character"
```

eitsupi:
> While you're on this, should both version numbers return `package_version` objects?

Good point.
Since Rust allows versions containing strings (e.g., `1.0.0rc`), it may be incompatible with the R's `numeric_version` object.

Should we change `r_package` to a character?

etiennebacher:
> Should we change r_package to a character?

I'm fine with that

etiennebacher:
Just for info, `arrow` returns a `package_version` (but they only return the R package version):

``` r
arrow::arrow_info()$version |> class()
#> [1] "package_version" "numeric_version"
```

eitsupi:
In fact, the R package version can be easily taken from within R, so there is little need to obtain it via this function.

etiennebacher (approved):

PR #844, opened 2024-02-24
refactor: remove some old implementations of methods for `Series`

### LINKED ISSUE #839: `as.list.Expr` is needed?
https://github.com/pola-rs/r-polars/pull/121#discussion_r1500957508

To me this does not seem useful.

etiennebacher:
Close #839 

Also remove some methods for `Series` since they automatically call `Expr` methods (followup of #831)


eitsupi (approved):
Good catch, thanks!

PR #849, opened 2024-02-25
refactor: remove the `data.table` package from Suggests

eitsupi:
In fact, it is used only within the tests and there is little need to use it.

### DIFF
--- a/tests/testthat/test-melt.R
+++ b/tests/testthat/test-melt.R
@@ -16,25 +16,8 @@ patrick::with_parameters_test_that("melt example",
         value = c(1, 3, 5, 2, 4, 6)
       )
     )
-
-    df_2 = pl[[create_func]](
-      a = c("x", "y", "z"),
-      b = c(1, 3, 5),
-      c = c(2, 4, 6)
-    )
-
-    expect_true(is_func(df_2))
-
     expect_identical(
-      df_2$melt(id_vars = "a", value_vars = c("b", "c")) |> as.data.frame(),
-      data.frame(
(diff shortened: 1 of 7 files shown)


etiennebacher (changes requested):
I'm ok to remove it but the examples that are removed should be replaced with a base R alternative. I see only one remaining test for `$melt()` and that's not enough for this functionality

eitsupi:
Thanks for your review.
I rewrote tests for `$melt()`.

tests/testthat/test-melt.R
```diff
@@ -0,0 +1,75 @@
...
+        variable = c("b", "b", "b", "c", "c", "c"),
+        value = c(1, 3, 5, 2, 4, 6)
+      )
+    )
+
+    df_2 = pl[[create_func]](
+      a = c("x", "y", "z"),
+      b = c(1, 3, 5),
+      c = c(2, 4, 6)
+    )
```
etiennebacher, line 24: This is identical to `df_1`?

tests/testthat/test-melt.R
```diff
@@ -0,0 +1,75 @@
...
+      ) |> as.data.frame(),
+      data.frame(
+        a = c("x", "y", "z"),
+        b = c(1, 3, 5),
+        bob = rep("c", 3),
+        alice = c(2, 4, 6)
+      )
+    )
+
+    expect_error
```
etiennebacher, line 70: missing something here

etiennebacher (changes requested):

etiennebacher (approved):
Thanks, LGTM

PR #850, opened 2024-02-25
ci(website): use `r-lib/actions/setup-r-dependencies` to cache R packages

eitsupi:
Currently the binary for the new version of the `arrow` package is not yet available via P3M, so that is installed source from CRAN.
Switching to `r-lib/actions/setup-r-dependencies` should speed up the process.


etiennebacher (approved):
Thanks

PR #873, opened 2024-03-01
Implement `$name$prefix_fields()` and `$name$suffix_fields()`

etiennebacher:


NEWS.md
```diff
@@ -48,6 +48,8 @@
     (#869).
 -   Converts `clock_time_point` and `clock_zoned_time` objects from
     the `{clock}` package to Polars datetime type (#861).
+-   New methods for the `name` subnamespace: `$prefix_fields()` and
```
eitsupi, line 51:
```suggestion
-   New methods for the `name` subnamespace: `$prefix_fields()` and
```

tests/testthat/test-expr_name.R
```diff
@@ -13,3 +13,71 @@ test_that("name to_uppercase", {
...
+        "was not a string"
+      )
+
+      # TODO: this works but always prints the error message of log("a"), how
+      # can we silence it?
+      # expect_error(
+      #   pl$DataFrame(list(alice=1:3))$select(
+      #     pl$col("alice")$name$map(\(x) log("a"))
+      #   )
+      # )
```
eitsupi, line 50: How about using `suppressWarnings`

eitsupi (changes requested):

tests/testthat/test-expr_name.R
etiennebacher: It doesn't work, `log("a")` generates an error, not a warning. I'd say we can skip this for now, there's another test above for wrong function output

eitsupi: Hmmm, I don't know what this test case is trying to test.
`log()` is an R function, not a polars one, right?
If it doesn't make sense, why not simply delete the comment?

etiennebacher: We want to test 2 types of errors: 

* the function called in `$map()` works but doesn't return a string -> Polars error
* the function called in `$map()` fails (like `log("a")` -> both R and Polars error, and we want to ensure that it doesn't lead to a panick or another issue with threads

The test that is commented out is for the second case. Problem is that it returns the two errors and `expect_error()` only hides the Polars one, not the R one. Wrapping `expect_error()` in a second `expect_error()` doesn't work either.

To clarify, if we uncomment this test, it correctly passes. The only issue is that it prints the error message of `log("a")` in the output

eitsupi: Thanks for the explanation. Okay, I think we can merge it now for now.

eitsupi (approved):

PR #879, opened 2024-03-01
[DO NOT MERGE] ci debug

eitsupi:
For https://github.com/pola-rs/polars/issues/14794#issuecomment-1973300864


eitsupi:
No longer needed. See https://github.com/pola-rs/polars/issues/14794#issuecomment-1994361609.

PR #882, opened 2024-03-02
Remove argument `eager` in `pl$date_range()` and `pl$struct()`

### LINKED ISSUE #880: Remove the `eager` argument of `pl$date_range()` and `pl$struct()`
I think it would be better to remove it and simplify it, since it simply allows the subsequent `$to_series()` to be omitted.

etiennebacher:
Close #880


eitsupi (approved):
Thanks!

Issue #551, opened 2023-11-28
Documents for S3 methods (of operators)


eitsupi:
Currently, the S3 methods of the base R functions are listed in a separate category in the documentation, but as for the operators, they belong to the functions to which they are translated.

https://github.com/pola-rs/r-polars/blob/2fbf67d0d8e4d2bcfcd317873c08c9fb464b0cc2/R/expr__expr.R#L170-L179

I think it would be more convenient to be able to easily check which functions are supported by S3, so I think it would be better to move these to the S3 category.

etiennebacher:
All S3 methods are now gathered in the same section: https://github.com/pola-rs/r-polars/blob/a80ca22fddc89f0c95edd984ec6f36f8b1e784e3/altdoc/mkdocs.yml#L702

PR #886, opened 2024-03-03
feat!: rewrite arithmetic operators for Expr and Series

eitsupi:
Part of #551

- Rewrite docs
- Rename `<Series>$rem()` -> `<Series>$mod()`
- Add missing operators for Series
- Other minor refactorings


tests/testthat/test-s3-methods-operator.R
```diff
@@ -0,0 +1,38 @@
...
+    ~.test_name, ~fn,
+    "add", `+`,
+    "sub", `-`,
+    "div", `/`,
+    "floor_div", `%/%`,
+    "mul", `*`,
+    "mod", `%%`,
+    "pow", `^`,
+  )
+}
```
etiennebacher, line 12: didn't we already have some parametric tests for that in test-expr_expr.R? It would be good to keep them in a single place

etiennebacher (changes requested):
Thanks, looks great! Just a couple of comments

eitsupi:
Thanks for your review!

PR #892, opened 2024-03-04
fix: the default value of `$str$strptime()`'s format

### LINKED ISSUE #884: Update `$str$strptime()`
The Python one will parse without explicit arguments, but the R one will not work unless the format is specified.
Finding the format in the chrono documentation can be a pain.

eitsupi:
Fix #884

The documentation said the default was `NULL`, but in fact there was no default value and an error would occur if the user did not specify one.

In addition, documentation has been improved and tests for default parsable values have been added.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-- The default value of the `format` of `$str$strptime()` is now correctly set (#892).
+-   The default value of the `format` of `$str$strptime()` is now correctly set (#892).
 
 ## Polars R Package 0.15.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 ## Polars R Package (development version)
 
+### Bug fixes
+
+- The default value of the `format` of `$str$strptime()` is now correctly set (#892).
+
 ## Polars R Package 0.15.0
 
 ### Breaking changes due to Rust-polars update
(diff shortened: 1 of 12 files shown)


NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## Polars R Package (development version)
 
+### Bug fixes
+
+- The default value of the `format` of `$str$strptime()` is now correctly set (#892).
```
etiennebacher, line 7: Can you add a tab instead of a space after `-`?

R/expr__string.R
```diff
@@ -81,93 +105,78 @@ ExprStr_strptime = function(
...
-#'   with_columns(datetime = pl$col("str_date")$str$to_datetime(strict = FALSE))
+#' @inheritParams ExprStr_strptime
+#' @param time_unit Unit of time for the resulting Datetime column. If `NULL` (default),
+#' the time unit is inferred from the format string if given,
+#' eg: `"%F %T%.3f"` => [`pl$Datetime("ms")`][pl_Datetime].
+#' If no fractional second component is found, the default is `"us"` (microsecond).
+#' @param time_zone for the resulting [Datetime][pl_Datetime] column.
+#' @param exact If `TRUE` (default), require an exact format match. If `FALSE`, allow the format to match
+#' anywhere in the target string. Note that using `exact = FALSE` introduces a performance
+#' penalty - cleaning your data beforehand will almost certainly be more performant.
```
etiennebacher, line 171: Aren't those already inherited from `$strptime()`?

eitsupi: I recall, because of the reference to the Time type, it cannot be reused.

etiennebacher (approved):

PR #896, opened 2024-03-05
feat: some improves of `as_polars_df`

### LINKED ISSUE #755: Rewrite `as_polars_df.nanoarrow_array`
I think we should add `as_polars_df.nanoarrow_array` and use that inside of `as_polars_df.nanoarrow_array_stream`. (concat all DataFrames)

May require the `vstack` method.
https://stackoverflow.com/questions/71654966/how-can-i-append-or-concatenate-two-dataframes-in-python-polars

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/pull/730#discussion_r1471397822_

### LINKED ISSUE #893: `as_polars_df()` for `nanoarrow_array_stream` seems slow
Slower than `as_tibble` surprised me.

``` r
library(adbcdrivermanager)
library(arrow)
#>
#> Attaching package: 'arrow'
#> The following object is masked from 'package:utils':
#>
#>     timestamp
library(tibble)
library(polars)

polars_info()
#> Polars R package version : 0.15.0
#> Rust Polars crate version: 0.38.1
#>
#> Thread pool size: 16
#>
#> Features:
#> default                    TRUE
#> full_features              TRUE
#> disable_limit_max_threads  TRUE
#> nightly                    TRUE
#> sql                        TRUE
#> rpolars_debug_print       FALSE
#>
#> Code completion: deactivated

db <- adbc_database_init(adbcsqlite::adbcsqlite())
con <- adbc_connection_init(db)

flights <- nycflights13::flights
flights$time_hour <- NULL
flights |>
  write_adbc(con, "flights")

query <- "SELECT * from flights LIMIT 10000"

bench::mark(
  polars_df_1 = {
    con |>
      read_adbc(query) |>
      as_polars_df()
  },
  arrow_table = {
    con |>
      read_adbc(query) |>
      as_arrow_table()
  },
  tibble = {
    con |>
      read_adbc(query) |>
      as_tibble()
  },
  polars_df_2 = {
    con |>
      read_adbc(query) |>
      as_polars_df()
  },
  check = FALSE,
  min_iterations = 5
)
#> # A tibble: 4 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 polars_df_1  158.5ms  158.8ms      5.58    1.13MB     3.72
#> 2 arrow_table   36.5ms     39ms     25.6     1.62MB     2.13
#> 3 tibble        40.8ms   43.2ms     22.9     1.62MB     0
#> 4 polars_df_2  153.8ms  185.8ms      5.38    89.1KB     8.07
```

<sup>Created on 2024-03-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

eitsupi:
- Simplify `as_polars_df(<nanoarrow_array_stream>)` (Fix #893)
- Add `as_polars_df(<nanoarrow_array>)` (Close #755)
- Simplify `as_polars_df(<data.frame>)` and fix some test and docs.

The performance of the conversion from `nanoarrow_array_stream` is as follows, slower than `as_arrow_table` but faster than `as_tibble`, possibly due to the String type conversion. (And I am surprised that `as_tibble` is so fast)

``` r
library(adbcdrivermanager)
library(arrow)
#>
#> Attaching package: 'arrow'
#> The following object is masked from 'package:utils':
#>
#>     timestamp
library(tibble)
library(polars)

polars_info()
#> Polars R package version : 0.15.0.9000
#> Rust Polars crate version: 0.38.1
#>
#> Thread pool size: 16
#>
#> Features:
#> default                    TRUE
#> full_features              TRUE
#> disable_limit_max_threads  TRUE
#> nightly                    TRUE
#> sql                        TRUE
#> rpolars_debug_print       FALSE
#>
#> Code completion: deactivated

db <- adbc_database_init(adbcsqlite::adbcsqlite())
con <- adbc_connection_init(db)

flights <- nycflights13::flights
flights$time_hour <- NULL
flights |>
  write_adbc(con, "flights")

query <- "SELECT * from flights"

bench::mark(
  polars_df_1 = {
    con |>
      read_adbc(query) |>
      as_polars_df()
  },
  arrow_table = {
    con |>
      read_adbc(query) |>
      as_arrow_table()
  },
  tibble = {
    con |>
      read_adbc(query) |>
      as_tibble()
  },
  polars_df_2 = {
    con |>
      read_adbc(query) |>
      as_polars_df()
  },
  check = FALSE,
  min_iterations = 5
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 4 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 polars_df_1    1.62s    1.78s    0.116     4.94MB   0.162
#> 2 arrow_table    1.25s    1.39s    0.722     1.62MB   0
#> 3 tibble          3.3s    5.39s    0.0938    46.5MB   0.0751
#> 4 polars_df_2    1.58s     3.4s    0.223   149.91KB   0.134
```

<sup>Created on 2024-03-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/as_polars.R
```diff
@@ -202,8 +202,8 @@ as_polars_df.RPolarsLazyGroupBy = function(x, ...) {
 #' @param rechunk A logical flag (default `TRUE`).
 #' Make sure that all data of each column is in contiguous memory.
 #' @param schema named list of DataTypes, or character vector of column names.
-#' Should be the same length as the number of columns of `x`.
-#' If schema names or types do not match `x`, the columns will be renamed/recast.
+#' Should match the number of columns in `x` and correspond to each column in `x` by position.
+#' If a column in `x` does not match the name or type at the same position, it will be renamed/recast.
```
etiennebacher, line 206: Wow I didn't notice that it would automatically rename / recast but that seems like a really bad behavior:

``` r
library(polars)
df = data.frame(a = 1:3, b = 4:6)
as_polars_df(df, schema = list(b = pl$String, y = pl$Int32))
#> shape: (3, 2)
#> ┌─────┬─────┐
#> │ b   ┆ y   │
#> │ --- ┆ --- │
#> │ str ┆ i32 │
#> ╞═════╪═════╡
#> │ 1   ┆ 4   │
#> │ 2   ┆ 5   │
#> │ 3   ┆ 6   │
#> └─────┴─────┘
```

This doesn't work in py-polars:

```python
test = pl.DataFrame(
    {
        "a": [1, 2, 3],
        "b": [4, 5, 6],
    },
    schema={"b": pl.String, "y": pl.Int32},
)
ValueError: the given column-schema names do not match the data dictionary

pl.DataFrame(
    {
        "a": [1, 2, 3],
        "b": [4, 5, 6],
    },
    schema_overrides={"b": pl.String, "y": pl.Int32},
)

shape: (3, 2)
┌─────┬──────┐
│ a   ┆ b    │
│ --- ┆ ---  │
│ i64 ┆ str  │
╞═════╪══════╡
│ 1   ┆ null │
│ 2   ┆ null │
│ 3   ┆ null │
└─────┴──────┘
```
This wasn't introduced in this PR so I don't think it's a blocker, but it should definitely be fixed

etiennebacher (changes requested):
Thanks, do you want to bump NEWS to add a note on improved performance?

etiennebacher (approved):
Thanks, do you want to bump NEWS to add a note on improved performance?

R/as_polars.R
eitsupi: Like this:

```python
>>> import polars as pl
>>> import pyarrow as pa
>>> data = pa.table({"a": [1, 2, 3], "b": [4, 5, 6]})
>>> pl.from_arrow(data, schema = [("b", pl.String), ("y", pl.Int32)])
shape: (3, 2)
┌─────┬─────┐
│ b   ┆ y   │
│ --- ┆ --- │
│ str ┆ i32 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 2   ┆ 5   │
│ 3   ┆ 6   │
└─────┴─────┘
```

etiennebacher: Hum, so I guess they need to harmonize the behavior upstream between `pl.DataFrame` and `pl.from_arrow` then?

eitsupi: It is quite surprising that this behavior is performed, especially when specified with the dict type.

In general, order in dict is not guaranteed, so processing based on order is not a good idea.

eitsupi: > Hum, so I guess they need to harmonize the behavior upstream between `pl.DataFrame` and `pl.from_arrow` then?

Not sure about that.
Perhaps the `schema` is intended for Series rather than DataFrame (e.g., `pyarrow.Array` or something has no name), since it is guaranteed that the column names already exist in the `arrow::Table` for `as_polars_df`, so we can simply remove the `schema` argument from here and use only `schema_override`.

eitsupi: In my opinion, there is no user demand to change column names in `as_polars_df()`.
So simply removing the `schema` argument and leaving only `schema_override` would be sufficient.

In terms of type change, the `schema` argument is more difficult to use than `schema_override` in that all columns must be specified.

Issue #686, opened 2024-01-09
Improve search on website


etiennebacher:
The search doesn't take into account the function names in the sidebar in "Reference", which makes the search experience a bit useless. For example, searching "sum_horizontal" doesn't point to `pl$sum_horizontal()` but only shows the output when we print all methods in one of the examples:

![image](https://github.com/pola-rs/r-polars/assets/52219252/b76df8a6-0829-4a72-adc8-0ede2798ebe8)

etiennebacher:
The problem is that the search index is based on the md files, not the HTML ones: in the md file, `pl_sum_horizontal` exists, but not simply `sum_horizontal`. In addition to that, it can't search for chunks of words (for example searching for "ackage" doesn't suggest the word "package"). 

Searching for "pl_sum_horizontal" works fine (but is clearly not ideal because we'd need to prefix lazy functions with "LazyFrame_", etc.):
![image](https://github.com/pola-rs/r-polars/assets/52219252/f66ee329-db6a-443c-8e0e-4dd5e8a9450b)

etiennebacher:
That is something to fix in `altdoc` anyway because the search index is built when we run `render_docs()`

etiennebacher:
This is annoying, the website loses half of its interest if search is impossible (or only with the trick of searching for e.g "DataFrame_"). To summarize the comments above, the problem is that the search index used by `mkdocs` is built using the markdown files (that are then converted to HTML by `mkdocs`). This is why we need to type "DataFrame_" to search in DataFrame methods: the "altdoc postprocessing" script that fixes the Usage section by going from "DataFrame_" to "\<DataFrame\>$" comes after the search index is built.

We may think that this can be fixed by running the "altdoc postprocessing" script earlier in the workflow, basically after every `devtools::document()`. However, this is not correct because R CMD check will error when it parses the ("fixed") Rd files (cf #701). 

Therefore, one solution would be to do this step only in the `altdoc` workflow. This is what it would look like:

* make a copy of the "man" folder (with the original Rd files) named "man_old"
* run the script in #701 on the "man" folder. This fixes the formatting in the Usage section.
* run the `altdoc` workflow (preprocessing script + `render_docs()`). Since it runs on the fixed "man" folder, the search index should be built correctly (typing "DataFrame$" in the search bar should point to DataFrame methods)
* remove the "man" folder, and rename "man_old" to "man". Therefore, the "man" folder is now the same as what was generated by `devtools::document()`

[...]

PR #905, opened 2024-03-09
build(website): use the preview version of mkdocs-material

eitsupi:
Related to #686

See the comment https://github.com/pola-rs/r-polars/issues/686#issuecomment-1986846558


etiennebacher:
FYI I tried this locally and it doesn't seem to make a difference in the search results

Issue #912, opened 2024-03-11
Rewrite `pl$col()`


eitsupi:
The `name` argument may cause unexpected behavior.

PR #914, opened 2024-03-11
Change `$drop()` input from `columns` to `...`

### LINKED ISSUE #913: Update `$drop()` input
I'm surprised this wasn't needed before , I think `DataFrame$drop()` should have a similar input as `DataFrame$partition_by()` based on the py-polars API

_Originally posted by @etiennebacher in https://github.com/pola-rs/r-polars/pull/898#discussion_r1518979645_

etiennebacher:
Close #913

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,10 +2,11 @@
 
 ## Polars R Package (development version)
 
-### New features
+### Breaking changes
 
-- `$drop()` now accepts several character vectors, such as `$drop("a", "b")` 
-  (#912).
+- The argument `columns` in `$drop()` is removed. `$drop()` now accepts several 
+  character scalars, such as `$drop("a", "b", "c")`. Explicitly using the 
+  `columns` argument now errors (#912).
 
 ## Polars R Package 0.15.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,11 @@
 
 ## Polars R Package (development version)
 
+### New features
+
+- `$drop()` now accepts several character vectors, such as `$drop("a", "b")` 
+  (#912).
+
 ## Polars R Package 0.15.1
 
 ### New features
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -115,7 +115,7 @@ Collate:
     'sql.R'
     'vctrs.R'
     'zzz.R'
-Config/rextendr/version: 0.3.1
+Config/rextendr/version: 0.3.1.9000
 VignetteBuilder: knitr
 Config/polars/LibVersion: 0.38.1
 Config/polars/RustToolchainVersion: nightly-2024-02-23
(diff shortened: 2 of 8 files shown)


NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 ## Polars R Package (development version)
 
+### New features
+
+- `$drop()` now accepts several character vectors, such as `$drop("a", "b")`
```
eitsupi, line 7: I suspect this is a potentially breaking change.
In other words, if you specify it with the columns argument instead of the position argument, won't unexpected behavior occur?

eitsupi (approved):

NEWS.md
etiennebacher: Good catch. Actually I think there's a bug in `dots_to_colnames()` because passing `c()` in second values doesn't work:

```r
### this works
pl$DataFrame(mtcars)$drop(c("mpg", "drat"), "hp") |> ncol()
[1] 8

### this doesn't
pl$DataFrame(mtcars)$drop("hp", c("mpg", "drat")) |> ncol()

   0: In R: in $drop():
   0: During function call [ncol(pl$DataFrame(mtcars)$drop("hp", c("mpg", "drat")))]
   1: When constructing a Column Expr
   2: The argument [ `...` ] caused an error
   3: Possibly because element no. [1] 
   4: Expected a value of type [alloc::string::String]
   5: Got value [Rvalue: ["mpg", "drat"], Rsexp: Strings, Rclass: ["character"]]
```

eitsupi: This is a problem of `pl$col()`, not `dots_to_colnames()`.

```r
> library(polars)

> pl$col(c("mpg", "drat"), "hp")
polars Expr: cols(["mpg", "drat", "hp"])

> pl$col("hp", c("mpg", "drat"))
   0: In R: in pl$col()
   0: During function call [pl$col("hp", c("mpg", "drat"))]
   1: When constructing a Column Expr
   2: The argument [ `...` ] caused an error
   3: Possibly because element no. [1] 
   4: Expected a value of type [alloc::string::String]
   5: Got value [Rvalue: ["mpg", "drat"], Rsexp: Strings, Rclass: ["character"]]

> pl$col("hp", name = c("mpg", "drat"))
polars Expr: cols(["mpg", "drat", "hp"])

> pl$col("hp", columns = c("mpg", "drat"))
   0: In R: in pl$col()
   0: During function call [pl$col("hp", columns = c("mpg", "drat"))]
   1: When constructing a Column Expr
   2: The argument [ `...` ] caused an error
   3: Possibly because element no. [1] 
   4: Expected a value of type [alloc::string::String]
   5: Got value [Rvalue: ["mpg", "drat"], Rsexp: Strings, Rclass: ["character"]]
```

We should rewrite `pl$col()`...(#912)

eitsupi: Something like `pl$DataFrame(mtcars)$drop(columns = c("mpg", "drat"))` works in the main branch, but should fail in this PR because of `pl$col(name = "", ...)`.

etiennebacher: Indeed, I've updated news, thanks

PR #925, opened 2024-03-15
docs(website): improve search

### LINKED ISSUE #686: Improve search on website

etiennebacher:
Close #905 

It seems that the `index.json` used for search cannot take into account pieces of words, e.g `sum_hor` in `pl_sum_horizontal`, which explains why the search results were so bad. 

This PR adds custom separators when building the `index.json` for search, meaning that `pl_` and other prefixes are no longer taken into account when looking for function names. Note that `[\\s\\-]+` is the default, so only what comes after is customized.

Here are some search suggestions before and after this change.

### `sum_horizontal`

Old:

image

New:

image

### `strptime`

Old:

image

New:

image

### `sink`

Old:

image

New:

image


altdoc/mkdocs_static.yml
```diff
@@ -7,7 +7,8 @@ repo_name: $ALTDOC_PACKAGE_NAME
 
 ### Plugins
 plugins:
-  - search
+  - search:
+      separator: "[\\s\\-]+|(pl_|DataFrame_|LazyFrame_|Expr_|ExprArr_|ExprBin_|ExprCat_|ExprDT_|ExprList_|ExprMeta_|ExprName_|ExprStr_|ExprStruct_)"
```
eitsupi, line 11: Can't it just be an underscore `_`?

etiennebacher: Nope, the results are not good with this when we look for functions that have a `_` like `sum_horizontal()` or `concat_str()`. I'll add a section in the preprocessing script to automatically add those prefixes.

eitsupi:
I understand. But I don't want to close #686 because this is not a fundamental solution.
Do we need to also add `$` as a separator?

etiennebacher:
> I don't want to close #686 because this is not a fundamental solution. 

We can close #686 later, I've updated the initial post of this PR.

> Do we need to also add `$` as a separator?

No because the search index is built on the Markdown files, which still have the `_` as separator, e.g `DataFrame_clone()`. We only replace those by `$` in the post-processing script, which is the last thing to run.

eitsupi:
> No because the search index is built on the Markdown files, which still have the `_` as separator, e.g `DataFrame_clone()`. We only replace those by `$` in the post-processing script, which is the last thing to run.

I understand that. My point was that since $ is used in places such as examples, shouldn't that also be registered as a separator?

etiennebacher:
> My point was that since $ is used in places such as examples, shouldn't that also be registered as a separator?

I added it, it allows a bit more matches indeed

eitsupi (approved):

PR #926, opened 2024-03-16
Revert refactor of `pl$DataFrame()` with `schema` only

etiennebacher:
Refactor in #904 considerably slowed down the `DataFrame` creation with `schema` only:

``` r
library(polars)

schema <- list()
for (i in 1:1000) {
  schema[[paste0("var_", i)]] <- pl$Float64
}

bench::mark(
  pre_904 = {
    largs <- lapply(seq_along(schema), \(x) {
      pl$lit(numeric(0))$cast(schema[[x]])$alias(names(schema)[x])
    })
    pl$select(largs)
  }, 
  post_904 = {
    lapply(schema, \(dtype) pl$Series(NULL)$cast(dtype)) |>
      pl$select()
  }
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 pre_904    992.55ms 992.55ms     1.01     10.5MB     17.1
#> 2 post_904      2.99s    2.99s     0.335    14.7MB     10.4
```

PR #927, opened 2024-03-16
Export `is_polars_dtype()`

### LINKED ISSUE #924: export `is_polars_dtype`

etiennebacher:
Close #924


etiennebacher:
Added small change from this comment: https://github.com/pola-rs/r-polars/pull/926#discussion_r1527500867

R/datatype.R
```diff
@@ -85,12 +85,20 @@ print.RPolarsDataType = function(x, ...) {
...
-#' @return a list DataType with an inner DataType
-#' @examples .pr$env$is_polars_dtype(pl$Int64)
+#' Check if input is a valid DataType
+#'
+#' @param x An object to be tested.
+#' @param include_unknown If `FALSE` (default), `pl$Unknown` is considered as
+#' an invalid datatype.
+#'
+#' @export
+#' @return A boolean scalar.
```
eitsupi, line 95: Is the expression "A boolean scalar" common?

etiennebacher: No, I just introduced it when working on the Series documentation to differentiate with "A numeric scalar". I can replace it with "A boolean" or just "Boolean", but then I guess I should also replace "A numeric scalar" by something?

eitsupi: I think so. I am sure it is not just a problem here.......

eitsupi: In general, I think these are "logical", not "boolean".

etiennebacher: `?all` and `?any` say:

> The value is a logical vector of length one.

But that's very verbose IMO

eitsupi: Pls use this:
https://github.com/pola-rs/r-polars/blob/4307c837b6f02022d550d7525f96fe58d49e95b0/R/is_polars.R#L5

eitsupi: And move this function to `is_polars.R`

eitsupi: So my point is that neither "scalar" nor "boolean" are formal R terms and should not be used in the documentation.

etiennebacher: I did the replacement for "scalar", but except in some internal function names (`is_numeric_of_length_1()` would be very verbose)

eitsupi: Thanks, looks good.

etiennebacher: For "boolean", it's a bit trickier in some places because the datatype is also named "Boolean", so depending on the context we can prefer "logical" or "boolean". I made some replacements where it's clear we want "logical" but I leave the rest for later.

DESCRIPTION
```diff
@@ -115,7 +115,7 @@ Collate:
     'sql.R'
     'vctrs.R'
     'zzz.R'
-Config/rextendr/version: 0.3.1
+Config/rextendr/version: 0.3.1.9000
```
eitsupi, line 118: Is this needed?

R/is_polars.R
```diff
@@ -39,3 +39,21 @@ is_polars_lf = function(x) {
 is_polars_series = function(x) {
   inherits(x, "RPolarsSeries")
 }
+
+#' Test if the object a polars DataType
+#'
+#' @param x An object to be tested.
```
eitsupi, line 45: I think we can use inherit feature of roxygen2.

R/is_polars.R
```diff
@@ -39,3 +39,21 @@ is_polars_lf = function(x) {
...
+#' @param include_unknown If `FALSE` (default), `pl$Unknown` is considered as
+#' an invalid datatype.
+#'
+#' @export
+#' @return A logical value
```
eitsupi, line 50: Ditto.

eitsupi (approved):

PR #928, opened 2024-03-16
Implement `$to_dot()` for `LazyFrame`

etiennebacher:
``` r
library(polars)

query = pl$LazyFrame(mtcars)$
  filter(pl$col("drat") > 3)$
  with_columns(foo = pl$col("mpg") + pl$col("cyl"), bar = pl$mean("mpg"))

query$show_graph(raw_output = TRUE) |> cat()
#> graph  polars_query {
#> "WITH COLUMNS [\"foo\",\"bar\"] [(0, 0)]" -- "TABLE
#> π */11;
#> σ (col(\"drat\")) > (3.0) [(0, 1)]"
#> 
#> "WITH COLUMNS [\"foo\",\"bar\"] [(0, 0)]"[label="WITH COLUMNS [\"foo\",\"bar\"]"]
#> "TABLE
#> π */11;
#> σ (col(\"drat\")) > (3.0) [(0, 1)]"[label="TABLE
#> π */11;
#> σ (col(\"drat\")) > (3.0)"]
#> 
#> }

query$show_graph(width = "400px")
```


etiennebacher:
I don't understand why snapshots fail on some platforms only. @eitsupi can you reproduce some of those failures locally?

R/lazyframe__lazy.R
```diff
@@ -1952,3 +1952,60 @@ LazyFrame_group_by_dynamic = function(
...
+    raw_output = FALSE,
+    type_coercion = TRUE,
+    predicate_pushdown = TRUE,
+    projection_pushdown = TRUE,
+    simplify_expression = TRUE,
+    slice_pushdown = TRUE,
+    comm_subplan_elim = TRUE,
+    comm_subexpr_elim = TRUE,
+    streaming = FALSE,
+    ...) {
```
eitsupi, line 1988: Is there a reason for the order of the arguments?
It seems that positional arguments are not allowed in Python.

tests/testthat/_snaps/lazy_profile/basic-query-graph.svg
eitsupi, line 1: According to GitHub, this file seems to be broken.

eitsupi (changes requested):

R/lazyframe__lazy.R
etiennebacher: That was just a mistake, fixed now

tests/testthat/_snaps/lazy_profile/basic-query-graph.svg
etiennebacher: I removed this one, it's too annoying to do a snapshot of an SVG, testing the text output should be enough

R/lazyframe__lazy.R
```diff
@@ -1952,3 +1952,60 @@ LazyFrame_group_by_dynamic = function(
...
+    unwrap("in $show_graph():")
+
+  if (isTRUE(raw_output)) {
+    return(graph)
+  }
+
+  if (!requireNamespace("DiagrammeR", quietly = TRUE)) {
+    stop('The package "DiagrammeR" is required.')
+  }
+  DiagrammeR::grViz(graph, ...)
```
eitsupi, line 2010: I am not sure if it is a good idea to pull this package into a dependency here.
I think the situation is different in R and Python in that the user simply needs to run `<LazyFrame>$show_graph() |> DiagrammeR::grViz()`, and there are several other packages with similar functionality.

eitsupi:
> I don't understand why snapshots fail on some platforms only. @eitsupi can you reproduce some of those failures locally?

I found that running the function over and over did not always return the same result each time.

```r
> query$show_graph(raw_output = TRUE) |> cat()
graph  polars_query {
"WITH COLUMNS [\"foo\",\"bar\"] [(0, 0)]" -- "TABLE
π */11;
σ (col(\"drat\")) > (3.0) [(0, 1)]"

"TABLE
π */11;
σ (col(\"drat\")) > (3.0) [(0, 1)]"[label="TABLE
π */11;
σ (col(\"drat\")) > (3.0)"]
"WITH COLUMNS [\"foo\",\"bar\"] [(0, 0)]"[label="WITH COLUMNS [\"foo\",\"bar\"]"]

}
> query$show_graph(raw_output = TRUE) |> cat()
graph  polars_query {
"WITH COLUMNS [\"foo\",\"bar\"] [(0, 0)]" -- "TABLE
π */11;
σ (col(\"drat\")) > (3.0) [(0, 1)]"

"WITH COLUMNS [\"foo\",\"bar\"] [(0, 0)]"[label="WITH COLUMNS [\"foo\",\"bar\"]"]
"TABLE
π */11;
σ (col(\"drat\")) > (3.0) [(0, 1)]"[label="TABLE
π */11;
σ (col(\"drat\")) > (3.0)"]
```

R/lazyframe__lazy.R
etiennebacher: It would be great if there was a way to use `ggplot2` to get a static output (and not an `htmlwidget`), especially since we already suggests `ggplot2`

eitsupi: `Rgraphviz` and `DiagrammeR` are listed here.
<https://graphviz.org/resources/>

Since it is easy to chain functions by pipe operators in R, unlike Python, I feel it would be better to simply implement `$to_dot()` and let the user create the plot in a function of their choice.

etiennebacher: Yes, agreed

eitsupi (approved):
Thanks!

R/lazyframe__lazy.R
```diff
@@ -1952,3 +1952,57 @@ LazyFrame_group_by_dynamic = function(
...
+    optimized = TRUE,
+    type_coercion = TRUE,
+    predicate_pushdown = TRUE,
+    projection_pushdown = TRUE,
+    simplify_expression = TRUE,
+    slice_pushdown = TRUE,
+    comm_subplan_elim = TRUE,
+    comm_subexpr_elim = TRUE,
+    streaming = FALSE) {
+  lf = self$set_optimization_toggle(
```
eitsupi, line 1995: `lf` is assigned but never used.

PR #931, opened 2024-03-17
docs: skip the example if `arrow` package's dataset feature is not available

eitsupi:
Recently CI on macOS started failing. Probably because the binary of the arrow package on CRAN does not have the dataset feature enabled.


etiennebacher (approved):
Thanks!

Issue #903, opened 2024-03-08
The arguments posision of `pl$Series` should be change to match with Python Polars?


eitsupi:
In Python:

```py
pl.Series("a", [1, 2, 3])
```

In R:

```r
pl$Series(c(1, 2, 3), "a")
```

And the document says:

> It is possible to construct a Series with values as the first positional argument. This syntax considered an anti-pattern

https://github.com/pola-rs/polars/blob/25d0a2f028bd2de0bdc140a2c040f2e2b4c2f46e/py-polars/polars/series/series.py#L225-L227

I suspect that `as_polars_series` should be used as a function that takes a vector as its first argument, and `pl$Series` should be rewritten in the argument position to match Python.
This would be a very huge breaking change, so we will probably have to continue warnings for several months and then make the change.

@etiennebacher Thoughts?

eitsupi:
@grantmcdermott Any thoughts on this?

grantmcdermott:
> This would be a very huge breaking change, so we will probably have to continue warnings for several months and then make the change.

Oof, I think you're correct and there's probably no other way around it. Consistency across r-polars and py-polars has always been a (the?) North Star for this project, so this is just an unfortunate cost.

etiennebacher:
If we do this, yes we should add a deprecation warning for a few releases before changing the behavior. However, I'd rather wait for https://github.com/pola-rs/polars/issues/13405 to be settled before going forward with this. It seems like the order of arguments is unlikely to change but let's wait for a final decision

eitsupi:
Thanks both.

> However, I'd rather wait for pola-rs/polars#13405 to be settled before going forward with this. It seems like the order of arguments is unlikely to change but let's wait for a final decision

If a change is made in Python, I think it can be reverted, but shouldn't it be fixed that the function argument is incorrect right now?

etiennebacher:
I don't understand your point. Right now, they `name` - `values` and we have `values` - `name`. If they end up changing to `values` - `name`, then our implementation becomes correct, right? We must add a deprecation warning only if we're sure they stay with `name` - `values`.

eitsupi:
I'm just suspicious that it needs to be changed because they're there right now.
At the very least, I think it's worth warning that positional is vulnerable, and always recommending named arguments. (By the way, the `value` argument doesn't exist in R right now, so you need to change that as well.)

eitsupi:
pola-rs/polars#13405 has been marked as "not planed", so we will change the argument order.

PR #933, opened 2024-03-18
refactor!: rename the `pl$Series()`'s argument

eitsupi:
Related to #903


eitsupi:
The workflow failure is probably related to Rtools44, but I don't know what caused it.

R/series__series.R
```diff
@@ -265,7 +265,7 @@ Series_struct = method_as_active_binding(\() series_make_sub_ns(self, expr_struc
 #' s2 = pl$Series(1:3, "a", dtype = pl$Float32)
 #' s2
 pl_Series = function(
-    x,
+    values,
```
etiennebacher, line 268: The problem here is that code that used `x = foo` will be silently captured by `...`, which breaks the expected usage. Can you check that `x` is in `...` and throw a warning/error if it is the case? (and add a test for this scenario)

etiennebacher (changes requested):

R/series__series.R
eitsupi: In my opinion, if such usage were done, an error would occur somewhere in the code and people would notice the breaking change.
Of course we can capture `x` as you suggest, but I don't think it is worth doing as I have written here.
https://github.com/pola-rs/r-polars/pull/939#pullrequestreview-1946651267

etiennebacher (approved):

PR #934, opened 2024-03-18
refactor: use `map_err` for error handling

eitsupi:


etiennebacher:
Not convinced this is worth it, it makes the error message less readable:

* `main`
```r
> pl$col(list("a"), list("b"))
   0: In R: in pl$col():
   0: During function call [pl$col(list("a"), list("b"))]
   1: pl$col()'s arguments must be one of the following:
      - character vectors
      - RPolarsDataTypes
      - a list of RPolarsDataTypes
  ```

* PR
```r
> pl$col(list("a"), list("b"))
   0: In R: in pl$col():
   0: During function call [pl$col(list("a"), list("b"))]
   1: pl$col()'s arguments must be one of the following:
      - character vectors
      - RPolarsDataTypes
      - a list of RPolarsDataTypes
   2: The argument [list_of_dtypes] caused an error
   3: Possibly because element no. [1] 
   4: Expected a value of type [r_polars::rdatatype::RPolarsDataType]
   5: Got value [list!(["a"])]
   6: ExpectedExternalPtr(list!(["a"]))
```

eitsupi:
@etiennebacher I reverted the error message.

Issue #921, opened 2024-03-15
More consistency in methods to convert Series/DataFrame to R


etiennebacher:
Currently this is what we have for `DataFrame` and `Series`:

``` r
library(polars)

options(polars.do_not_repeat_call = TRUE)

####### Series

series_vec = pl$Series(letters[1:3])

# output depends on datatype
series_vec$to_r()
#> [1] "a" "b" "c"

# identical
series_vec$to_r_vector() 
#> [1] "a" "b" "c"
series_vec$to_vector() 
#> [1] "a" "b" "c"

# not identical
series_vec$to_r_list() 
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] "b"
#> 
#> [[3]]
#> [1] "c"
series_vec$to_list() 
#> Error: Execution halted with the following contexts
#>    1: $ - syntax error: to_list is not a method/attribute of the class RPolarsSeries


####### DataFrame

df = pl$DataFrame(x = letters[1:3])

# output depends on datatype
df$to_r()
#> Error: Execution halted with the following contexts
#>    1: $ - syntax error: to_r is not a method/attribute of the class RPolarsDataFrame

# not identical
df$to_r_data_frame() 
#> Error: Execution halted with the following contexts
#>    1: $ - syntax error: to_r_data_frame is not a method/attribute of the class RPolarsDataFrame
df$to_data_frame() 
#>   x
#> 1 a
#> 2 b
#> 3 c

# not identical
df$to_r_list() 
#> Error: Execution halted with the following contexts
#>    1: $ - syntax error: to_r_list is not a method/attribute of the class RPolarsDataFrame
df$to_list()
#> $x
#> [1] "a" "b" "c"
```

I think we should have more consistency in the names of the methods:

* remove the `_r_` in the names of Series methods, so that we have `to_r()`, `to_list()` and `to_vector()`. That would be more consistent with `to_data_frame()` and `to_list()` that we have for `DataFrame`.
* I'm even wondering if we should have `to_r()` for Series. This is quite vague about what is returned, so it's only useful when we want to convert a Series to R no matter what, and I'm not sure we should allow that.

@eitsupi what do you think? Also, @grantmcdermott in case you want to participate (mostly about the second point)

eitsupi:
Agree.
[...]

PR #938, opened 2024-03-19
Harmonize functions to convert from Polars to R

etiennebacher:
Part of #921.

First of 3 PRs: this one, one to deprecate `$to_r()` (if we agree on that), and one to replace the `$to_*()` functions by their `as.*()` counterparts in the docs.

@eitsupi do you think we should deprecate those functions instead of plainly removing them?


eitsupi (approved):
Thanks, I think removing them makes sense.

PR #939, opened 2024-03-19
Align arguments of several string to date(time) functions with python polars

etiennebacher:


R/expr__string.R
```diff
@@ -176,12 +195,18 @@ ExprStr_to_time = function(format = NULL, strict = TRUE, cache = TRUE) {
...
     format = NULL,
+    ...,
     time_unit = NULL,
     time_zone = NULL,
     strict = TRUE,
     exact = TRUE,
     cache = TRUE,
     ambiguous = "raise") {
+  dots = list2(...)
+  if (length(dots) > 0) {
```
eitsupi, line 206:
```suggestion
  if (length(list2(...))) {
```

R/expr__string.R
```diff
@@ -12,14 +12,15 @@
 #' When parsing a Datetime the column precision will be inferred from the format
 #' string, if given, e.g.: `"%F %T%.3f"` => [`pl$Datetime("ms")`][pl_Datetime].
 #' If no fractional second component is found then the default is `"us"` (microsecond).
-#' @param datatype The data type to convert into. Can be either Date, Datetime,
-#' or Time.
+#' @param dtype The data type to convert into. Can be either `pl$Date`,
+#' `pl$Datetime()`, or `pl$Time`.
```
eitsupi, line 16: Please link to the document.

eitsupi (approved):
Thanks, but honestly, I feel that checking dots is overkill here.
(If we do it, wouldn't it be better to import `rlang` and use `check_dots_empty` etc to simplify things?)

eitsupi (changes requested):
Once again, on second thought, I disagree with the idea of doing a dots check process for variable name changes.
There are currently so many variable mismatches with Rust/Python Polars in this package that I don't want to take on the huge amount of work involved in writing and later removing all those checks.

It would be better to get the variable name fixes done as quickly as possible than to expend the effort on such a thing.

etiennebacher:
> There are currently so many variable mismatches with Rust/Python Polars in this package that I don't want to take on the huge amount of work involved in writing and later removing all those checks.

True but then I think we should advertise more in the README that the package is unstable and that naming all args is a good practice to reduce breaking changes.

eitsupi:
> we should advertise more in the README that the package is unstable and that naming all args is a good practice to reduce breaking changes.

I think that makes sense.
(Although I personally believe that breaking changes can happen at any time for packages that follow semantic versioning.)

R/expr__string.R
etiennebacher: Actually we only have docs for `pl$Datetime()`, not the others (`?pl_Date` links to `pl$date()`, which is not what we want here)

README.md
```diff
@@ -84,17 +88,18 @@ df$sort("fruits")$select(
...
+#> │ ---    ┆ ---    ┆ ng_fruits    ┆ --- ┆ s            ┆ its          ┆ its          ┆ by_fruits    │
+#> │ str    ┆ str    ┆ ---          ┆ i32 ┆ ---          ┆ ---          ┆ ---          ┆ ---          │
+#> │        ┆        ┆ str          ┆     ┆ i32          ┆ i32          ┆ i32          ┆ i32          │
+#> ╞════════╪════════╪══════════════╪═════╪══════════════╪══════════════╪══════════════╪══════════════╡
+#> │ apple  ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 7            ┆ 4            ┆ 4            │
+#> │ apple  ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 7            ┆ 3            ┆ 3            │
+#> │ banana ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 8            ┆ 5            ┆ 5            │
+#> │ banana ┆ audi   ┆ fruits       ┆ 11  ┆ 2            ┆ 8            ┆ 2            ┆ 2            │
+#> │ banana ┆ beetle ┆ fruits       ┆ 11  ┆ 4            ┆ 8            ┆ 1            ┆ 1            │
+#> └────────┴────────┴──────────────┴─────┴──────────────┴──────────────┴──────────────┴──────────────┘
```
eitsupi, line 102: Seems unnecessary change.

eitsupi (approved):
Thanks.

One minor comment.

README.md
etiennebacher: I had to re-knit the README because I added a sentence there, but I don't know how to avoid this change

eitsupi: This is related to the width of the terminal.

PR #949, opened 2024-03-19
fix: pass `join_nulls` and `validate` to the lazy method for `$join()`

### LINKED ISSUE #945: `$join()` with `join_nulls = TRUE` doesn't produce expected result with DataFrames
I was using the `join_nulls` argument in `$join()$`, added in #716. Joining LazyFrames with `join_nulls = TRUE` produces the expected result. However, `join_nulls` seems to be ignored when joining DataFrames. I'd be happy to look into this if it might be a good first issue?

``` r
library(polars)

df1 <-
  pl$DataFrame(
    x = "John",
    y = NULL
  )

df2 <-
  pl$DataFrame(
    x = "John",
    y = NULL,
    value = 1
  )

df1$
  join(
    other = df2,
    on = c("x", "y"),
    how = "left",
    join_nulls = TRUE
  )
#> shape: (1, 3)
#> ┌──────┬──────┬───────┐
#> │ x    ┆ y    ┆ value │
#> │ ---  ┆ ---  ┆ ---   │
#> │ str  ┆ null ┆ f64   │
#> ╞══════╪══════╪═══════╡
#> │ John ┆ null ┆ null  │
#> └──────┴──────┴───────┘

# lazy
df1$lazy()$
  join(
    other = df2$lazy(),
    on = c("x", "y"),
    how = "left",
    join_nulls = TRUE
  )$
  collect()
#> shape: (1, 3)
#> ┌──────┬──────┬───────┐
#> │ x    ┆ y    ┆ value │
#> │ ---  ┆ ---  ┆ ---   │
#> │ str  ┆ null ┆ f64   │
#> ╞══════╪══════╪═══════╡
#> │ John ┆ null ┆ 1.0   │
#> └──────┴──────┴───────┘
```

<sup>Created on 2024-03-19 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

There is a test for `join_nulls = TRUE` using LazyFrames, but not DataFrames:
https://github.com/pola-rs/r-polars/blob/0f98aa27a4f747bdaba8d21277938e10fe860004/tests/testthat/test-joins.R#L173

george-wood:
Fix #945.


etiennebacher (commented):
Thanks @george-wood, that looks good to me. Can you also add tests for the arg 
`validate` for `DataFrame`?

R/dataframe__frame.R
eitsupi, line 1: I'm wondering if we can use `do.call` instead of specifying all arguments here.
https://github.com/pola-rs/r-polars/blob/b5e0927262cb3823980b768bae5c20bb5e852190/R/io_parquet.R#L97-L100

george-wood: Good idea. I've used `do.call` instead. I also added tests for the `validate` arg.

eitsupi (changes requested):
Could you resolve conflicts of the NEWS file?

eitsupi (approved):
LGTM, thanks for the contribution!

NEWS.md
```diff
@@ -45,6 +45,11 @@
   graphviz dot syntax (#928).
 - Argument `ambiguous` can now take the value `"null"` to convert ambigous
   datetimes to null values (#937).
+
```
eitsupi, line 48:
```suggestion

```

george-wood:
> LGTM, thanks for the contribution!

Thanks for reviewing!

PR #954, opened 2024-03-20
Harmonize error message when wrong input

etiennebacher:
I think we should also review the various error types here:

https://github.com/pola-rs/r-polars/blob/6827afca398088c5cb1671d3ae4ce17a8b151379/src/rust/src/rpolarserr.rs#L9-L36

Currently we use `BadVal` for wrong inputs, but I don't understand the difference between `BadVal`, `MisValued` and `NotAChoice`


eitsupi (approved):
Thanks!
I am not sure about the structure of the error.

@Sicheng-Pan Any thoughts on this?

etiennebacher:
Let's move this discussion to #957

PR #958, opened 2024-03-21
refactor: add a new internal function `parse_as_polars_duration_string()`

eitsupi:
The generic function to make it easy to add transformations from other classes.


R/parse_as_duration.R
```diff
@@ -0,0 +1,82 @@
...
+#' - 1w (1 calendar week)
+#' - 1mo (1 calendar month)
+#' - 1q (1 calendar quarter)
+#' - 1y (1 calendar year)
+#'
+#' Or combine them: `"3d12h4m25s"` # 3 days, 12 hours, 4 minutes, and 25 seconds
+#'
+#' By "calendar day", we mean the corresponding time on the next day
+#' (which may not be 24 hours, due to daylight savings).
+#' Similarly for "calendar week", "calendar month", "calendar quarter", and "calendar year".
```
eitsupi, line 30: @etiennebacher I am thinking of making this section inherited by all other functions and writing something like "parsed by the `parse_as_polars_duration_string()`" for each params.

etiennebacher: Having this explanation in a single place and then inherit it in other docs sounds good to me, but shouldn't we keep this function internal? 

If the idea is to enable other packages to add their own method for this, then why do we implement `clock` or `vctrs` methods in `polars` and not in their original package?

etiennebacher: What I mean is that I don't understand why this function is exported. Who is expected to use this?

eitsupi: - The S3 method allows for code simplicity.
- If functions are not exported, error messages may be difficult to understand.

For example, if you run the following with 0.15.1, you will see the function name `as_pl_duration` in the error message, but if this is not exported, it will be difficult for users to debug.

```r
> pl$date_range(as.Date("2022-01-01"), as.Date("2022-03-01"), as.difftime(1:2, units = "days")) |>
      as_polars_series()
   0: In R: in pl$date_range()
   0: During function call [as_polars_series(pl$date_range(as.Date("2022-01-01"), as.Date("2022-03-01"),
          as.difftime(1:2, units = "days")))]
   1: The argument [every] caused an error
   2: internal error in as_pl_duration: did not return a string
   3: Expected a value of type [alloc::string::String]
   4: Got value [Rvalue: ["1d", "2d"], Rsexp: Strings, Rclass: ["character"]]
```

etiennebacher: Can't we simplify it like this?

```r
parse_as_polars_duration_string = function(x, ...) {
  if (length(x) != 1L || is.na(x)) {
    Err_plain("`x` should be a single non-NA value.") |>
      unwrap("in parse_as_polars_duration_string():")
  }
  if (inherits(x, "difftime")) {
    difftime_to_duration_string(x)
  } else if (inherits(x, "character")) {
    x
  }
}
```
and then call `polars:::parse_as_polars_duration_string()` on the Rust side? Here's the error I get with that:
```
> pl$date_range(as.Date("2022-01-01"), as.Date("2022-03-01"), as.difftime(1:2, units = "days"))
   0: In R: in pl$date_range():
   0: During function call [pl$date_range(as.Date("2022-01-01"), as.Date("2022-03-01"), as.difftime(1:2, 
          units = "days"))]
   1: The argument [interval] caused an error
   2: `x` should be a single non-NA value.
```
---

While we're at it, I don't see why the conversion from string to duration happens on the Rust side. Currently we pass a string to Rust and then convert it there using an R function with `R!()`. Why don't we just call `parse_as_polars_duration_string()` in R and then pass this value to Rust? It sounds simpler to me

eitsupi: > Can't we simplify it like this?

It would be simpler to use S3 generic functions, which would simplify the code and reduce bugs.
I would like to add support for clock_duration later. ( we need to support the clock package classes in as many places as possible as there is no nanosecond support in base R)

> Currently we pass a string to Rust and then convert it there using an R function with `R!()`. Why don't we just call `parse_as_polars_duration_string()` in R and then pass this value to Rust? It sounds simpler to me

I think so too. (Needless to say, I didn't write the original process for this part)

src/rust/src/utils/mod.rs
```diff
@@ -748,18 +748,15 @@ pub fn robj_to_datatype(robj: extendr_api::Robj) -> RResult<RPolarsDataType> {
     Ok(RPolarsDataType(ext_dt.0.clone()))
 }
 
+// TODO: move to R side?
```
etiennebacher, line 751: Can this be done in this PR? I think it's just a matter of removing the Rust part and call `parse_as_polars_duration_string()` in each R function with a duration input, right?

eitsupi: If we remove it, we would also need to implement on the R side the conversion to `Duration` (not `Duration` string) (which is called in the following location), which would require quite a bit of additional work.
I do not have time to do that right now.
https://github.com/pola-rs/r-polars/blob/a028047550c8bf5ad23a75d9c02f8c3825b7beb2/src/rust/src/rlib.rs#L102

eitsupi: Perhaps it would be more appropriate to convert the difftime object directly to a Duration type rather than converting it once to a String.

etiennebacher: >  I do not have time to do that right now.

Alright, let's not include it in this PR, I'll see if I have some time later for that

etiennebacher (approved):
Looks good, thanks!

Issue #854, opened 2024-02-26
Implement `pl$mean_horizontal()`


eitsupi:

etiennebacher:
@george-wood just pinging in case you want to take a stab at this one, but no pressure of course :)

george-wood:
@etiennebacher I'll have a go. Feel free to assign this to me.

PR #959, opened 2024-03-22
Implement `pl$mean_horizontal()` function

george-wood:
Proposed in #854.


george-wood:
* I reverted a few changes to unrelated Rd files.
* I also added `pl$mean_horizontal()` to the @seealso of `pl$mean()` and removed the related TODO.
* I ran `task test-snapshot-accept` following the R CMD failure.

eitsupi (approved):
Thanks!

I have updated the Rd files and test snapshots.

etiennebacher:
Nice, thank you @george-wood 👍 

There were a couple of minor things to improve in the docs, see the followup here: #960

george-wood:
@etiennebacher Good improvements. I used `NaN` and `-Inf` in the tests, but not the examples. It's good to clearly show this behavior (especially for `NaN`/`null`) in #960.

PR #966, opened 2024-03-25
refactor!: temporary deprecate positional arguments of `pl$Series()`

eitsupi:
Part of #903

As of version 0.16, positional arguments are deprecated and a warning is occurred.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,19 +18,21 @@
     - The argument `x` is renamed `values` (#933).
     - Using positional arguments in `pl$Series()` throws a warning, since the
       argument positions will be changed in the future (#966).
+
       ```r
       # polars 0.15.1 or earlier
       # The first argument is `x`, the second argument is `name`.
       pl$Series(1:3, "foo")
-      
+
       # The code above will warn in 0.16.0
       # Use named arguments to silence the warning.
       pl$Series(values = 1:3, name = "foo")
       pl$Series(name = "foo", values = 1:3)
-      
+
       # polars 0.17.0 or later (future version)
       # The first argument is `name`, the second argument is `values`.
       pl$Series("foo", 1:3)
+      ```
   - `pl$implode(...)` is rewritten to be a syntactic sugar for `pl$col(...)$implode()` (#923).
   - Unify names of input/output function arguments (#935).
     - All arguments except the first argument must be named arguments.
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,25 +15,22 @@
 
 - Several functions have been rewritten to match the behavior of Python Polars.
   - In `pl$Series()` arguments are changed.
-    - The argument converted to the values of the Series is renamed from `x` to `values` (#933).
-    - Using positional arguments in `pl$Series()` is warned, since the
+    - The argument `x` is renamed `values` (#933).
+    - Using positional arguments in `pl$Series()` throws a warning, since the
       argument positions will be changed in the future (#966).
       ```r
       # polars 0.15.1 or earlier
       # The first argument is `x`, the second argument is `name`.
       pl$Series(1:3, "foo")
-      ```
-      ```r
-      # polars 0.17.0 or later (future version)
-      # The first argument is `values`, the second argument is `name`.
-      pl$Series("foo", 1:3)
-      ```
-      To silence the warning, please use named arguments.
-      ```r
-      # polars 0.16.0 or later
+      
+      # The code above will warn in 0.16.0
+      # Use named arguments to silence the warning.
       pl$Series(values = 1:3, name = "foo")
       pl$Series(name = "foo", values = 1:3)
-      ```
+      
+      # polars 0.17.0 or later (future version)
+      # The first argument is `name`, the second argument is `values`.
+      pl$Series("foo", 1:3)
   - `pl$implode(...)` is rewritten to be a syntactic sugar for `pl$col(...)$implode()` (#923).
   - Unify names of input/output function arguments (#935).
     - All arguments except the first argument must be named arguments.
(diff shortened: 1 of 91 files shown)


etiennebacher (commented):
I'd like to clarify a bit when this warning appears and when it doesn't before reviewing the rest

R/series__series.R
```diff
@@ -265,19 +268,34 @@ Series_struct = method_as_active_binding(\() series_make_sub_ns(self, expr_struc
...
-    values,
+    ...,
+    values = NULL,
     name = NULL,
     dtype = NULL,
-    ...,
     nan_to_null = FALSE) {
   uw = function(x) unwrap(x, "in pl$Series():")
 
+  if (!missing(...)) {
```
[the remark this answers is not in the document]
eitsupi: > IIUC, `pl$Series(1:3)` should still work as before so this warning should appear only if `!missing(...) && length(...) > 1` or sth equivalent.

I don't want to implement the fallback mechanism of Python Polars for `name` to `values` here, because in Python we can guess the first argument as `name` in the following case, but in R it is not possible to determine whether they intend `name` or `values` impossible.

```python
pl.Series("foo") # <- name, not values
```

```r
pl$Series("foo") # <- name or values????
```

Also, R has `as_polas_series()`, which takes a vector as its first argument, so users can use that.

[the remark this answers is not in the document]
etiennebacher: I see, I didn't realize `pl.Series("foo")` only created a named but empty Series in Python. In that case I'm fine with warning anytime there's something in `...`.

Still I think the warning could be clearer and mention `as_polars_series()` since we want to push the usage of our `as_*` functions anyway:

```r
warning(
  "`pl$Series()` will handle unnamed arguments differently as of 0.17.0:\n",
  "- until 0.17.0, the first argument corresponds to the values and the second argument to the name of the Series.\n",
  "- as of 0.17.0, the first argument will correspond to the name and the second argument to the values.\n",
  "Use named arguments in `pl$Series()` or replace `pl$Series(<x>)` by `as_polars_series(<x>)` to silence this warning.\n"
)
```

NEWS.md
```diff
@@ -14,7 +14,26 @@
 ### Other breaking changes
 
 - Several functions have been rewritten to match the behavior of Python Polars.
-  - In `pl$Series()`, the first argument `x` is renamed to `values` (#933).
+  - In `pl$Series()` arguments are changed.
+    - The argument converted to the values of the Series is renamed from `x` to `values` (#933).
+    - Using positional arguments in `pl$Series()` is warned, since the
```
etiennebacher, line 19:
```suggestion
    - Using positional arguments in `pl$Series()` throws a warning, since the
```

NEWS.md
```diff
@@ -14,7 +14,26 @@
...
+      # polars 0.17.0 or later (future version)
+      # The first argument is `values`, the second argument is `name`.
+      pl$Series("foo", 1:3)
+      ```
+      To silence the warning, please use named arguments.
+      ```r
+      # polars 0.16.0 or later
+      pl$Series(values = 1:3, name = "foo")
+      pl$Series(name = "foo", values = 1:3)
+      ```
```
etiennebacher, line 36:
```suggestion
      ```r
      # polars 0.15.1 or earlier
      # The first argument is `x`, the second argument is `name`.
      pl$Series(1:3, "foo")
      # The code above will warn in 0.16.0
      # Use named arguments to silence the warning.
     
      # polars 0.16.0 or later
      pl$Series(values = 1:3, name = "foo")
      pl$Series(name = "foo", values = 1:3)
      
      # polars 0.17.0 or later (future version)
      # The first argument is `name`, the second argument is `values`.
      pl$Series("foo", 1:3)
```

etiennebacher: Reordered the chunks to have 0.15.0 then 0.16.0 then 0.17.0. Put everything in a single chunk

etiennebacher (approved):
Thanks, some comments. I'm also wondering if we should made the change in argument position in 0.18.0 since we mention in #903 that 

>  This would be a very huge breaking change, so we will probably have to continue warnings for several months and then make the change.

Given the rapid pace of development, I think 0.18.0 would be released in a couple of months. What I have in mind is that it's possible rust-polars 0.39.0 (and therefore r-polars 0.16.0) are released in a week or so, and 0.17.0 around the end of April. What do you think about keeping this deprecation warning for 2 full releases?

eitsupi:
Thanks for your review.

> What do you think about keeping this deprecation warning for 2 full releases?

This is a difficult question.

My concern is that the mechanism to fall back to the positional arguments is complicated, and I think it would be less confusing to remove this logic quickly.
That is, ideally `pl$(name = "foo", "bar")` should work fine in 0.15 and future versions, but this version will warning for this.
(Since there are originally three positional arguments, it is quite cumbersome to absorb all the patterns of positional and named arguments.)

So I think it would be better to replace the positional argument in 0.17 instead of 0.18 and remove this logic early to avoid the warning.

NEWS.md
```diff
@@ -14,7 +14,26 @@
...
+
+      # The code above will warn in 0.16.0
+      # Use named arguments to silence the warning.
+      pl$Series(values = 1:3, name = "foo")
+      pl$Series(name = "foo", values = 1:3)
+
+      # polars 0.17.0 or later (future version)
+      # The first argument is `name`, the second argument is `values`.
+      pl$Series("foo", 1:3)
+      ```
```
etiennebacher, line 36:
```suggestion
      ```
      This warning can also be silenced by replacing `pl$Series(<values>, <name>)` 
      by `as_polars_series(<values>, <name>)`.
```

etiennebacher (approved):

NEWS.md
etiennebacher: We should advertise `as_polars_series()` as a way to silence this warning

eitsupi:
> What I propose is to keep what we have for now, i.e make the switch in arguments in 0.17.0, but keep the possibility of extending the warning by one release if we consider that it was too short.

I think it is worth doing so too.

PR #979, opened 2024-03-28
Implement `$str$extract_groups()`

### LINKED ISSUE #975: Feature request: add `extract_groups` for str methods
Extract multiple groups at once: https://github.com/pola-rs/polars/pull/10179

etiennebacher:
Close #975 

@Yunuuuu as you can see there isn't much Rust to implement for this kind of feature. What I do in general is refer to the original PR in python-polars (in this case, this one: https://github.com/pola-rs/polars/pull/10179/files). There are a bunch of changes but we're only interested in the last 4 files which implement the documentation and make the connection between Rust and python.

Feel free to ask questions if you're interested in this


etiennebacher:
@eitsupi this failure with R-devel on windows has been here for a week or so, do you know what's happening? 

> Call `pkgbuild::check_build_tools(debug = TRUE)` to diagnose the problem.

Is it a problem related to Rtools 4.4?

eitsupi:
I believe this is separate from the R-devel problem that has been occurring recently.
My guess is that the latest GHA Windows images include R4.3.3 and Rtools44 by default, which prevents R from using Rtools.

Maybe fixed by #981

eitsupi (approved):

PR #984, opened 2024-03-29
refactor: remove `RPolarsProtoExprArray` and rewrite `<Expr>$over()`

etiennebacher:
It's called only 3 times on the R side. I think we now have other mechanisms to replace it.

This PR also introduces `mapping_strategy` for `$over()` but it's mostly because it was needed to compile. I'll add tests, examples, and a bullet point in NEWS in a followup PR.


R/expr__expr.R
```diff
@@ -1892,9 +1840,14 @@ Expr_last = use_extendr_wrapper
...
 #' )
-Expr_over = function(...) {
-  pra = construct_ProtoExprArray(...)
-  .pr$Expr$over(self, pra)
+Expr_over = function(expr, ..., mapping_strategy = "group_to_rows") {
+  expr = c(
+    wrap_elist_result(expr, str_to_lit = FALSE) |>
+      unwrap("in $over():"),
+    list2(...)
+  )
```
etiennebacher, line 1848: Not very happy about this but I couldn't find an alternative that works with all our tests.

eitsupi: I don't fully understand what this is doing, but on the docs says `...` is not use, is that perhaps a mistake?

eitsupi: I rewrote this in 828d0041072c3b1adba7ce5e3423c19488ed6fad
How about that?

```r
  list_of_exprs = list2(...) |>
    lapply(\(x) {
      if (is.character(x)) {
        as.list(x)
      } else {
        x
      }
    }) |>
    unlist(recursive = FALSE) |>
    lapply(\(x) {
      if (is.character(x)) {
        pl$col(x)
      } else {
        x
      }
    })
```

This would be the equivalent of `parse_as_list_of_expressions` in Python, so it would be worth rewriting the following sections and others using this.
https://github.com/pola-rs/r-polars/blob/0c3ce3c695bf00b389fc30dc799d4fb79a98d488/R/functions__lazy.R#L592-L595

eitsupi (approved):
Good catch, thanks!

R/expr__expr.R
etiennebacher: > How about that?

This seems good since it passes tests but it departs a bit from the [python version](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.over.html) which has `expr` + `...`. This was the reason why I had implemented it like that:
```r
  expr = c(
    wrap_elist_result(expr, str_to_lit = FALSE) |>
      unwrap("in $over():"),
    list2(...)
  )
```
Do you think that's a problem?

eitsupi: It is similar to `pl$col()` before rewriting it in #923.
In R, I think it is more natural to handle it in variable length arguments `...`.

etiennebacher: > It is similar to pl$col() before rewriting it in https://github.com/pola-rs/r-polars/pull/923.

Indeed.

Thanks for completing this PR!

PR #985, opened 2024-03-29
Implement `$str$find()`

etiennebacher:


eitsupi (approved):
Thanks!

R/expr__string.R
```diff
@@ -429,24 +429,31 @@ ExprStr_pad_start = function(width, fillchar = " ") {
...
 #' @param literal Logical. If `TRUE` (default), treat `pattern` as a literal string,
 #' not as a regular expression.
 #' @param strict Logical. If `TRUE` (default), raise an error if the underlying pattern is
 #' not a valid regex, otherwise mask out with a null value.
+#'
+#' @details To modify regular expression behaviour (such as case-sensitivity)
+#' with flags, use the inline `(?iLmsuxU)` syntax. See the regex crate’s section
+#' on [grouping and flags](https://docs.rs/regex/latest/regex/#grouping-and-flags)
+#' for additional information about the use of inline expression modifiers.
+#'
```
eitsupi, line 446: Given that the detals would be after the third paragraph from the top of the roxygen comment when written without the `@details` tag, I think this would be easier to understand if it were in its original position.

etiennebacher: Sorry, I don't understand if you want to change the position of this block in the roxygen code or in the rendered docs. 

To me, in the roxygen blocks, the most natural order is title - description - params - details - ...  which is why I moved this `@details` tag

eitsupi: I am concerned about the roxygen comment.

```r
#' @title foo
#' @description bar
#' @details baz
```

Is equivalent to...

```r
#' foo
#'
#' bar
#'
#' baz
```

So I think the `@details` tag should be at the top.
The tag is now necessary because `@description` does not currently exist, but I feel it will be confusing if description is added later.

etiennebacher: I don't think that the position of `@details` matters but I don't want to lose time over this so I just reverted this position change

PR #987, opened 2024-03-30
refactor!: rewrite `$str$count_matches()`, `$str$relace()`, `$str$replace_all()`

eitsupi:


R/expr__string.R
```diff
@@ -757,46 +748,80 @@ ExprStr_splitn = function(by, n) {
...
+#' df$with_columns(pl$col("text")$str$replace_all("a", "-"))
 #'
-#' @seealso `$str$replace()`: Replace first matching regex/literal substring.
+#' # Capture groups are supported.
+#' # Use `${1}` in the value string to refer to the first capture group in the pattern,
+#' # `${2}` to refer to the second capture group, and so on.
+#' # You can also use named capture groups.
+#' df = pl$DataFrame(word = c("hat", "hut"))
+#' df$with_columns(
+#'   positional = pl$col("word")$str$replacePall("h(.)t", "b${1}d"),
```
etiennebacher:
```suggestion
#'   positional = pl$col("word")$str$replace_all("h(.)t", "b${1}d"),
```
Causes the failures

R/expr__string.R
```diff
@@ -757,46 +748,80 @@ ExprStr_splitn = function(by, n) {
...
-#' @param pattern Regex pattern, can be an Expr.
-#' @param value Replacement, can be an Expr.
-#' @param literal Treat pattern as a literal string.
-#'
-#' @return Expr of String Series
+#' @inherit ExprStr_contains details
+#' @section Capture groups:
+#' The dollar sign (`$`) is a special character related to capture groups.
+#' To refer to a literal dollar sign, use `$$` instead or set `literal` to `TRUE`.
+#' @inheritParams ExprStr_contains
```
etiennebacher, line 762: you can remove this and add `params` in `@inherit`

tests/testthat/test-expr_string.R
```diff
@@ -604,20 +604,30 @@ test_that("str$replace", {
...
     pl$lit(c("abcabc", "123a123"))$str$replace("ab", "__")$to_r(),
     c("__cabc", "123a123")
   )
+
+  expect_identical(
+    pl$lit(c("ababab", "123a123"))$str$replace("a", "_", n = 2)$to_r(),
+    c("_b_bab", "123_123")
+  )
+
+  expect_error(
```
etiennebacher, line 627: To check but I'm not sure `expect_error()` works correctly. Can you try with a fake expected regex? Maybe `expect_grepl_error()` should be used instead

etiennebacher (approved):

tests/testthat/test-expr_string.R
eitsupi: I do not know what your concern is. I just want to confirm here that the combination of regular expression and `n>2` will not work and will result in an error.

Also it is not clear to me if `expect_grepl_error()` is better than `expect_error()`. (I don't know why `expect_grepl_error()` exists here)
As far as I know, `expect_error()` supports capturing error messages with regular expressions.

etiennebacher: > As far as I know, `expect_error()` supports capturing error messages with regular expressions.

The structure of polars errors doesn't seem to work well with regexes in `expect_error()`:

```r
# This errors (as it should)
pl$lit("?")$str$decode("base64")$to_r()

# This passes, even though "foobar" is not in the error message
expect_error(
    pl$lit("?")$str$decode("base64")$to_r(),
    "foobar"
  )

# This doesn't pass, as expected
  expect_grepl_error(
    pl$lit("?")$str$decode("base64")$to_r(),
    "foobar"
  )

Error: expected_err[which(!founds)[1]] (`actual`) not identical to `err` (`expected`).

    lines(actual) | lines(expected)                                                                    
[1] "foobar"      - "Error: Execution halted with the following contexts"                           [1]
                  - "   0: In R: in $select()"                                                      [2]
                  - "   1: Encountered the following error in Rust-Polars:"                         [3]
                  - "      \tinvalid `base64` encoding found; try setting `strict=false` to ignore" [4]
                  - ""                                                                              [5]
                  - ""                                                                              [6]

eitsupi: I see, thanks.

PR #991, opened 2024-03-30
refactor!: rewrite `<DataFrame>$get_columns()`

eitsupi:


src/rust/src/rdataframe/mod.rs
```diff
@@ -196,14 +196,9 @@ impl RPolarsDataFrame {
...
-            self.0
-                .get_columns()
-                .iter()
-                .map(|x| RPolarsSeries(x.clone())),
-        );
-        l.set_names(self.0.get_column_names()).unwrap();
-        l
+        let cols = self.0.get_columns().to_vec();
+        let vec = unsafe { std::mem::transmute::<Vec<pl::Series>, Vec<RPolarsSeries>>(cols) };
+        List::from_values(vec)
```
etiennebacher, line 201: I don't understand this change, the original was readable and didn't have any `unsafe`

eitsupi: https://github.com/pola-rs/polars/blob/345ca75baf2df276de017e2270515252b9e54a34/py-polars/src/dataframe/general.rs#L158-L161
https://github.com/pola-rs/polars/blob/345ca75baf2df276de017e2270515252b9e54a34/py-polars/src/series/mod.rs#L58-L67

etiennebacher: > This implementation is copied from Python.

I didn't see it in `get_columns()` but indeed `to_pyseries()` has that. So I'm fine with that

> What is the problem with unsafe?

I don't have a great knowledge of Rust, so I'd rather avoid `unsafe` since I have no idea in which context it's fine and in which it's not. Maybe there's no problem in this case, but I don't know that in advance.

etiennebacher (approved):

src/rust/src/rdataframe/mod.rs
eitsupi: I don't think we can completely eliminate `unsafe` in terms of accessing the C ABI of Python or R.
(As evidenced by the fact that we sometimes crash R sessions.)

PR #992, opened 2024-03-30
Implement `$item()` for `DataFrame` and `Series`

### LINKED ISSUE #976: Feature request: enhance `Series` basic operation
`polars` have supported some `Series` operation, by bring them into R, we can use `Series` as the usual R atomic vector, which would benefit a lot since we don't need use `$to_r()` method, this would be time consuming, I have tried to use polars but it's 100x slower than raw R operation when I need use `$to_r()` method and `pl$Series()` to transform between R and polars objects over 10000 times.

- Series$set: https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.set.html
- Series$scatter: https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.scatter.html (like `data.table::set` function, but not modify in place)
- Series$item: https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.item.html (instead of depend on `[` S3 method of DataFrame)
- Series$filter: https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.filter.html

And if polars can support something like `data.table::set` and `data.table::setDT` (modify data in place)?

``` r
library(polars)
bool <- pl$Series(TRUE)
bench::mark(
    to_r = lapply(seq_len(10000L), function(i) bool$to_r()),
    to_series = lapply(seq_len(10000L), function(i) pl$Series(TRUE)),
    raw_r = lapply(seq_len(10000L), function(i) TRUE),
    check = FALSE
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 3 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 to_r          1.57s    1.57s     0.638      38MB     37.6
#> 2 to_series     1.15s    1.15s     0.874     102KB     35.8
#> 3 raw_r        1.89ms   2.01ms   405.        103KB     63.8
```

<sup>Created on 2024-03-27 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>
~                                                     
~

etiennebacher:
Close #976


R/dataframe__frame.R
```diff
@@ -2225,3 +2225,64 @@ DataFrame_partition_by = function(
...
+  row_null = is.null(row)
+  col_null = is.null(column)
+
+  if (row_null && col_null) {
+    if (!identical(self$shape, c(1, 1))) {
+      Err_plain(
+        "Can only call $item() if the DataFrame is of shape (1, 1) or if explicit row/col values are provided."
+      ) |> uw()
+    }
+    return(.pr$DataFrame$select_at_idx(self, 0)$ok[1]$to_r())
```
eitsupi, line 2260: I think it would be better to use `uw()` instead of calling `$ok`.

eitsupi (approved):

Yunuuuu:
Great!

Issue #993, opened 2024-03-30
`map_elements` cause R session crach for `struct` object


Yunuuuu:
`map_elements` crash for `struct` object, Here is the `map_elements` document for struct in Python:
https://docs.pola.rs/user-guide/expressions/user-defined-functions/#combining-multiple-column-values
```r
library(polars)
pl$DataFrame(mtcars)$
    select(pl$col("mpg", "cyl")$to_struct()$map_elements(\(x) x$mpg))

pl$DataFrame(mtcars)$
    select(pl$col("mpg", "cyl")$to_struct()$map_elements(\(x) x[["mpg"]]))

pl$DataFrame(mtcars)$
    select(
    pl$col("mpg", "cyl")$to_struct()$
        map_elements(\(x) x$struct$field("mpg"))
)

```

eitsupi:
Struct type is not supported yet.

https://github.com/pola-rs/r-polars/blob/acc29eb62a44273164e2a1e66936e9f905aaa79b/src/rust/src/series.rs#L363-L390

PRs are welcome!

Yunuuuu:
I'll try to implement it.

etiennebacher:
Shorter reprex:

```r
s = pl$lit(1)$to_struct()$to_series()
s$map_elements(\(x) x)

Error in .pr$Series$map_elements(self, fun, datatype, strict_return_type,  : 
  user function panicked: map_elements
```

eitsupi:
I want removing `map_elements`.
It's much simpler to convert it back to an R vector, process it, and then convert it back to Series, and I don't see any performance difference.

etiennebacher:
`map_elements()` is no longer present in polars (R implementation). We recommend converting to a `data.frame` and use purrr or mirai to apply custom functions that use the standard R syntax.

Note that `map_batches()` still exists but we don't guarantee it works fine with the streaming engine for instance:

``` r
library(polars)

as_polars_df(mtcars)$
  select(pl$struct("mpg", "cyl")) |> 
  as.list() |> 
  purrr::map(\(x) {
    x <- x$to_r_vector()
    x$mpg + 100
  }) |> 
  as_polars_df()
#> shape: (32, 1)
#> ┌───────┐
#> │ mpg   │
#> │ ---   │
#> │ f64   │
#> ╞═══════╡
#> │ 121.0 │
#> │ 121.0 │
#> │ 122.8 │
#> │ 121.4 │
#> │ 118.7 │
#> │ …     │
[...]

PR #995, opened 2024-03-31
chore: update extendr-api to prevent crashing R session

eitsupi:
See extendr/extendr#716 (Related to #929 and #993)


etiennebacher (approved):
Thanks!

Issue #998, opened 2024-03-31
Misssing methods for `struct`: `fields` and `unnest`


Yunuuuu:
<https://github.com/pola-rs/polars/blob/main/py-polars/polars/series/struct.py>

Sorry, I don't know how to transform this rust code into `r-polars`.
https://github.com/pola-rs/polars/blob/main/py-polars/src/series/mod.rs#L77

It easily to use `$to_frame` and `$unnest` to implement `$unnest()` method in R, but I see python call rust method, I don't know if it's good idea to implement in R side only.

```r
struct_unnest = function() self$to_frame()$unnest()
```

eitsupi:
These seem fairly easy to implement, but updating the website to display the Series subnamespace in the sidebar is quite cumbersome.

Yunuuuu:
It would be great if the website can display the full series method (some methods from expr object) like python reference. when I learn `r-polars` the first time from the r-polars website, I consider `$filter` and `$slice` method (other methods displayed under `Expressions`) cannot be used for `Series`.

eitsupi:
All that is needed here is for someone to write the documentation. However, I don't know if it is worth adding more Rd files to the 600+ that already exist.

PR #1002, opened 2024-03-31
feat: `<Series>$struct$fields`

eitsupi:
Part of #998

Previously, there was no way to add Series-specific methods to the Series subnamespace, so I have added the new functionality.
This should be sufficient since there are so few of them.


eitsupi:
Note about the R CMD check error: r-lib/R6#230

etiennebacher (approved):

PR #1005, opened 2024-04-01
Refactor `expect_grepl_error()`

etiennebacher:
This PR rewrites `expect_grepl_error()` and uses it instead of `expect_error()` since the latter doesn't work well with `polars` error's structure (https://github.com/pola-rs/r-polars/pull/987#discussion_r1545263396).

We could also overwrite `expect_error()` instead of having this additional function (after all we already overwrite `expect_snapshot_file()`).


eitsupi (approved):
Thanks!

Issue #1006, opened 2024-04-01
`<Expr>$rep_extend()` is needed?


eitsupi:
I was wondering if Series was dispatching extra methods from Expr, so I was comparing it to Python[^1] and found this.
It does not exist in Rust/Python and it is not clear to me what it is for.

https://github.com/pola-rs/r-polars/blob/421d2c952576a05ccf7050200eb8ca37e1b68cc0/R/expr__expr.R#L3028-L3042

@sorhawell Please comment if you remember what this was added for and why it is needed.
<https://github.com/sorhawell/minipolars/commit/f7a260714d10df6b38292fb2557be253581d52ae>

If nothing else I will remove this.

[^1]: Get diff from `dir(polars.Series)` in Python and `ls(polars::.pr$env$RPolarsSeries)` in R

PR #1008, opened 2024-04-01
refactor: remove more useless dispatched methods from Series

eitsupi:
I listed the RPolarsSeries-specific methods that are not present in the Series of Python Polars as shown below, and checked to see if any extra methods were dispatched from Expr.

In Python:

```python
import polars as pl

l = dir(pl.Series)

with open("list.txt", "w") as f:
    for item in l:
        f.write(f"{item}\n")
```

Then, in R:

```r
> methods_py <- readr::read_lines("list.txt")
> methods_r <- ls(polars::.pr$env$RPolarsSeries)
> setdiff(methods_r, methods_py)
 [1] "add"             "and"             "approx_n_unique" "backward_fill"   "compare"        
 [6] "div"             "exclude"         "first"           "flatten"         "floor_div"      
[11] "forward_fill"    "gt_eq"           "inspect"         "last"            "lt_eq"          
[16] "map_batches"     "mod"             "mul"             "neq"             "neq_missing"    
[21] "not"             "or"              "print"           "rep"             "rep_extend"     
[26] "repeat_by"       "sort_by"         "sub"             "to_lit"          "to_r"           
[31] "to_struct"       "to_vector"       "xor"
```

As a result, several methods were determined to be unnecessary and will be removed.
(There are others that are only available in R, but Python does not dispatch Expr methods outside of the subnamespace, so it appears that not all Expr methods are available on the Series.)

See #1006 for `rep_extend`.


tests/testthat/_snaps/after-wrappers.md
```diff
@@ -203,39 +203,6 @@
...
-      [37] "shift_and_fill"          "sink_csv"               
-      [39] "sink_ipc"                "sink_json"              
-      [41] "sink_parquet"            "slice"                  
-      [43] "sort_by_exprs"           "std"                    
-      [45] "sum"                     "tail"                   
-      [47] "to_dot"                  "unique"                 
-      [49] "unnest"                  "var"                    
-      [51] "with_columns"            "with_columns_seq"       
-      [53] "with_context"            "with_row_index"         
-
```
eitsupi, line 238: This is probably due to unnecessary changes that occurred when resolving merge conflicts on some PR and were incorporated into the main branch.

etiennebacher (approved):
Thanks

PR #1010, opened 2024-04-02
Implement `$struct$unnest()`

etiennebacher:
Part of #998 

```r
pl$Series(name = "foo", values = 1)$to_struct()$struct$unnest()
shape: (1, 1)
┌─────┐
│ foo │
│ --- │
│ f64 │
╞═════╡
│ 1.0 │
└─────┘
```

@eitsupi I did as in #1002 but I'm not sure it's what you had in mind


eitsupi:
I think this is a good implementation, but the Rd file is not generated.
I think we can merge them for the time being and think about the documentation later.

I was imagining that it would be better to define the `SeriesStruct_unnest` and make it something like `unnest=SeriesStruct_unnest`

eitsupi (approved):

etiennebacher:
> the Rd file is not generated

It is generated, I just had to specify the `@name` otherwise it's named `Series_unnest`

etiennebacher:
Ah no my bad, the usage section is wrong

etiennebacher:
I'd rather find a solution here than coming back later for this.

eitsupi:
Can we merge this?

eitsupi:

Issue #919, opened 2024-03-13
Support conversion from `hms::hms()`


etiennebacher:
Good point.
`arrow` package converts Arrow `time` classes to `hms::hms`.
https://arrow.apache.org/docs/r/articles/data_types.html#translations-from-arrow-to-r
https://hms.tidyverse.org/reference/hms.html

We should support converting to `hms::hms`.

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/pull/918#discussion_r1523293866_

eitsupi:
Fixed on the default branch now.

PR #1014, opened 2024-04-06
Vignette to list some differences between R and Python API

### LINKED ISSUE #1012: Do we have a way to create `object` and `struct` with classic R functions?
I don't think we have a way to create `object` and `struct` from our standard `c()` and `list()` but maybe I'm missing something?

It would be good to have a small table in the docs to show the equivalent (if any) of those:

```python
pl.Series(values=[1])
shape: (1,)
Series: '' [i64]
[
        1
]

> pl$Series(values = 1)
polars Series: shape: (1,)
Series: '' [f64]
[
	1.0
]
```

```python
>>> pl.Series(values=[[1]])
shape: (1,)
Series: '' [list[i64]]
[
        [1]
]

> pl$Series(values = list(1))
polars Series: shape: (1,)
Series: '' [list[f64]]
[
	[1.0]
]
```

```python
>>> pl.Series(values=[{1}])
shape: (1,)
Series: '' [o][object]
[
        {1}
]

???
```

```python
>>> pl.Series(values=[{"a": 1}])
shape: (1,)
Series: '' [struct[1]]
[
        {1}
]

???
```

etiennebacher:
Close #1012

We can add more things later if necessary.


vignettes/differences-with-python.Rmd
```diff
@@ -0,0 +1,146 @@
...
+
+## Structs and objects
+
+### Structs
+
+A `Struct` is a data type that is composed of several `Field`s, which themselves
+have a name and a datatype.
+
+While it is possible to create a `Series` of type `Struct` in Python via
+`pl.Series()`, this is not possible in R:
```
eitsupi, line 72: See https://github.com/pola-rs/r-polars/issues/1012#issuecomment-2047399903

eitsupi (changes requested):

vignettes/differences-with-python.Rmd
```diff
@@ -0,0 +1,160 @@
...
+x = pl$DataFrame(
+  a = 1:2,
+  b = list(c("x", "y"), "z")
+)
+
+out = x$select(pl$col("a", "b")$to_struct())
+out
+
+out$schema
+```
```
eitsupi, line 116: I don't see this as something that should be highlighted as a difference from the Python API, just that the data.frame in R is converted to a Struct in Polars?

etiennebacher: Yes, in the end this can be stored in the docs of `DataType_Struct` instead

vignettes/differences-with-python.Rmd
```diff
@@ -0,0 +1,106 @@
...
+> pl$DataFrame(x = hms::hms(56, 34, 12))
+shape: (1, 1)
+┌─────────┐
+│ x       │
+│ ---     │
+│ f64     │
+╞═════════╡
+│ 45296.0 │
+└─────────┘
+```
```
eitsupi, line 87: Could this be in reprex style so that it can be copied and pasted and executed?

vignettes/differences-with-python.Rmd
```diff
@@ -0,0 +1,106 @@
...
+[1] <LINESTRING (0 1, 3 9)>
+
+> pl$DataFrame(x = geos::as_geos_geometry("LINESTRING (0 1, 3 9)"))
+Error: Execution halted with the following contexts
+   0: In R: in $DataFrame():
+   0: During function call [pl$DataFrame(x = geos::as_geos_geometry("LINESTRING (0 1, 3 9)"))]
+   1: When constructing polars literal from Robj
+   2: Encountered the following error in Rust-Polars:
+      	expected Series
+```
```
eitsupi, line 105: Ditto. (Maybe with `tryCatch`)

eitsupi, line 105: Ditto. (Maybe with `tryCatch`)

vignettes/differences-with-python.Rmd
etiennebacher: I added the reprex style but those chunks are not evaluated every time (that would require us to put `hms` and `geos` in Suggests)

eitsupi: Of course I don't think this needs to be an executable code block for now, but I think `hms` will need to be included in Suggests at some point. (#919)

eitsupi:
Maybe altdoc file(s) should be updated?
https://github.com/pola-rs/r-polars/blob/de1fd05ebe7e22469b10bd8451d94c67a0b8e1f3/altdoc/mkdocs_static.yml#L53

Issue #425, opened 2023-10-15
Mapping table of R types and Polars types


eitsupi:
Like this:
https://arrow.apache.org/docs/11.0/r/articles/data_types.html#list-of-default-translations

eitsupi:
Particular attention may need to be paid to datetime conversions. (#578)

eitsupi:
https://github.com/pola-rs/r-polars/blob/750f3246338d81d8a5b0db0dde95793dd03a3a97/src/rust/src/lazy/dsl.rs#L82-L182

PR #1015, opened 2024-04-06
refactor: swap `pl$Series` and `as_polars_series.default`

eitsupi:
Related to #1012.

`as_polars_series.default()` used to call `pl$Series()`, but will be changed so that `pl$Series()` calls `as_polars_series()`.
This will ensure that every object is properly converted to a Series.

Python's `polars.Series.__init__()` internally converts various objects into a Series through a large number of conditional branches. This update is very similar to that.

### DIFF
--- a/R/as_polars.R
+++ b/R/as_polars.R
@@ -303,11 +303,12 @@ as_polars_lf.RPolarsLazyGroupBy = function(x, ...) {
 
 #' To polars Series
 #'
-#' [as_polars_series()] is a generic function that converts an R object to a
-#' polars Series. It is basically a wrapper for [pl$Series()][pl_Series].
-#' @param x Object to convert into a polars Series
-#' @param name A string to use as the name of the Series.
-#' If `NULL` (default), the name of `x` is used or an unnamed Series is created.
+#' [as_polars_series()] is a generic function that converts an R object to
+#' [a polars Series][Series_class].
+#' @param x Object to convert into [a polars Series][Series_class].
+#' @param name A character to use as the name of the [Series][Series_class].
+#' If `NULL` (default), the name of `x` is used or an empty character `""`
+#' will be used if `x` has no name.
 #' @inheritParams as_polars_df
 #' @return a [Series][Series_class]
 #' @export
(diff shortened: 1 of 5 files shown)


R/as_polars.R
```diff
@@ -367,13 +369,15 @@ as_polars_series.POSIXlt = function(x, name = NULL, ...) {
 #' @rdname as_polars_series
 #' @export
 as_polars_series.data.frame = function(x, name = NULL, ...) {
-  pl$DataFrame(unclass(x))$to_struct(name = name)
+  as_polars_df(x)$to_struct(name = name)
```
eitsupi, line 372: This change is not directly related, but is a small refactoring to make object conversions more appropriate.

R/as_polars.R
```diff
@@ -367,13 +369,15 @@ as_polars_series.POSIXlt = function(x, name = NULL, ...) {
...
 }
 
 
 #' @rdname as_polars_series
 #' @export
-as_polars_series.vctrs_rcrd = as_polars_series.data.frame
+as_polars_series.vctrs_rcrd = function(x, name = NULL, ...) {
+  pl$select(unclass(x))$to_struct(name = name)
+}
```
eitsupi, line 380: Ditto.

R/as_polars.R
```diff
@@ -303,11 +303,12 @@ as_polars_lf.RPolarsLazyGroupBy = function(x, ...) {
 
 #' To polars Series
 #'
-#' [as_polars_series()] is a generic function that converts an R object to a
-#' polars Series. It is basically a wrapper for [pl$Series()][pl_Series].
-#' @param x Object to convert into a polars Series
-#' @param name A string to use as the name of the Series.
-#' If `NULL` (default), the name of `x` is used or an unnamed Series is created.
+#' [as_polars_series()] is a generic function that converts an R object to
+#' [a polars Series][Series_class].
```
etiennebacher, line 307: Should we have a list of R objects that can be converted to Series?

etiennebacher (approved):

R/as_polars.R
eitsupi: Definitely needed... See #425.
The documentation for the `arrow` package is helpful.

etiennebacher: Ok, we can address #425 in full in another PR then

PR #1021, opened 2024-04-10
fix!: convert R list to Series correctly via `as_polars_series()`

eitsupi:
Fix #1020

Ideally, this would require significant refactoring on the Rust side, but for now it will work correctly.

PR #1022, opened 2024-04-11
fix!: fix list type column conversion from data frame

eitsupi:
A follow up for #1021

I was trying more complex nested type cases and learned that the columns of list class of data.frame are `AsIs` class and the list class is hidden.

``` r
data.frame(a = I(list(data.frame(b = 1L))))$a |> class()
#> [1] "AsIs"
```

<sup>Created on 2024-04-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

So it was necessary to add a process here to convert this column back to the list class to convert it to the Series as intended.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@
 ### Other breaking changes
 
 - R objects inside an R list are now converted to Polars data types via
-  `as_polars_series()` (#1021). For example, up to polars 0.15.1,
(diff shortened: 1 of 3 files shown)


etiennebacher (approved):
Thanks, not sure if you want to merge this or #1023 first so I'll let you do that

PR #1023, opened 2024-04-11
fix: nested null data type Series treated as an empty Series

eitsupi:
A follow up for #1021

This change will allow for more proper conversion from lists that contain NULLs or empty lists.

For example, in 0.15.1, a strange error occurs in the following case:

```r
> as_polars_series(list(list(), list(1, 2)))
   0: In R: in pl$Series():
   0: During function call [as_polars_series(list(list(), list(1, 2)))]
        data types don't match: When building series from R list; some parsed sub-elements did not match: One element was f64 and another was list[f64]
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@
 ### Other breaking changes
 
 - R objects inside an R list are now converted to Polars data types via
-  `as_polars_series()` (#1021, #1022). For example, up to polars 0.15.1,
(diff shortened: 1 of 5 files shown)


eitsupi:
Updating the NEWS file will cause conflicts with #1022, so I want to update later.

etiennebacher (approved):
LGTM, thanks

PR #1027, opened 2024-04-12
docs(news): polish news

eitsupi:
- The section "Breaking changes due to Rust-polars update" was removed as I felt it was not really a breaking change or should not be classified as Python API like changes.
- Moved the position of what should be classified as changes to mimic the Python API.

(The release notes for this release are so long that it would be easier to read if they were further categorized, but I don't want to put too much effort into that considering that the changelog is automatically generated from the PR titles in the upstream polars repository as well. Of course, it would be great if someone would work on it.)

PR #1029, opened 2024-04-13
Reformat changelog

etiennebacher:
Followup of #1027

Mainly two things:

* split breaking changes related to matching the py-polars in API in 4 categories. @eitsupi I'm not sure that's what you meant by "it would be easier to read if they were further categorized" in #1027 but I think it makes sense
* split new features by type of feature (function, method, etc.)

Let me know if that's fine for you or if you prefer #1027 

(PS: the bullet numbers must be 1, 2, 3 and 4 and not a, b, c, d. `mkdocs` automatically convert the 1-4 to a-d but using letters directly doesn't work)


eitsupi (approved):
Looks good, thanks!

PR #1030, opened 2024-04-13
Trigger documentation CI on changes in NEWS

etiennebacher:


.github/workflows/docs.yaml
```diff
@@ -20,6 +20,7 @@ on:
       - vignettes/**
       - DESCRIPTION
       - README.md
+      - NEWS.md
```
eitsupi, line 23: Please add this to `pull_request` too.

eitsupi (changes requested):

eitsupi (approved):

Issue #952, opened 2024-03-20
0.16.0 release


eitsupi:
The main branch currently contains a large number of breaking changes to bring its behavior consistent with that of Python Polars.
On the other hand, there are very few changes on the Rust side.

If there are no further comments, I would like to delay the release of 0.16.0 until Rust 0.39.x is released, and do further breaking refactoring before then.

etiennebacher:
I'm fine with waiting for Rust 0.39.0. I think it's better to pack a lot of breaking changes to align with py-polars in one or two releases than splitting them across many releases.

eitsupi:
> I think it's better to pack a lot of breaking changes to align with py-polars in one or two releases than splitting them across many releases.

I agree, but it's too huge and tedious work...
I work here as a hobby, so I don't have the motivation to concentrate on boring tasks.

eitsupi:
The release notes have become too long and I will immediately release 0.16.0 when the next release of Rust Polars is made.

eitsupi:
It looks like there will be no Rust Polars release this week. (Since Python has been released)

etiennebacher:
I don't think there's a link between the two, rust-polars 0.38 was released at the same time as py-polars 0.20.13.

eitsupi:
Hmmm, looks like it was just released. #1034

etiennebacher:
#1034 is now merged. We could also include #1033 in 0.16. I'll finish it tomorrow if you don't do it first

eitsupi:
Thanks, I have a couple of things I'm considering doing to follow up for #1034.

- Rename `$str$parse_int()` to `$str$to_integer()`.
- Note pola-rs/polars#15646 as a known issue to the NEWS.

eitsupi:
I believe the changes on the Rust side are complete, so I have triggered a library release.

eitsupi:
0.16.0 released in #1041

PR #1037, opened 2024-04-15
Rename `$str$parse_int()` to `$str$to_integer()`

etiennebacher:
Mentioned in #952


eitsupi:
Thank you.
However, upon closer inspection of the Python documentation, I realized that the problem is not just a name change.
Sorry for the duplication of work, but please see #1038.

etiennebacher:
Nvm I see you updated #1038

PR #1038, opened 2024-04-15
refactor!: `$str$parse_int()` -> `$str$to_integer()`

eitsupi:


NEWS.md
```diff
@@ -84,6 +85,8 @@
       the `literal` argument must be named (#987).
     - In `$str$strptime()`, `$str$to_date()`, `$str$to_datetime()`, and
       `$str$to_time()`, all arguments (except the first one) must be named (#939).
+    - In `$str$to_integer()` (renamed from `$str$parse_int()`), all arguments
+      should be named (#1038).
```
etiennebacher, line 89: > 
```suggestion
      must be named (#1038).
```

NEWS.md
```diff
@@ -136,7 +139,8 @@
       only accepts `"auto"`, `"columns"`, `"row_groups"`, and `"none"`.
       Previously, it also accepted upper-case notation of `"auto"`, `"columns"`,
       `"none"`, and `"RowGroups"` instead of `"row_groups"` (#1033).
-
+    - In `$str$to_integer()` (renamed from `$str$parse_int()`), the default
+      value of `base` is changed from `2` to `10L` (#1038).
```
etiennebacher, line 143: > 
```suggestion
      value of `base` is changed from `2` to `10` (#1038).
```

NEWS.md
```diff
@@ -154,6 +158,7 @@
 - The following deprecated functions are now removed: `pl$threadpool_size()`,
   `<DataFrame>$with_row_count()`, `<LazyFrame>$with_row_count()` (#965).
 - In `$group_by_dynamic()`, the first datapoint is always preserved (#1034).
+- `$str$parse_int()` is renamed to `$str$to_integer()` (and completely rewritten) (#1038).
```
etiennebacher, line 161: ``
```suggestion
- `$str$parse_int()` is renamed to `$str$to_integer()` (#1038).
```

R/expr__string.R
```diff
@@ -862,29 +862,31 @@ ExprStr_slice = function(offset, length = NULL) {
...
-#' @param base Positive integer which is the base of the string we are parsing.
-#' Default is 2.
-#' @param strict If `TRUE` (default), integer overflow will raise an error.
-#' Otherwise, they will be converted to `null`.
-#' @return Expr: Series of dtype i32.
+#' @param ... Ignored.
+#' @param base A positive integer or expression which is the base of the string
+#' we are parsing. Characters are parsed as column names. Default: `10L`.
+#' @param strict A logical. If `TRUE` (default), raise any ParseError or overflow
+#' as ComputeError. If `FALSE`, silently convert to `null`.
```
etiennebacher, line 875:
```suggestion
#' @param strict If `TRUE` (default), parsing errors or integer overflow will
#' raise an error. Otherwise, they will be converted to `null`.
```

man/ExprStr_len_bytes.Rd
```diff
@@ -2,7 +2,7 @@
 % Please edit documentation in R/expr__string.R
 \name{ExprStr_len_bytes}
 \alias{ExprStr_len_bytes}
-\title{Get the number of bytes in strings}
+\title{Get the number of bytes in $strings}
```
etiennebacher, line 5: Not sure where this comes from?

man/ExprStr_len_chars.Rd
```diff
@@ -2,7 +2,7 @@
 % Please edit documentation in R/expr__string.R
 \name{ExprStr_len_chars}
 \alias{ExprStr_len_chars}
-\title{Get the number of characters in strings}
+\title{Get the number of characters in $strings}
```
etiennebacher, line 5: same

src/rust/src/lazy/dsl.rs
```diff
@@ -2307,14 +2307,15 @@ impl RPolarsExpr {
...
+        let base = robj_to!(PLExprCol, base)?;
+        let strict = robj_to!(bool, strict)?;
         Ok(self
             .0
             .clone()
             .str()
-            .to_integer(robj_to!(PLExprCol, base)?, robj_to!(bool, strict)?)
-            .with_fmt("str.parse_int")
+            .to_integer(base, strict)
+            .with_fmt("str.to_integer")
```
etiennebacher, line 2318: I think this can be removed (especially since it's the python syntax)

tests/testthat/test-expr_string.R
```diff
@@ -685,28 +685,28 @@ test_that("str$str_explode", {
...
+test_that("str$to_integer", {
   expect_identical(
-    pl$lit(c("110", "101", "010"))$str$parse_int(2)$to_r(),
+    pl$lit(c("110", "101", "010"))$str$to_integer(2)$to_r(),
     c(6, 5, 2)
   )
 
   expect_identical(
-    pl$lit(c("110", "101", "010"))$str$parse_int()$to_r(),
+    pl$lit(c("110", "101", "010"))$str$to_integer()$to_r(),
```
etiennebacher, line 695: isn't this wrong since the default base changed?

etiennebacher (approved):

R/expr__string.R
eitsupi: I think it is actually only for `FALSE`, not otherwise. (This is a subtle nuance difference, such as when strictly logical may take `NA`, as opposed to when `isTRUE()` is used to determine it.......)

tests/testthat/test-expr_string.R
eitsupi: Yeah, I forgot to update the test.

src/rust/src/lazy/dsl.rs
eitsupi: I think it might be the case too, but I cannot be sure because there are a large number of the same items in other places.

R/expr__string.R
eitsupi: In addition, this is copied from the Python documentation.

etiennebacher: Ok to replace "otherwise", but I think we should remove the ComputeError and ParseError since they don't mean much to R users

PR #1040, opened 2024-04-15
docs(news): tweak NEWS.md

eitsupi:


eitsupi:
Hmm Markdownlint and Mkdocs seem to be incompatible.......

etiennebacher (approved):
Can you check whether this is fine on the rendered website?

eitsupi:
> Can you check whether this is fine on the rendered website?

Yes, of course I checked locally.

image

As an aside, it would be useful to have an option to render only the NEWS.md file and skip the execution of the man file.
pkgdown have functions to render only certain files, I believe.

etiennebacher:
For now we only have `freeze = TRUE` to skip files that didn't change

eitsupi:
> For now we only have `freeze = TRUE` to skip files that didn't change

Yes, I have used it and seen the results. However, right after deleting the generated files, we need to wait until all man pages are rendered.

```r
altdoc::render_docs(freeze = TRUE, parallel = TRUE, verbose = TRUE)
```

PR #1042, opened 2024-04-16
fix: bump rust polars to 0.39.1

eitsupi:


tests/testthat/test-expr_array.R
```diff
@@ -17,38 +17,38 @@ test_that("arr$sum", {
...
-#   expect_identical(
-#     df$select(pl$col("ints")$arr$max())$to_list(),
-#     list(ints = c(2L, 1L, NA_integer_))
-#   )
-#   expect_identical(
-#     df$select(pl$col("floats")$arr$max())$to_list(),
-#     list(floats = c(2, 1, NA_real_))
-#   )
+test_that("arr$max and arr$min", {
+  skip_if_not(polars_info()$features$nightly)
```
etiennebacher, line 22: I think we had this flag because it used to panick but I don't think it's needed anymore?

etiennebacher (approved):
Thanks, I'm fine with releasing 0.16.1 right away since this patch release fixes an important perf regression

tests/testthat/test-expr_array.R
eitsupi: Good point.
I think this could have been a bug of rustc (https://github.com/pola-rs/polars/issues/14359#issuecomment-1936910731), but it is worth seeing if it works with the current latest release.

eitsupi: It seems working.

PR #1051, opened 2024-04-19
chore: bump rust polars to 0.39.2

eitsupi:


etiennebacher (approved):
LGTM but I don't think we should release 0.16.2 right away. Let's wait one week so that we can include other patch versions of rust-polars if there are any?

eitsupi:
> LGTM but I don't think we should release 0.16.2 right away. Let's wait one week so that we can include other patch versions of rust-polars if there are any?

Yeah, I don't want to do the release this week because I too am tired of having to do the release process over and over.

PR #1050, opened 2024-04-18
Rename argument `drop_nulls` to `ignore_nulls` in `$all()` and `$any()`

etiennebacher:
Also improve docs and add tests

PR #1053, opened 2024-04-22
Prevent unnamed non-`Field` elements in `pl$Struct()`

### LINKED ISSUE #1011: `pl$Struct()` must have named elements
```r
> pl$Struct(pl$Float32)
DataType: Struct(
    [
        Field {
            name: "",
            dtype: Float32,
        },
    ],
)
```

```python
>>> pl.Struct(pl.Float32)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python311\Lib\site-packages\polars\datatypes\classes.py", line 837, in __init__
    self.fields = list(fields)
                  ^^^^^^^^^^^^
TypeError: 'DataTypeClass' object is not iterable
```

etiennebacher:
Close #1011

Breaking so should be in 0.17.0


etiennebacher:
@eitsupi do you think the next release will be 0.16.4 or 0.17.0? Just wondering since I have a couple of PR that can be merged if the next one is 0.17

eitsupi:
I am also worried about this.......
The next release of Rust Polars is going to be pretty huge, and considering that Python was not even released last week, I think we may have trouble incorporating Rust Polars' breaking changes here.
So I think it would be better to merge in potentially breaking changes like this and #1078, but probably rarely used, and release 0.16.4. (If we follow semantic versioning, all releases prior to 1.0.0 may contain braking changes, so it is acceptable for 0.16.4 to contain breaking changes)

etiennebacher:
I don't think the next rust-polars release will be a big issue (I mean I don't see why it would be more problematic than the previous ones). Also, this PR and #1050 shouldn't have many conflicts with the next release of rust-polars so I'm fine with waiting the rust-polars update before merging those.

eitsupi:
Since #1078 is no longer a breaking change, I would like to merge #1078 and release 0.16.4.

etiennebacher:
Ah sorry, I understood that you would like to merge this PR because of your comment but I just misread, back to draft

eitsupi (approved):

PR #1070, opened 2024-05-03
release: 0.16.3

eitsupi:
I would like to create a release to reflect the website URL change to R-universe.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,8 @@
 
 - Move the API document website to the new place (#1067, #1068).
   Access to the old website is set to redirect to the top page of the new website.
+  - Old URL: `https://rpolars.github.io/`
+  - New URL: `https://pola-rs.github.io/r-polars/`
 
 ## Polars R Package 0.16.2
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@
 ### Miscellaneous
 
 - Move the API document website to the new place (#1067, #1068).
+  Access to the old website is set to redirect to the top page of the new website.
 
 ## Polars R Package 0.16.2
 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,6 +1,6 @@
 # NEWS
 
-## Polars R Package (development version)
+## Polars R Package 0.16.3
```
etiennebacher, line 3: Can you mention the new URL in the news and that old links are redirected?

etiennebacher (approved):

NEWS.md
eitsupi: Thanks for your quick review.
How about b768b3d37fbad070a41d337937b1da86ef300fab?

eitsupi: By the way, I wanted to release this earlier because the automatic link by downlit was broken as shown below. (The redirect is to the top page, so downlit can no longer find pkgdown.yml)

https://github.com/rpolars/r-polarssql/commit/c4c6478a1d64cfd4f3ff003815476859a84bea34#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R119

etiennebacher: Thanks, can you just mention the old and new root URLs in the news?

eitsupi: How about c816cd6194be24e5124256b88498b5d4a72d2862?

I did not do this because I thought it might cause an error if the URL check by R CMD check detected a redirect, but perhaps it does not have to be an actual link here.

etiennebacher: Ah i didn't think about that, let's remove it if R CMD check errors. Thanks!

eitsupi: It passed the tests.

PR #1073, opened 2024-05-04
feat: serialize/deserialize LazyFrame

eitsupi:
The Python class method `polars.LazyFrame.deserialize()` cannot be expressed well here, so I assigned a function `pl$deserialize_lf()`.
The function name could be change.

In addition, in Python, these functions can also have files as input/output destinations, but here I am simplifying to interacting with character vectors. Reading and writing to files must be done by other functions.


eitsupi:
It might be worthwhile to enable the format to be specified as chracter or raw, as in nodejs.
https://pola-rs.github.io/nodejs-polars/interfaces/pl.LazyDataFrame.html#serialize

etiennebacher (approved):
Thanks, I don't know much about serializing/deserializing but LGTM. Should we add a comment about security concerns when deserializing an unknown input as they do [in the py-polars docs](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.deserialize.html#polars.LazyFrame.deserialize)?

eitsupi:
> Should we add a comment about security concerns when deserializing an unknown input as they do [in the py-polars docs](https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.deserialize.html#polars.LazyFrame.deserialize)?

I think there is a special feature for Python that saves Python objects as pickle for Python.
I don't think the concern exists because there is no R-only functionality here and Polars does not restore anything like rds files.

PR #1093, opened 2024-05-15
refactor: several updates to `dsl.rs`

etiennebacher:
No changes visible by user (except for error messages):

* Replace `r_result_list` by `RResult<RPolarsExpr>`
* Remove calls to `.map()` when they are unnecessary
* Update the message when a value is not accepted. 

Old:
``` r
library(polars)

df = pl$DataFrame(
  foo = 1:3,
  bar = c(6, 7, 8),
  ham = c("a", "b", "c")
)

other_df = pl$DataFrame(
  apple = c("x", "y", "z"),
  ham = c("a", "b", "d")
)

df$join(other_df, on = "ham", how = "foobar")
#>    0: In R: in $join():
#>    1: The argument [how] caused an error
#>    2: Got value [JoinType choice ('foobar') must be one of 'cross', 'inner', 'left', 'outer', 'semi', 'anti']
```
New:
```r
   0: In R: in $join():
   1: The argument [how] caused an error
   2: Invalid choice: JoinType ('foobar') must be one of 'cross', 'inner', 'left', 'outer', 'semi', 'anti'
```


eitsupi (approved):
Thanks!

PR #1097, opened 2024-05-18
refactor: move conversion to string-duration fully to R

etiennebacher:
Currently, parsing duration is done like that:

* pass the string from R to Rust
* eval `parse_as_polars_duration_string()` from Rust using the `R!` macro
* use this result in Rust

Since `parse_as_polars_duration_string()` is already defined in R, we could just call this on the string and then pass the result to Rust.

Also added a couple of small refactor and `clippy` suggestions.


eitsupi (approved):
Thanks!

PR #1098, opened 2024-05-18
refactor: always use `Self` or `RResult<Self>` in Rust `impl` for `DataFrame`, `LazyFrame` and `Expr`

etiennebacher:


eitsupi (approved):
Thanks!

PR #1104, opened 2024-05-21
Bump rust-polars to 0.40.0

etiennebacher:
https://github.com/pola-rs/polars/releases/tag/rs-0.40.0

@eitsupi I'll have limited time in the next few days to address this (we are not in rush, it's just so you know)


eitsupi:
> @eitsupi I'll have limited time in the next few days to address this (we are not in rush, it's just so you know)

Thanks for starting the work. Of course I'll take a look at this, but it looks like a pretty tough update and I don't know if I can complete it in a few days.

eitsupi:
It can now be compiled, but the rolling functions will probably need to be rewritten.

R/io_ipc.R
```diff
@@ -32,7 +29,6 @@ pl_scan_ipc = function(
     source,
     ...,
     n_rows = NULL,
-    memory_map = TRUE,
```
eitsupi, line 35: Is this change needed?

etiennebacher: I think so because of https://github.com/pola-rs/r-polars/pull/1104/commits/1eb42452e6740e67b58352ab1849be640fba4f02

eitsupi: Oh, I see. Perhaps it was only removed from `.pr$DataFrame$from_raw_ipc`, so there should be no need to remove it from anywhere else.
I suppose that this is still valid when reading from files.

etiennebacher:
I added support for `rolling_*_by()` on the Rust side in my last commit but I think we can implement the R side in a separate PR so that this one can be purely on fixing compilation errors and test failures (since I already pushed the changes, we might as well leave them here)

R/io_ipc.R
etiennebacher: I'm confused, I don't know if `memory_map` should be removed from other places functions and from all tests where it is set. I also don't really understand what this argument does, so can you either deal with this or let me know if it's fine to remove it from the `scan/read_ipc` functions?

eitsupi: This only makes sense when reading from an Arrow file. My understanding is that it is an option to determine whether or not to keep a reference to the file without copying the data, and disabling it will degrade performance.

``` r
polars::pl$DataFrame(a = 1)$write_ipc("test.arrow")

df <- polars::pl$read_ipc("test.arrow")
df2 <- polars::pl$read_ipc("test.arrow", memory_map = FALSE)

df
#> shape: (1, 1)
#> ┌─────┐
#> │ a   │
#> │ --- │
#> │ f64 │
#> ╞═════╡
#> │ 1.0 │
#> └─────┘
```

``` r
df2
#> shape: (1, 1)
#> ┌─────┐
#> │ a   │
#> │ --- │
#> │ f64 │
#> ╞═════╡
#> │ 1.0 │
#> └─────┘
```

``` r

polars::pl$DataFrame(a = 2)$write_ipc("test.arrow")

df
#> shape: (1, 1)
#> ┌─────┐
#> │ a   │
#> │ --- │
#> │ f64 │
#> ╞═════╡
#> │ 2.0 │
#> └─────┘
```

``` r
df2
#> shape: (1, 1)
#> ┌─────┐
#> │ a   │
#> │ --- │
#> │ f64 │
#> ╞═════╡
#> │ 1.0 │
#> └─────┘
```

<sup>Created on 2024-05-26 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

eitsupi:
@etiennebacher Thanks for your great works!

On my machine, the R session hangs on the expr_name test (so I have renamed the test file to disable the tests for now) and the following part seems to cause a panic.
https://github.com/pola-rs/r-polars/blob/3d2a33361e6765294e3e44201b92429c28b67402/tests/testthat/test-series.R#L508

I think the hang in the R session may be due to the implementation of Send and Sync at the following location, but I do not know why this has now occurred.
I also do not know why @sorhawell implemented this.
https://github.com/pola-rs/r-polars/blob/3d2a33361e6765294e3e44201b92429c28b67402/src/rust/src/utils/extendr_concurrent.rs#L11-L18

eitsupi:
Hmmm, it looks like there is a problem with the way the DataFrame is constructed.
It panics when I try to create a DataFrame with columns of nested List types, which is not a problem in Python.

```r
> pl$DataFrame(a = as_polars_series(list(list(1, 2, 3), list(4, 5, 6))))
thread '<unnamed>' panicked at /usr/local/cargo/git/checkouts/polars-b0d90607192fd414/7bc7014/crates/polars-core/src/series/mod.rs:149:37:
called `Result::unwrap()` on an `Err` value: InvalidOperation(ErrString("`vec_hash` operation not supported for dtype `list[f64]`"))
thread '<unnamed>' panicked at src/rdataframe/mod.rs:89:1:
explicit panic
   0: In R: in $DataFrame():
   0: During function call [pl$DataFrame(a = as_polars_series(list(list(1, 2, 3), list(4, 
          5, 6))))]
   1: user function panicked: select
```

eitsupi (dismissed):
There are several unresolved issues with this PR, but I think we can merge and do the follow-up work for now.

eitsupi:
map method errors may be related to pola-rs/polars#16365
I thought I had made modifications, but they may not have been sufficient.

eitsupi:
More related issue: pola-rs/polars#12568

eitsupi:
I do not know why `$name$map()` does not work only on Ubuntu.

eitsupi:
I don't understand why this process has to be done on the Rust side.
Wouldn't it be enough to receive the function as an argument in rename, process it on the R side, and then pass it as a list to Rust's rename?
https://github.com/pola-rs/polars/blob/17cf8b7e17974b51c9bdc760422005fe591bc7c5/py-polars/polars/lazyframe/frame.py#L4393-L4445

In any case, I think we can merge them as they are now and fix them later. (I don't think I will have enough time for a while down the road)

eitsupi:
This is a side note, but what I have learned is that upstream is not very interested in Rust's breaking changes, and the release notes are of little use.

So in the future, I think it is better to proactively incorporate changes from HEAD without waiting for Rust releases.

etiennebacher:
Thanks a lot, I'll take a look at the remaining issues to see if I can come up with a fix for them. In any case, I'll update the NEWS before merging.

> This is a side note, but what I have learned is that upstream is not very interested in Rust's breaking changes, and the release notes are of little use.

The release notes are sometimes useful. They are messy but it can really help going through bugs or changes in implementation. I agree we shouldn't feel forced to wait for an official release but I also think it matches quite well the pace of development (release every ~1.5 month).

There were several trivial bug fixes upstream just after 0.40.0 were released, I think we should bump the version to an unreleased version but we should do it in another PR.

NEWS.md
```diff
@@ -4,6 +4,21 @@
...
+  - `$top_k()` and `$bottom_k()` gain three arguments `nulls_last`,
+    `maintain_order` and `multithreaded`.
+  - All `$rolling_*()` functions lose the arguments `by`, `closed` and
+    `warn_if_unsorted`. Rolling computations based on `by` must be made via the
+    corresponding `rolling_*_by()`, e.g `rolling_mean_by()` instead of
+    `rolling_mean(by =)`.
+  - `pl$scan_parquet()` and `pl$read_parquet()` gain an argument `glob` which
+    defaults to `TRUE`. Set it to `FALSE` to avoid considering `*` as a globing
+    pattern.
+  - `pl$read_ipc()` loses the argument `memory_map`.
```
eitsupi, line 18: Again, I think this is incorrect. But it can be fixed later (no need to wait a large amount of compile time again for a single line of md file)

eitsupi (approved):

eitsupi:
Follow-up tasks should include the followings:

- Add `*_by()` methods.
- Fix `pl$select()` for nested list types.
- Remove `$name$map()` and add a functionality rename with R function to `$rename()` (or adds `$rename_with()`)

NEWS.md
etiennebacher: I can replace by "When reading from an Arrow file with `pl$scan_ipc()` or `pl$read_ipc()`, the argument `memory_map` is no longer used internally."

If you agree I can add that after the CI passes and merge directly

eitsupi: No it is used. I have reverted the changes you made.
The following comment is generated by installing the latest commit at that point in time.
https://github.com/pola-rs/r-polars/pull/1104#discussion_r1615002017

The only problem is that there is probably no test at this time to confirm this behavior.

etiennebacher: Sorry if I miss something but this argument was removed here: https://github.com/pola-rs/r-polars/pull/1104/files#diff-02c4f6c445f96e42a1dcbf81cc271224cd2f115936c291f58990e2c0fc45ba66

So either it's a mistake and it should be put back or it's voluntary and it means the `memory_map` value passed by the user doesn't matter, right?

eitsupi: No, it is not. The internal function `from_raw_ipc()` is `pl$read_ipc()` from the R side, but it is actually a separate function and is only called when reading R raw vectors.
Since raw vectors are not Arrow files, this argument has no meaning and is just removed.

eitsupi: Look at the entire function. It is not removed.

etiennebacher: Ok, thanks for the explanation. There's still a failure in the user guide vignette that I'm trying to fix

PR #1116, opened 2024-05-27
Rename arguments in `$dt$combine()`

etiennebacher:


R/expr__datetime.R
```diff
@@ -89,47 +89,40 @@ ExprDT_round = function(every, offset = NULL) {
...
+#' mapped to the end of its bucket.
 #'
-#' @param tm Expr or numeric or PTime, the number of epoch since or before(if negative) the Date
-#' or tm is an Expr e.g. a column of DataType 'Time' or something into an Expr.
-#' @param tu time unit of epochs, default is "us", if tm is a PTime, then tz passed via PTime.
+#' @param time The number of epoch since or before (if negative) the Date. Can be
+#' an Expr or a PTime.
+#' @inheritParams DataType_Datetime
 #'
+#' @inherit ExprDT_truncate details return
```
eitsupi, line 103: I don't think it is possible to inherit details from `ExprDT_truncate` because the argument names are different.

eitsupi (changes requested):

NEWS.md
```diff
@@ -43,6 +43,8 @@
 - New methods for `RPolarsDataType`: `$is_enum()`, `$is_categorical()`,
   `$is_known()`, `$is_string()`, `$contains_views()`, `$contains_categorical()`
   (#1112).
+- In `$dt$combine()`, the arguments `tm` and `tu` are renamed `time` and
```
eitsupi, line 46:
```suggestion
- In `$dt$combine()`, the arguments `tm` and `tu` are renamed `time` and
```

R/expr__datetime.R
etiennebacher: Good catch, I think I got confused actually because none of the args uses the string duration language. I removed it.

eitsupi (approved):
Thanks!

Issue #1108, opened 2024-05-26
Follow-up tasks for #1104 (Rust Polars 0.40.0 update)


eitsupi:
Follow-up tasks should include the followings:

- [x] Update polars to newer edge version. (#1110, #1117, #1124)
- [x] ~~Add `*_by()` methods.~~ (moved to #1082) (#1115)
- [x] Fix `$collect()` when nested types are included. (Turned out this is an upstream issue, see pola-rs/polars#16553) (#1117)
- [x] Fix `$name$map()` or remove that and add a functionality rename with R function to `$rename()` (or adds `$rename_with()`) (#1114, #1122, #1123)
- [x] Add tests for `pl$scan_ipc(memory_map = TRUE)` (#1120, #1121)

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/issues/1104#issuecomment-2132234148_

etiennebacher:
I'll bump to the latest commit upstream

eitsupi:
The panic issue seems to be a fairly fundamental problem.
The problem is in the LazyFrame's `collect` method, but there is a large amount of code to support background execution and I don't know what is going on.

```r
# This will be paniced
polars::.pr$DataFrame$select(polars::pl$DataFrame(),list(list(list(1))))

# This is ok
polars::.pr$LazyFrame$select(polars::pl$LazyFrame(),list(list(list(1))))
```

@sorhawell @Sicheng-Pan Any thoughts on this?
The problem is happening in line 5 of the following section included upstream, but I don't know where in this repository the Series is hashed.

```rust
impl Hash for Wrap<Series> {
    fn hash<H: Hasher>(&self, state: &mut H) {
        let rs = RandomState::with_seeds(0, 0, 0, 0);
        let mut h = vec![];
        self.0.vec_hash(rs, &mut h).unwrap();
        let h = h.into_iter().fold(0, |a: u64, b| a.wrapping_add(b));
        h.hash(state)
    }
}
```

eitsupi:
Done.

PR #1120, opened 2024-06-02
test: add a test case for the memory_map option

eitsupi:
Part of #1108


eitsupi:
Just adding a small test. So merging.

PR #1122, opened 2024-06-02
feat: add `rename_with`

eitsupi:
Part of #1108


eitsupi:
This just adds new methods, so merging.

etiennebacher:
I don't understand why this is necessary, couldn't we check in `$rename()` when the user passes a function?

eitsupi:
> I don't understand why this is necessary, couldn't we check in `$rename()` when the user passes a function?

I don't see the worth of inspecting the contents of `...` here and switching between completely different functions depending on if block.
`rename` is already different in Python and R, and I don't think it's worth forcing Python to match it.

PR #1123, opened 2024-06-02
docs: remove broken `$name$map`'s document

eitsupi:
Part of #1108


eitsupi:
Only removing broken documents, so merging.

PR #239, opened 2023-06-10
Add `rename()` for DataFrame and LazyFrame

etiennebacher:
Not sure about which args we should have here. I used `existing` and `new` as in the [Rust implementation](https://github.com/pola-rs/polars/blob/814025213d098e66fa73d5f43e14f686c994b282/py-polars/src/lazyframe.rs#L668) but the [Python implementation](https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.rename.html#polars.DataFrame.rename) uses a named list like `list(old = "new")` that I also like.

What do you think?

``` r
library(polars)

pl$
  DataFrame(mtcars)$
  rename(c("mpg", "hp"), c("miles_per_gallon", "horsepower"))
#> shape: (32, 11)
#> ┌──────────────────┬─────┬───────┬────────────┬───┬─────┬─────┬──────┬──────┐
#> │ miles_per_gallon ┆ cyl ┆ disp  ┆ horsepower ┆ … ┆ vs  ┆ am  ┆ gear ┆ carb │
#> │ ---              ┆ --- ┆ ---   ┆ ---        ┆   ┆ --- ┆ --- ┆ ---  ┆ ---  │
#> │ f64              ┆ f64 ┆ f64   ┆ f64        ┆   ┆ f64 ┆ f64 ┆ f64  ┆ f64  │
#> ╞══════════════════╪═════╪═══════╪════════════╪═══╪═════╪═════╪══════╪══════╡
#> │ 21.0             ┆ 6.0 ┆ 160.0 ┆ 110.0      ┆ … ┆ 0.0 ┆ 1.0 ┆ 4.0  ┆ 4.0  │
#> │ 21.0             ┆ 6.0 ┆ 160.0 ┆ 110.0      ┆ … ┆ 0.0 ┆ 1.0 ┆ 4.0  ┆ 4.0  │
#> │ 22.8             ┆ 4.0 ┆ 108.0 ┆ 93.0       ┆ … ┆ 1.0 ┆ 1.0 ┆ 4.0  ┆ 1.0  │
#> │ 21.4             ┆ 6.0 ┆ 258.0 ┆ 110.0      ┆ … ┆ 1.0 ┆ 0.0 ┆ 3.0  ┆ 1.0  │
#> │ …                ┆ …   ┆ …     ┆ …          ┆ … ┆ …   ┆ …   ┆ …    ┆ …    │
[...]

PR #1129, opened 2024-06-03
feat!: make `$rename()`'s behavior closer to python polars'

eitsupi:
A follow up for #1122

- `old_name = "new_name"` instead of `new_name = "old_name"`
- Allow a function input


eitsupi:
@etiennebacher This will definitely break tidypolars, but hopefully it is a meaningful change in terms of consistency with Python.

R/lazyframe__lazy.R
```diff
@@ -1540,66 +1540,46 @@ LazyFrame_melt = function(
...
   mapping = list2(...)
   if (length(mapping) == 0) {
     return(self)
+  } else if (is.function(mapping[[1L]])) {
+    {
+      existing = names(self)
+      new = mapping[[1L]](existing)
+    } |>
+      result() |>
+      uw()
```
etiennebacher, line 1571: Nitpick but I find 

```r
result({
 existing = names(self)
 new = mapping[[1L]](existing)
}) |>
  uw()
```
easier to read

etiennebacher (approved):
Nice, I prefer this order and actually I originally put this `old = new` order when I implemented it one year ago but we decided to follow `dplyr` for more consistency (https://github.com/pola-rs/r-polars/pull/239). 

> This will definitely break tidypolars

Not a big deal, I'll just need to update the internals but it won't change the external behaviour. 

Also for those who use `polars` directly, this will clearly error and not just fail silently.

eitsupi:
> Nice, I prefer this order and actually I originally put this `old = new` order when I implemented it one year ago but we decided to follow `dplyr` for more consistency (#239).

Hmmm, I completely forgot about this.
And I always have trouble with the order of `dplyr::rename()` as well (although `new=old` makes more sense since it is consistent with `dplyr::select()`)

Issue #942, opened 2024-03-19
[DISC] Import the `rlang` package


eitsupi:
Related to #939 (and potentially #367)

With limited development resources, I don't want to reinvent the wheel any further with what `rlang` has to offer.
I think there used to be value in not having dependencies to install binary packages on GitHub releases, but I don't see much value in not having dependencies now that installing from the R-universe works so well.

`rlang` is very widely used and `rlang` itself is a very small dependency since it does not depend on any other package.

## Pros

- Features such as trailing commas and dot checks will be available.

## Cons

- This will be the first external required dependency.

---

@etiennebacher @sorhawell @vincentarelbundock @grantmcdermott @Sicheng-Pan Thoughts?

lorenzwalthert:
Or use https://github.com/wch/staticimports?

eitsupi:
Recently to enforce named arguments with `...` is increasingly used, and given the frequency with which Polars change argument names, it would be worthwhile to use the `rlang::check_dots_empty()` function to check arguments to reduce implicit failures for the user.

I am not sure how much emphasis on no dependencies is worth.

eitsupi:
Closing in favor of #1152

Issue #1126, opened 2024-06-02
Switch from extendr to savvy


eitsupi:
I don't think this can be accomplished in a short term, but it would be far easier to maintain by eliminating the large amount of code written to return a Result type to R with extendr.
I also think we can simplify the source code in several other points.

etiennebacher:
Not familiar with `savvy` so I don't see how it would simplify the code while keeping the same functionalities, can you add some examples of simplification?

eitsupi:
@etiennebacher It would be quicker to look at a real example.
The following two R packages are based on savvy and DataFusion.
<https://github.com/yutannihilation/datafusion-r>
<https://github.com/eitsupi/r-glaredb>

For example, implementing the Arrow C stream interface is much easier than extendr.
The following few lines will pull nanoarrow_array_stream, etc. into the Rust side.

```rust
use arrow::ffi_stream::{ArrowArrayStreamReader, FFI_ArrowArrayStream};

let stream_reader = unsafe {
    let stream = savvy::ExternalPointerSexp::try_from(stream_ptr)?
        .cast_mut_unchecked::<FFI_ArrowArrayStream>();
    ArrowArrayStreamReader::from_raw(stream).map_err(|e| e.to_string())?
};
```

It would also allow for more general Rust rules for file structure, such as the ability to split impl blocks into multiple files. (extendr/extendr#739)

eitsupi:
Currently, the file structure on the Rust side is quite different from that of py-polars and uses a uniquely defined error type, so it is necessary to write functions according to completely different rules than for general packages using extendr.
This is fine for this package alone (?), but if we are creating something like pyo3-polars, I expect that similar code will be required for that package as well, making maintenance quite cumbersome.
So I think it might be worthwhile to completely rewrite r-polars in savvy to be a copy of py-polars.

Another thing I am thinking is that just as Python Polars have, for example, the DataFrame class (defined in Python) and the (internal) PyDataFrame class (defined in Rust), it would be better to have a dual structure for each class in R Polars.
Note: extendr and savvy have different ways of representing the related functions of Struct on the R side (I feel that savvy is more appropriate).

eitsupi:
@etiennebacher Although only basic type conversion from R to Polars has been implemented yet, I have created a polars binding using savvy.

Note that the directory structure matches that of py-polars.
This is not possible with extendr, which cannot have multiple impl blocks (extendr/extendr#538).
[...]

Issue #116, opened 2023-04-14
polars should prefer to convert to  R double over bit64 as default. bit64 should be optional.


sorhawell:
see [surprising output from null_count in ](https://github.com/pola-rs/r-polars/pull/105#issuecomment-1508461861) when library(bit64) is not loaded. Better to return as f64, and let those who appreciate the increase value range actively opt-in to bit64

sorhawell:
would make sense to solve #95 first, to avoid passing any possible option as a litteral arg through all conversion functions.

etiennebacher:
Superseded by #465

PR #1144, opened 2024-06-13
Add a base R `check_empty_dots()`

etiennebacher:
Related to #942 

It seems unnecessary to me to import `rlang` to check for empty dots. Also `rlang::check_empty_dots()` is very expensive on my machine, about 200ms, but I didn't find why (if it's due to `rlang` update, RStudio, etc.). I'm not completely against having `rlang` but if we can have easy base R alternatives I don't think it's worth importing it.

This PR is just an example of what it would look like, for example with `$top_k()`:

```r
pl$DataFrame(a = c(6, 1, 0, NA, Inf, NaN))$select(pl$col("a")$top_k(5, nulls_last = TRUE, FALSE, 1))
```
```
   0: In R: in $select()
   0: During function call [pl$DataFrame(a = c(6, 1, 0, NA, Inf, NaN))$select(pl$col("a")$top_k(5, 
          nulls_last = TRUE, FALSE, 1))]
   1: `...` must be empty. Problematic values: FALSE, 1
```

### DIFF
--- a/R/expr__expr.R
+++ b/R/expr__expr.R
@@ -1417,6 +1417,7 @@ Expr_sort = function(..., descending = FALSE, nulls_last = FALSE) {
 #' @examples
 #' pl$DataFrame(a = c(6, 1, 0, NA, Inf, NaN))$select(pl$col("a")$top_k(5))
 Expr_top_k = function(k, ..., nulls_last = FALSE, maintain_order = FALSE, multithreaded = TRUE) {
+  check_dots_empty(...)
   if (!is.numeric(k) || k < 0) stop("k must be numeric and positive, prefereably integerish")
   .pr$Expr$top_k(self, k, nulls_last = nulls_last, maintain_order = maintain_order, multithreaded = multithreaded) |>
     unwrap("in $top_k():")
--- a/R/utils.R
+++ b/R/utils.R
@@ -678,3 +678,15 @@ is_named = function(x) {
   }
   TRUE
 }
+
+get_dots <- function(...) {
+  eval(substitute(alist(...)))
+}
+
+check_dots_empty <- function(...) {
+  dots <- get_dots(...)
+  if (length(dots) > 0) {
+    Err_plain("`...` must be empty. Problematic values:", toString(dots)) |>
+      unwrap()
+  }
+}


eitsupi:
You may want to see `check_dots_empty0`.

> we can have easy base R alternatives I don't think it's worth importing it.

Your implementation does not handle named arguments properly.

etiennebacher:
The main issue for now is that I have perf issues with `rlang`: https://github.com/r-lib/rlang/issues/1718

This is rare, but it's the kind of thing we can avoid by having small replacements for those functions. Again I'm not completely opposed to importing `rlang` but I'd like to know what are the features we're going to use before importing it. We have a mechanism to report Rust errors that works and any refactor is quite tedious considering the size of the code base.

> Your implementation does not handle named arguments properly.

Not sure I follow, do you have an example? Below it seems that my implementation and `rlang` work or error on the same cases.

``` r
library(rlang)

my_check_dots_empty <- function(...) {
  dots <- eval(substitute(alist(...)))
  if (length(dots) > 0) {
    polars:::Err_plain("`...` must be empty. Problematic values:", toString(dots)) |>
      polars:::unwrap()
  }
}

f_rlang <- function(..., xyz = 1) {
  check_dots_empty0(...)
  xyz
}
f_base <- function(..., xyz = 1) {
  my_check_dots_empty(...)
  xyz
}


f_rlang(xyz = 1)
#> [1] 1
```

``` r
f_base(xyz = 1)
#> [1] 1
```

``` r

f_rlang(xy = 1)
#> Error in `f_rlang()`:
#> ! `...` must be empty.
#> ✖ Problematic argument:
#> • xy = 1
```

``` r
f_base(xy = 1)
#>    1: `...` must be empty. Problematic values: 1
```

``` r

f_rlang(xyz = 1, b =)
#> Error in `f_rlang()`:
#> ! `...` must be empty.
#> ✖ Problematic argument:
#> • b = <empty>
```

``` r
f_base(xyz = 1, b =)
#>    1: `...` must be empty. Problematic values:
```

eitsupi:
> > Your implementation does not handle named arguments properly.
> 
> Not sure I follow, do you have an example? Below it seems that my implementation and `rlang` work or error on the same cases.

Sorry for my lack of explanation. I meant to point out that the argument names are not included in the error message.

eitsupi:
> The main issue for now is that I have perf issues with `rlang`, even with `check_dots_empty0()`: [r-lib/rlang#1718](https://github.com/r-lib/rlang/issues/1718)

I do not understand the importance of speed when errors are expected to occur.

etiennebacher:
> I do not understand the importance of speed when errors are expected to occur.

True, it's not crucial (except if one plays a lot with `try()`).

Just wondering: supposing we move to `savvy` (https://github.com/pola-rs/r-polars/issues/1126), would it make sense to refactor the way errors are structured based on how `savvy` handles them? If so, maybe at that point importing `rlang` can be also useful for clearer error messages. 

To clarify, I find the current error messages okay in the sense that usually all the information is there, but they are quite dense are different from more standard R messages. Having something more user friendly if we're going to refactor the internals anyway would be nice.

eitsupi:
> Just wondering: supposing we move to `savvy` (#1126), would it make sense to refactor the way errors are structured based on how `savvy` handles them? If so, maybe at that point importing `rlang` can be also useful for clearer error messages.

As I mentioned in #116, I think you would benefit from using rlang because the savvy error messages are from standard R (similar to the tidyverse package on top of the cpp11 package).

See also the following issues for background on why r-polars needed to implement its own error handling.
extendr/extendr#278
PRQL/prqlc-r#46 (As for prqlr, we removed the code for error handling when we switched to savvy to support webR. But since polars can't be WASM compiled, it's irrelevant here.)

By the way, did you have any problems with tidypolars with a mix of rlang errors and errors coming from r-polars?

etiennebacher:
> By the way, did you have any problems with tidypolars with a mix of rlang errors and errors coming from r-polars?

I didn't have any technical problems but the style of errors clearly depends on whether it comes from `tidypolars` (e.g using a function that isn't translated) or from `polars` (e.g passing a float column to a string function):

``` r
library(dplyr, warn.conflicts = FALSE)
library(tidypolars)

test <- as_polars_df(iris)

### tidypolars doesn't know the function
test |> 
  mutate(x = data.table::chorder(Species))
#> Error in `mutate()`:
#> ! `tidypolars` doesn't know how to translate this function: `data.table::chorder()`
```

``` r

### tidypolars knows the function but error from polars because wrong input
test |> 
  mutate(x = stringr::str_count(Sepal.Length))
#>    0: In R: in $with_columns()
#>          invalid series dtype: expected `String`, got `f64`
```

PR #1147, opened 2024-06-23
Bump rust-polars to 0.41

etiennebacher:
I want to start checking empty dots after this PR  (meaning that we don't release a new version just after this PR)

---

TODO:

- [x] uncomment commented tests for `map_batches()`
- [x] tests for `statistics` options in sink/write_parquet


eitsupi:
Thanks for working on this!

Looking at the polars repository it looks like a patch version is planned (https://github.com/pola-rs/polars/issues/17133#issuecomment-2184678696), so perhaps we should wait for that before merging.

etiennebacher:
Sure, there's still plenty of work to do anyway but I'll add this in the top post, thanks

etiennebacher:
The rust part now compiles

NEWS.md
```diff
@@ -2,21 +2,69 @@
...
 ## Polars R Package (development version)
 
+Updated rust-polars to 0.41.2 (#1147).
+
 ### Breaking changes
 
 - In `$n_chunks()`, the default value of `strategy` now is `"first"` (#1137).
--`$sample()` for Expr and DataFrame (#1136):
+
+- `$sample()` for Expr and DataFrame (#1136):
```
eitsupi, line 11: My understanding is that if a blank line is inserted in a Markdown list, it may no longer be considered the same list. What is the intention of adding a blank line here?

etiennebacher: Ah I didn't know that. I just added that to make it easier to parse in raw mode but I'll change that if it breaks markdown

etiennebacher:
@eitsupi everything passes except the README because of this bug: https://github.com/pola-rs/polars/issues/17308

Let's see how it is handled upstream, there might be another minor release of rust-polars that we could integrate in this PR.

vignettes/userguide.Rmd
eitsupi, line 1: It is not clear to me what the purpose of these changes are.
If we simply want to temporarily disable an executable code block, I think it would be easiest to change `{r}` to `{.r}`.

R/utils.R
```diff
@@ -678,3 +678,52 @@ is_named = function(x) {
   }
   TRUE
 }
+
+# Used in parquet write/sink
+translate_statistics = function(statistics) {
+  if (is.logical(statistics)) {
```
eitsupi, line 684: I think `length(statistics) == 1L` is needed for before checking the vector type.

eitsupi (approved):
Thanks for the great work!
A few minor comments.

vignettes/userguide.Rmd
etiennebacher: Yeah I just wanted to make the CI pass. I changed to use `.r` now. Btw this vignette really needs an update

etiennebacher:
Thanks for the review. The issue reported upstream has been fixed so later today I will update the rust-polars version to unreleased (unless 0.41.3 is reported in the meantime).

etiennebacher:
@eitsupi I'll try to deal with checking dots later in the week, and then I think we can release 0.18

Issue #1152, opened 2024-06-27
[Proposal] Completely rewritten based on savvy and rlang


eitsupi:
(The following text is copied from <https://github.com/eitsupi/neo-r-polars>'s README)

---

## Motivation

I have been developing r-polars for over a year, and I felt that a significant rewrite was necessary.
r-polars is a clone of [py-polars](https://github.com/pola-rs/polars/tree/main/py-polars),
but the package structure is currently quite different.
Therefore, it was difficult to keep up with frequent updates.

I thought that now, around the release of Python Polars 1.0.0, is a good time for a complete rewrite, so I decided to try it.

There are several reasons to rewrite r-polars on both the Rust and R sides.

### Rust side

1. Appropriate file division. Due to the limitations of [extendr](https://github.com/extendr/extendr),
   it is not possible to place multiple impl blocks.
   ([extendr/extendr#538](https://github.com/extendr/extendr/issues/538))
2. Error handling.
   There is a lot of custom code to use the Result type with extendr, which is quite different from other packages based on extendr.
   ([extendr/extendr#650](https://github.com/extendr/extendr/issues/650))
3. Simplify type conversion.
   The code is difficult to follow because it uses a macro called `robj_to` for type conversion (at least in rust-analyzer).

About 1 and 2, I expect that switching from extendr to [savvy](https://github.com/yutannihilation/savvy)
will improve the situation.

For 3, in py-polars and nodejs-polars, a thin `Wrap` struct wraps other types and processes them with standard `From` traits etc.,
which I think makes the code cleaner.

### R side

1. The structure of classes.
   In py-polars, the strategy is that classes defined on the Rust side (e.g., `PyDataFrame`) are wrapped by classes defined on the Python side (e.g., `DataFrame`).
   In r-polars, a complex strategy is adopted to update classes created by Rust side/extendr (e.g., `RPolarsDataFrame`) with a lot of custom code.
   (This is also related to the fact that extendr makes associated functions of Rust structs members of R classes. savvy does not mix associated functions and methods.)
2. S3 methods first.
   This is also related to the Rust side, in the current r-polars, generic functions like `as_polars_series` were added later,
   so there are several places where type conversion from R to Polars is done on the Rust side, making it difficult to understand where the type conversion is done.
   If type conversion from R to Polars is done with two generic functions, `as_polars_series` and `as_polars_expr`, the code will be much simpler and customization from the R side will be possible.
3. Error handling.
[...]

PR #1159, opened 2024-07-05
docs: use r-multiverse repo instead of rpolars repo for installing the latest release version

### LINKED ISSUE #1151: Switch recommended installation way to from R-multiverse community universe
Continued from #888.

R-multiverse (https://github.com/r-multiverse) now seems to be a project sponsored by the R Consortium and is a special R-universe repository.

Now that the released version can be installed from R-multiverse, our managed R-universe repository will be available for other uses, such as for pre-release versions.

eitsupi:
Close #1151

Related to #1152, I hope to be able to use GitHub pre-releases in the future to install rewritten versions of the polars package from the rpolars repo.
r-multivese is always tied to the latest release, but rpolars repo can be tied to specific git tag, so pre-releases are ignored by r-multivesre and only rpolars can be updated.
(of course, we will need to call for a switch in installation approach on the NEWS file before that can happen).

## Background

@shikokuchuo says that the R-multiverse community universe (<https://community.r-multiverse.org/builds>) is ready for use, but will not be actively promoted until the production universe (<https://production.r-multiverse.org/builds>) is complete.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -20,6 +20,7 @@ knitr::opts_chunk$set(
 # polars
 
 <!-- badges: start -->
+[![R-multiverse status](https://img.shields.io/badge/dynamic/json?url=https%3A%2F%2Fcommunity.r-multiverse.org%2Fapi%2Fpackages%2Fpolars&query=%24.Version&label=r-multiverse)](https://community.r-multiverse.org/polars)
 [![R-universe status badge](https://rpolars.r-universe.dev/badges/polars)](https://rpolars.r-universe.dev)
 [![CRAN status](https://www.r-pkg.org/badges/version/polars)](https://CRAN.R-project.org/package=polars)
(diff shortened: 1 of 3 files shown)


etiennebacher (commented):
So just to be sure I'm not missing anything, what will happen to installation from "https://rpolars.r-universe.dev"? Will it be removed?

etiennebacher (approved):

eitsupi:
> So just to be sure I'm not missing anything, what will happen to installation from "[rpolars.r-universe.dev](https://rpolars.r-universe.dev)"? Will it be removed?

No, it will not be deleted.
However, since we were currently using <https://rpolars.r-universe.dev> to distribute the release version, we had no way to distribute the development version binaries, so I think it might be worthwhile to switch <https://rpolars.r-universe.dev> to the pre-release version in the future.

PR #1183, opened 2024-08-14
Bump rust-polars to 0.42.0

### LINKED ISSUE #1185: Refactor optimization toggles
While working on #1183, I realized that py-polars doesn't have `$set_` and `$get_optimization_toggle()`. I think those were implemented early in r-polars to go with the `inherit_optimization` parameter in several functions (e.g. `$collect()`) but this parameter also doesn't exist in py-polars. Instead, py-polars has `.optimization_toggle()` that is used in several functions.

Bottom line, I think we should:

* remove `$get_optimization_toggle()`
* remove the `inherit_optimization` param
* rename `$set_optimization_toggle()` to `$optimization_toggle()`

@eitsupi, thoughts?

etiennebacher:
Close #1185 

https://github.com/pola-rs/polars/releases/tag/rs-0.42.0

~@eitsupi I think the most annoying change is the move from `pl_flavor` to `compatlevel` made here: https://github.com/pola-rs/polars/pull/17421~


etiennebacher:
Only 2 tests failing. 

```
df = pl$DataFrame(mtcars)
df$sort(c("cyl", "mpg"), descending = c(TRUE, FALSE), maintain_order = TRUE)
```

now returns both columns sorted in decreasing order and I can't figure out why.

etiennebacher:
@eitsupi really stuck on this failure in `$sort()`, it would be great if you could take a look. Small summary of what's wrong:

Python 1.5.0:
```python
import polars as pl
df = pl.DataFrame({"x": [1, 2, 1, 2], "y": [1, 2, 3, 4]})
df.sort(["x", "y"], descending=[True, False])
```
```
shape: (4, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 2   ┆ 2   │
│ 2   ┆ 4   │
│ 1   ┆ 1   │
│ 1   ┆ 3   │
└─────┴─────┘
```


This PR:
```r
dat = pl$DataFrame(x = c(1, 2, 1, 2), y = 1:4)
dat$sort(c("x", "y"), descending = c(TRUE, FALSE))
```
```
shape: (4, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ f64 ┆ i32 │ # <---------- col "y" should be in ascending order, not descending
╞═════╪═════╡
│ 2.0 ┆ 4   │
│ 2.0 ┆ 2   │
│ 1.0 ┆ 3   │
│ 1.0 ┆ 1   │
└─────┴─────┘
```

Current `main` (works correctly):
```
shape: (4, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ f64 ┆ i32 │
╞═════╪═════╡
│ 2.0 ┆ 2   │
│ 2.0 ┆ 4   │
│ 1.0 ┆ 1   │
│ 1.0 ┆ 3   │
└─────┴─────┘
```

eitsupi:
In a8c20da4448f979b1327368285690334b805b41b, it works correctly.

```r
> dat$sort(c("x", "y"), descending = c(TRUE, FALSE))
Error:
! Evaluation failed in `$sort()`.
Caused by error in `extend_bool()` at neo-r-polars/R/lazyframe-frame.R:58:5:
! the length of `descending` (2) does not match the length of `...` (1)
Run `rlang::last_trace()` to see where the error occurred.

> dat$sort("x", "y", descending = c(TRUE, FALSE))
shape: (4, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ f64 ┆ i32 │
╞═════╪═════╡
│ 2.0 ┆ 2   │
│ 2.0 ┆ 4   │
│ 1.0 ┆ 1   │
│ 1.0 ┆ 3   │
└─────┴─────┘
```

I do not know where the problem lies in the current implementation.

Is it possible that there is a problem where the `by` and `...` are concatenated?
I think the `by` argument should be eliminated, as well as the `next` branch.

https://github.com/pola-rs/r-polars/blob/a8c20da4448f979b1327368285690334b805b41b/R/lazyframe-frame.R#L48-L69

etiennebacher:
> dat$sort(c("x", "y"), descending = c(TRUE, FALSE))
> Error:
> ! Evaluation failed in `$sort()`.
> Caused by error in `extend_bool()` at neo-r-polars/R/lazyframe-frame.R:58:5:
> ! the length of `descending` (2) does not match the length of `...` (1)
> Run `rlang::last_trace()` to see where the error occurred.

This doesn't match the behaviour of py-polars, that accepts input as `["x", "y"]` which I think is the equivalent of `c("x", "y")`.

The issue is probably that this line:


transforms `c("x", "y")` to `[cols(["x", "y"])]`, while it transforms `"x", "y"` to `[col("x"), col("y")]`. Therefore, eliminating `by` doesn't change anything. I think on the R side, `unpack_list()` should be modified but I fear it's gonna break several things.

eitsupi:
> This doesn't match the behaviour of py-polars, that accepts input as `["x", "y"]` which I think is the equivalent of `c("x", "y")`.

Good point.
I was mistaken, Python's `parse_into_list_of_expressions(["x", "y"])` seems to return `[col("x"), col("y")]` instead of `[cols(["x", "y"])]`.

eitsupi:
> > This doesn't match the behaviour of py-polars, that accepts input as `["x", "y"]` which I think is the equivalent of `c("x", "y")`.
> 
> Good point. I was mistaken, Python's `parse_into_list_of_expressions(["x", "y"])` seems to return `[col("x"), col("y")]` instead of `[cols(["x", "y"])]`.

I've done a little digging into this, and rather than try to do a complete reproduction, I think it would be best to follow the general dynamic dots rule (use `!!!`) rather than a full reproduction.

```r
> dat$sort(!!!c("x", "y"), descending = c(TRUE, FALSE))
shape: (4, 2)
┌─────┬─────┐
│ x   ┆ y   │
│ --- ┆ --- │
│ f64 ┆ i32 │
╞═════╪═════╡
│ 2.0 ┆ 2   │
│ 2.0 ┆ 4   │
│ 1.0 ┆ 1   │
│ 1.0 ┆ 3   │
└─────┴─────┘
```

etiennebacher:
I got it, we had already the code for `<expr>$sort_by()`, the trick was to use `wrap_elist_result()` instead of `unpack_list()`.

I hope the rewrite is going well, I'm really confused with all those internal functions 😄 

> I think it would be best to follow the general dynamic dots rule (use !!!) rather than a full reproduction.

I think that's something to think about for the rewrite, but for now I'd rather keep the current behavior. I don't work a lot with Python, so overall I don't really understand the point of handling `by` and `...` separately in py-polars. Do you know if (and why) this is better than just having `...`?

etiennebacher:
@eitsupi I have changed rust-version to 1.81.0 in Cargo.toml but there's still this error in CI:

> error: rustc 1.80.1 is not supported by the following packages:
>   r-polars@0.42.0 requires rustc 1.81.0

Do you know how to fix it? I don't see 1.80.1 anywhere in the package

eitsupi:
Rust 1.81 is not released yet.

eitsupi (approved):
Thanks!

eitsupi:
> I don't work a lot with Python, so overall I don't really understand the point of handling `by` and `...` separately in py-polars. Do you know if (and why) this is better than just having `...`?

My understanding is that this is simply due to copying from Python Polars.
As you may remember, `pl$col()` used to have a similar structure, which could cause problems. I think only `...` is needed.

etiennebacher:
> My understanding is that this is simply due to copying from Python Polars.

Actually I was asking for the python implementation. Maybe it's common in python, but I don't understand the point of having `by` and `*more_by`. To me, it makes more sense to only have `*more_by`. The only reason I can come up with is that it's a standard approach in Python, so I was asking whether you know this.

I think it's a weird pattern in R and it makes it harder for us to maintain because we have to ensure that `by` and `...` are correctly handled instead of just `...`. At the same time, I'm a bit hesitant on diverting from the python API...

eitsupi:
I think this is because the things that can be specified are different.
In other words, here we can specify `Iterable[IntoExpr]` for `by`, but not for `*more_by`.

Python Polars' API is derived from pyspark, and we can find the same pattern in pyspark as well.
<https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.DataFrameWriter.bucketBy.html#pyspark.sql.DataFrameWriter.bucketBy>

> colstr, list or tuple
>    a name of a column, or a list of names.
> colsstr
>    additional names (optional). If col is a list it should be empty.

PR #1154, opened 2024-07-01
updated extendr to 0.7.0

CGMossa:
These are the changes necessary to update r-polars.
I discovered that we could make our macros a bit nicer, so I've done so in:

- [x] https://github.com/extendr/extendr/pull/818

We will consider a patch release for extendr, where these improvements can be in. We'll just wait a few days / (1-2 weeks) to figure out if there are more patches to be made.

Cheers!

Close #1105

PR #1187, opened 2024-08-17
Chore/extendr 0.7.0 support

ju6ge:
This MR builds upon #1154 and updates the test suite and rust code to work as expected.

### DIFF
--- a/src/rust/src/lazy/dsl.rs
+++ b/src/rust/src/lazy/dsl.rs
@@ -55,7 +55,6 @@ impl From<pl::Expr> for RPolarsExpr {
     }
 }
 
-use extendr_api::Result;
 #[extendr]
 impl RPolarsExpr {
     //constructors
--- a/src/rust/src/lazy/dsl.rs
+++ b/src/rust/src/lazy/dsl.rs
@@ -20,7 +20,6 @@ use polars::prelude as pl;
 use polars::prelude::{ExprEvalExtension, NestedType, SortOptions};
 use std::any::Any;
 use std::ops::{Add, Div, Mul, Rem, Sub};
-use std::result::Result;
 pub type NameGenerator = pl::Arc<dyn Fn(usize) -> String + Send + Sync>;
 use crate::rdatatype::robjs_to_ewm_options;
 use crate::utils::r_expr_to_rust_expr;
@@ -56,6 +55,7 @@ impl From<pl::Expr> for RPolarsExpr {
     }
 }
 
+use extendr_api::Result;
 #[extendr]
 impl RPolarsExpr {
     //constructors
@@ -2504,7 +2504,7 @@ impl RPolarsExpr {
             .into())
     }
 
-    pub fn bin_starts_with(&self, sub: Robj) -> Result<Self, String> {
+    pub fn bin_starts_with(&self, sub: Robj) -> std::result::Result<Self, String> {
         Ok(self
             .0
             .clone()
@@ -2513,7 +2513,7 @@ impl RPolarsExpr {
             .into())
     }
 
-    pub fn bin_ends_with(&self, sub: Robj) -> Result<Self, String> {
+    pub fn bin_ends_with(&self, sub: Robj) -> std::result::Result<Self, String> {
         Ok(self
             .0
             .clone()
@@ -2570,7 +2570,7 @@ impl RPolarsExpr {
         List::from_values(exprs.iter().map(|e| RPolarsExpr(e.clone())))
     }
 
-    fn meta_eq(&self, other: Robj) -> Result<bool, String> {
+    fn meta_eq(&self, other: Robj) -> std::result::Result<bool, String> {
         let other = robj_to!(Expr, other)?;
         Ok(self.0 == other.0)
     }
@@ -2585,7 +2585,7 @@ impl RPolarsExpr {
             .collect()
     }
 
-    fn meta_output_name(&self) -> Result<String, String> {
+    fn meta_output_name(&self) -> std::result::Result<String, String> {
         let name = self
             .0
             .clone()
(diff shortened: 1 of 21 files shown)


src/rust/src/lazy/dsl.rs
```diff
@@ -20,7 +20,6 @@ use polars::prelude as pl;
 use polars::prelude::{ExprEvalExtension, NestedType, SortOptions};
 use std::any::Any;
 use std::ops::{Add, Div, Mul, Rem, Sub};
-use std::result::Result;
```
etiennebacher, line 23: Why this change? Isn't it cleaner to keep this rather than putting `std::result::` everywhere?

src/rust/src/utils/mod.rs
```diff
@@ -743,8 +743,11 @@ pub fn robj_to_rarrow_field(robj: extendr_api::Robj) -> RResult<Robj> {
 
 pub fn robj_to_datatype(robj: extendr_api::Robj) -> RResult<RPolarsDataType> {
     let rv = rdbg(&robj);
-    let res: ExtendrResult<ExternalPtr<RPolarsDataType>> = robj.try_into();
-    let ext_dt = res.bad_val(rv).mistyped(tn::<RPolarsDataType>())?;
+    if rv != "ExternalPtr.set_class([\"RPolarsDataType\"]" {
+        return Err(RPolarsErr::new().bad_val(rv).mistyped(tn::<RPolarsDataType>().to_string()).into());
+    }
```
etiennebacher, line 748: As you said in https://github.com/pola-rs/r-polars/pull/1154#issuecomment-2294958253, checking on this string looks very fragile

tests/testthat/test-dataframe.R
```diff
@@ -157,10 +157,7 @@ test_that("DataFrame, custom schema", {
   )
   # incorrect datatype
   expect_grepl_error(pl$DataFrame(x = 1, schema = list(schema = foo)))
-  expect_grepl_error(
-    pl$DataFrame(x = 1, schema = list(x = "foo")),
-    "expected RPolarsDataType"
-  )
+  expect_grepl_error(pl$DataFrame(x = 1, schema = list(x = "foo")))
```
etiennebacher, line 160: This change (and others below) make the test pass but they degrade the error messages. For example, instead of 
`expected RPolarsDataType` we now have `Expected ExternalPtr got Strings` which is quite cryptic. Is there a way to not modify those error messages?

tests/testthat/test-lazy_functions.R
```diff
@@ -146,8 +146,8 @@ test_that("pl$n_unique", {
   expr_act = pl$n_unique("bob")
   expect_true(expr_act$meta$eq(pl$col("bob")$n_unique()))
 
-  expect_grepl_error(pl$n_unique(pl$all()))
-  expect_grepl_error(pl$n_unique(1))
+  expect_error(pl$n_unique(pl$all()))
+  expect_error(pl$n_unique(1))
```
etiennebacher, line 150: `expect_grepl_error()` is different from `expect_error()` because it can catch polars errors that have a weird structure in R.

If we don't check the pattern of the message, then `expect_error()` can be used indeed, but can you rather keep `expect_grepl_error()` and add some pattern matching?

etiennebacher (changes requested):
@ju6ge Many thanks for your efforts here and in #1154, they are really appreciated.

My main issue with this PR is that it can degrade some error messages, can you take a look at that?

Also, as you say in https://github.com/pola-rs/r-polars/pull/1154#issuecomment-2294958253:

> Though I am still unsure that the changes are sufficient. There might me be more parts of the code that use the previous behavior and expect to return the error to R. So there might be more cases where users can run into segfaults that are not tested for already!

> At the very least I would like a better way to check if the converesion will work before attempting it. Checking against a string is very unergonomic!

so I'm not too confident this should be merged as-is. @CGMossa I know  you already answered to some of that in #1154, but it would help if you could just skim through the changes in Rust code.

I'd also like @eitsupi to check that

ju6ge:
Sure thing, I will gladly update this MR to address the concerns about the tests. Though I will have to see if I find time for this next week. 

I am coming more from the rust side of things, and the `R` side of things is still a bit more cryptic to me. Can you give me some pointers on how `expect_grepl_error` expects an certain error type. Just setting the expected `error` as a string seems quite strange to me.

eitsupi (commented):
@ju6ge Thanks for working on this!
It would be great to see the R CMD check error resolved.

I am wondering if you are overlooking #1152.
I mean, we may abandon our entire current codebase in the near future, and since we will be switching from extendr to savvy in the future, I don't see much value in spending time on an extendr-only hack.

If you are willing to work on developing r-polars, it would be great if you could check out the `next` branch.

etiennebacher:
> Can you give me some pointers on how `expect_grepl_error` expects an certain error type. Just setting the expected `error` as a string seems quite strange to me.

Normally, using `expect_error(<code>, <error message pattern>)` would work, but the error handling in r-polars is a bit convoluted. It relies on some internal `unwrap()` that sort of mimics the Rust error handling but doesn't play super well with `testthat::expect_error()`, see e.g. https://github.com/pola-rs/r-polars/pull/987#discussion_r1545265365.

Therefore, we made a custom `expect_grepl_error()` that should be a drop-in replacement for `expect_error()`. You don't have to put the full error message in the second argument, just a piece of it to ensure that the expected error message actually matches what we expect. There are plenty of examples in the tests folder.

ju6ge:
@eitsupi I have added my thoughts to #1152

@etiennebacher thanks for the explanation, i will take a look at it and see what i come up with.

src/rust/src/utils/mod.rs
CGMossa: Oh I see now.. I think this can be addressed in https://github.com/extendr/extendr/pull/853. I just added a test that shows that we can deal with this situation, with very accurate type-checking.

eitsupi:

PR #878, opened 2024-03-01
fix!: POSIXct without timezone and naive time handling

eitsupi:
Fix #875

The latest version (0.14.1):

``` r
Sys.setenv(TZ = "America/New_York")

rdf = data.frame(
  non_exsitent_time = as.POSIXct("2020-03-08 02:00:00"),
  ambiguous_time = as.POSIXct("2020-11-01 01:00:00")
)
rdf
#>     non_exsitent_time      ambiguous_time
#> 1 2020-03-08 01:00:00 2020-11-01 01:00:00

pldf = polars::as_polars_df(rdf)
pldf
#> shape: (1, 2)
#> ┌─────────────────────┬─────────────────────┐
#> │ non_exsitent_time   ┆ ambiguous_time      │
#> │ ---                 ┆ ---                 │
#> │ datetime[ms]        ┆ datetime[ms]        │
#> ╞═════════════════════╪═════════════════════╡
#> │ 2020-03-08 06:00:00 ┆ 2020-11-01 05:00:00 │
#> └─────────────────────┴─────────────────────┘

as.data.frame(pldf)
#>     non_exsitent_time      ambiguous_time
#> 1 2020-03-08 01:00:00 2020-11-01 01:00:00
```

This PR:

``` r
Sys.setenv(TZ = "America/New_York")

rdf = data.frame(
  non_exsitent_time = as.POSIXct("2020-03-08 02:00:00"),
  ambiguous_time = as.POSIXct("2020-11-01 01:00:00")
)
rdf
#>     non_exsitent_time      ambiguous_time
#> 1 2020-03-08 01:00:00 2020-11-01 01:00:00

pldf = polars::as_polars_df(rdf)
pldf
#> shape: (1, 2)
#> ┌─────────────────────┬─────────────────────┐
#> │ non_exsitent_time   ┆ ambiguous_time      │
#> │ ---                 ┆ ---                 │
#> │ datetime[ms]        ┆ datetime[ms]        │
#> ╞═════════════════════╪═════════════════════╡
#> │ 2020-03-08 01:00:00 ┆ 2020-11-01 01:00:00 │
#> └─────────────────────┴─────────────────────┘
```

In the above example, an error occurs when converting Polars to R.
Perhaps something similar to the Int64 type conversion option should be added to the arguments and global options.

PR #1192, opened 2024-08-21
Remove usage of `TZ` envvar

### LINKED ISSUE #1188: Timezone Conversion Tests fail for me Locally
Hey guys,

https://github.com/pola-rs/r-polars/blob/1294fcc1b670f12aca42ba69186224f273764d2b/tests/testthat/test-datatype.R#L52-L104

this test is failing for me when I run the test suit on my local device. The reason is pretty simple. It seems that polars always assumes input in the local timezone (unless set via `time_zone = …`). This of course then leads to a mismatch with `as.POSIXct` that considers the environment variable. It also means that the invalid date test later in line 72 and 76 also fail because the timestamps are not invalid in my local time. 

I have not tested with older versions of polars so I don't know if this worked in an earlier version, and I am unsure if there is an elegant way to make polars behave like `as.POSIXct`. At least setting `time_zone` will not work direkly because `as.POSIXct` creates a result  without timezone information, and then the test still fails with a time zone mismatch. 

I think the most pragmatic solution would be to change the test, to explicitly set the time_zone both for `as.POSIXct` and for the polars version.

etiennebacher:
Close #1188

Thanks for the report and explanations @ju6ge, it would be great if you could check this PR

### DIFF
--- a/tests/testthat/test-as_polars.R
+++ b/tests/testthat/test-as_polars.R
@@ -391,18 +391,12 @@ patrick::with_parameters_test_that("clock package class support",
       as.POSIXct(clock_sys_time, tz = "Asia/Kolkata")
     )
 
-    # Test on other time zone
-    withr::with_envvar(
-      new = c(TZ = "Europe/Paris"),
-      {
-        expect_equal(as.POSIXct(as.vector(pl_naive_time)), as.POSIXct(clock_naive_time))
-        expect_equal(as.POSIXct(as.vector(pl_zoned_time_1)), as.POSIXct(clock_zoned_time_1))
-        expect_equal(as.POSIXct(as.vector(pl_sys_time)), as.POSIXct(clock_sys_time, tz = "UTC"))
-        expect_equal(
-          as.POSIXct(as.vector(pl_sys_time), tz = "Asia/Kolkata"),
-          as.POSIXct(clock_sys_time, tz = "Asia/Kolkata")
-        )
-      }
+    expect_equal(as.POSIXct(as.vector(pl_naive_time)), as.POSIXct(clock_naive_time))
+    expect_equal(as.POSIXct(as.vector(pl_zoned_time_1)), as.POSIXct(clock_zoned_time_1))
+    expect_equal(as.POSIXct(as.vector(pl_sys_time)), as.POSIXct(clock_sys_time, tz = "UTC"))
+    expect_equal(
+      as.POSIXct(as.vector(pl_sys_time), tz = "Asia/Kolkata"),
+      as.POSIXct(clock_sys_time, tz = "Asia/Kolkata")
     )
   },
   precision = c("nanosecond", "microsecond", "millisecond", "second", "minute", "hour", "day"),
--- a/tests/testthat/test-datatype.R
+++ b/tests/testthat/test-datatype.R
@@ -42,42 +42,32 @@ test_that("POSIXct data conversion", {
     pl$lit("2022-01-01")$str$strptime(pl$Datetime(), "%F")$to_r(),
     as.POSIXct("2022-01-01")
   )
-  # TODO: infer timezone from string, change the arugment name from `tz`
+
+  # Note: do not set the "TZ" envvar here as it doesn't work consistently
+  # https://github.com/pola-rs/r-polars/issues/1188
+  expect_identical(
+    pl$lit("2022-01-01")$str$strptime(pl$Datetime(time_zone = "America/New_York"), "%F")$to_r(),
+    as.POSIXct("2022-01-01", tz = "America/New_York")
+  )
   expect_true(
     as_polars_series("2022-01-01 UTC")$str$strptime(pl$Datetime(time_zone = "UTC"), "%F %Z")$eq(
       as_polars_series(as.POSIXct("2022-01-01", tz = "UTC"))
     )$to_r()
   )
 
-  withr::with_envvar(
-    new = c(TZ = "America/New_York"),
-    {
-      expect_identical(
-        pl$lit("2022-01-01")$str$strptime(pl$Datetime(), "%F")$to_r(),
-        as.POSIXct("2022-01-01")
-      )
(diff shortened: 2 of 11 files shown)


tests/testthat/test-as_polars.R
```diff
@@ -391,18 +391,12 @@ patrick::with_parameters_test_that("clock package class support",
...
-        expect_equal(as.POSIXct(as.vector(pl_naive_time)), as.POSIXct(clock_naive_time))
-        expect_equal(as.POSIXct(as.vector(pl_zoned_time_1)), as.POSIXct(clock_zoned_time_1))
-        expect_equal(as.POSIXct(as.vector(pl_sys_time)), as.POSIXct(clock_sys_time, tz = "UTC"))
-        expect_equal(
-          as.POSIXct(as.vector(pl_sys_time), tz = "Asia/Kolkata"),
-          as.POSIXct(clock_sys_time, tz = "Asia/Kolkata")
-        )
-      }
+    expect_equal(as.POSIXct(as.vector(pl_naive_time)), as.POSIXct(clock_naive_time))
+    expect_equal(as.POSIXct(as.vector(pl_zoned_time_1)), as.POSIXct(clock_zoned_time_1))
```
eitsupi, line 395: I am really confused. What are these tests testing?
Is it guaranteed that these tests will work with other systems' time zone settings?

etiennebacher: Yeah actually those could be removed since they're identical to those a few lines above. The only difference was that they were wrapped in `withr::with_envvar()`.

>  Is it guaranteed that these tests will work with other systems' time zone settings?

I don't know, but as we have seen in #1188, setting `TZ` wasn't enough to check this so I don't know the path to follow here. Do you or @ju6ge have any ideas on this?

tests/testthat/test-datatype.R
```diff
@@ -42,42 +42,32 @@ test_that("POSIXct data conversion", {
     pl$lit("2022-01-01")$str$strptime(pl$Datetime(), "%F")$to_r(),
     as.POSIXct("2022-01-01")
   )
-  # TODO: infer timezone from string, change the arugment name from `tz`
+
+  # Note: do not set the "TZ" envvar here as it doesn't work consistently
+  # https://github.com/pola-rs/r-polars/issues/1188
```
eitsupi, line 47: Under what conditions does this not work? I suspect this is a function bug, not a testing issue.

tests/testthat/test-datatype.R
```diff
@@ -42,42 +42,32 @@ test_that("POSIXct data conversion", {
...
+    pl$lit(as.POSIXct(non_existent_time_chr))$to_r(),
+    as.POSIXct(non_existent_time_chr)
+  )
+  expect_grepl_error(
+    pl$lit(non_existent_time_chr)$str$strptime(pl$Datetime(time_zone = "America/New_York"), "%F %T")$to_r(),
+    "non-existent"
+  )
+  expect_grepl_error(
+    pl$lit(ambiguous_time_chr)$str$strptime(pl$Datetime(time_zone = "America/New_York"), "%F %T")$to_r(),
+    "ambiguous"
```
eitsupi, line 70: What should be tested here is the event of an error in converting naive time to R, which has been replaced by another test that tests for errors when converting datetime with time zone to R.

eitsupi (changes requested):

tests/testthat/test-datatype.R
ju6ge: I think #1188 should have proved that envvar is not reliable.

tests/testthat/test-datatype.R
ju6ge: No, the point of both these test is to ensure that invalid timestamps are caught. The only difference to the previous test is the explicit parameter for `time_zone` in the call to polars. This has nothing to do with naive time. Both parsed values are valid times without knowledge of time_zone, they are only invalid if the time zone is `America/New_York`

tests/testthat/test-datatype.R
eitsupi: I mean to say that there may be a bug in polars that does not respect envvar.

ju6ge: I don't think envvars are universal. Unless you can reference a standard that explicitly was agreed upon by all operating systems that `TZ` is to be regarded as the source of truth when checking what time zone to operate in. So this is quite irrelevant, `polars` behaves correctly when explicitly told to use a specific time zone. And since these test are time_zone specific it is very sensible to explicitly state the time zone.

tests/testthat/test-datatype.R
eitsupi: > No, the point of both these test is to ensure that invalid timestamps are caught. The only difference to the previous test is the explicit parameter for `time_zone` in the call to polars. This has nothing to do with naive time. Both parsed values are valid times without knowledge of time_zone, they are only invalid if the time zone is `America/New_York`

I believe I added these tests, but you claim to know more about these tests than I?
Please check #878.

eitsupi:
@etiennebacher In my opinion this change is premature.
As I mentioned in my comment, I think this could be a bug in the way polars handles system time, not a testing issue.

tests/testthat/test-datatype.R
etiennebacher: Worth noting that `Sys.timezone()` uses this envvar (and then has a fallback):

```r
function (location = TRUE) 
{
    if (!location) 
        .Deprecated(msg = "Sys.timezone(location = FALSE) is defunct and ignored")
    if (!is.na(tz <- get0(".sys.timezone", baseenv(), mode = "character", 
        inherits = FALSE, ifnotfound = NA_character_))) 
        return(tz)
    cacheIt <- function(tz) {
        unlockBinding(".sys.timezone", baseenv())
        assign(".sys.timezone", tz, baseenv())
        lockBinding(".sys.timezone", baseenv())
    }
    tz <- Sys.getenv("TZ")
	...
```

However I don't see any `TZ` in the polars source code upstream.

PR #1195, opened 2024-08-22
test: use `withr::with_timezone()` for temporary timezone change

### LINKED ISSUE #1188: Timezone Conversion Tests fail for me Locally

eitsupi:
Fix #1188
Close #1194

@ju6ge Sorry for bother you, but could you try this branch? Thanks.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -118,5 +118,5 @@ Collate:
     'zzz.R'
 Config/rextendr/version: 0.3.1
 VignetteBuilder: knitr
-Config/polars/LibVersion: 0.41.1
+Config/polars/LibVersion: 0.42.0
 Config/polars/RustToolchainVersion: nightly-2024-07-26
--- a/R/dataframe__frame.R
+++ b/R/dataframe__frame.R
@@ -81,8 +81,8 @@
 #' # so this particular date-time doesn't exist
 #' non_existent_time = as_polars_series("2020-03-08 02:00:00")$str$strptime(pl$Datetime(), "%F %T")
 #'
-#' withr::with_envvar(
-#'   new = c(TZ = "America/New_York"),
+#' withr::with_timezone(
+#'   "America/New_York",
 #'   {
 #'     tryCatch(
 #'       # This causes an error due to the time zone (the `TZ` env var is affected).
@@ -92,8 +92,8 @@
 #'   }
 #' )
 #'
-#' withr::with_envvar(
-#'   new = c(TZ = "America/New_York"),
+#' withr::with_timezone(
+#'   "America/New_York",
 #'   {
 #'     # This is safe.
 #'     as.vector(non_existent_time$dt$replace_time_zone("UTC"))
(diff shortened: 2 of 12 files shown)


ju6ge:
@eitsupi no bother at all ;) 

just ran the test suite on this branch on my machine. Seems to work fine :+1:

etiennebacher (approved):
Thanks both!

eitsupi:
> just ran the test suite on this branch on my machine. Seems to work fine 👍

Great, thanks for checking in so quickly!

Looking at the contents of the `with_timezone()` function, it seems that the difference, other than just overwriting the `TZ` environment variable by `with_envvvar()`, is that it updates the `.sys.timezone` object.
Perhaps your machine has `.sys.timezone` set? (I didn't know such an object existed by default...)

ju6ge:
> > just ran the test suite on this branch on my machine. Seems to work fine 👍
> 
> Great, thanks for checking in so quickly!
> 
> Looking at the contents of the `with_timezone()` function, it seems that the difference, other than just overwriting the `TZ` environment variable by `with_envvvar()`, is that it updates the `.sys.timezone` object. Perhaps your machine has `.sys.timezone` set? (I didn't know such an object existed by default...)

just checked with an empty R session

```R
R

R version 4.4.1 (2024-06-14) -- "Race for Your Life"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> .sys.timezone
[1] NA
```

I think it is definitly better to use `withr::with_timezone` because it should have more os agnosting handlers than just relying on environment variables. But I have not had time to play around with that function to compare the behavior.

eitsupi:
Thanks for checking!
It is above my comprehension that nothing has been set......

Anyway, I now know that dealing with system time zones is more complicated than I imagined. I think we can merge this for now and see how it goes.

ju6ge:
I agree, this can be merged. 

With time zones there be dragons, lets be happy that they have been slayed by other warriors.

PR #1197, opened 2024-08-22
Make a dashboard comparing the Python and R APIs

etiennebacher:
We need some help to maintain `polars`. The project is really big, there's a lot of technical debt, and there are frequent updates upstream. In particular, it's hard to keep up with all the changes in function names or in parameter names. 

This PR is a first step to make this less painful and hopefully encourages more people to participate. Briefly, it is inspired by the [easystats dashboard](https://easystats.github.io/dashboard/dashboard.html) and it provides a Rmd file containing a table that compares functions available in the python and R APIs, as well as the argument names of those functions. It gives an easy way to see which functions/arguments are missing.

The python script should run after every release of the python API.

Here, for example we see that some arguments for some `LazyFrame` methods are missing:

image

---

TODO:
- [x] Make this available for all `Expr` and its subsnamespaces
- [ ] find a way to put this online with a CI script. I think we can't deploy two github pages from one repo so I might move this to a personal repo


etiennebacher:
Repo: https://github.com/etiennebacher/r-polars-dashboard
Website: https://etiennebacher.github.io/r-polars-dashboard/dashboard.html

PR #1215, opened 2024-08-31
docs(news): remove misleading expressions to users

eitsupi:
@etiennebacher As far as etiennebacher/tidypolars#137 is concerned, I don't think the cause of your CI failure is related to the update, so I think the document needs to be changed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,8 +4,7 @@
 
 ## Polars R Package 0.19.1
 
-- Add the option `-z noexecstack` for compilation on Linux to avoid the error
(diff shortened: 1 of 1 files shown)


etiennebacher:
Alright, does this also mean that changes to Makevars in #1212 should be reverted?

etiennebacher (approved):

eitsupi:
I don't know what that flag actually does, so I don't know if it needs to be reverted.
It doesn't seem to be causing any problems so I think it's fine to leave it as is.

PR #1224, opened 2024-09-04
fix: should not use `unwrap()` in `RPolarsLazyFrame.cast`

### LINKED ISSUE #1221: Rework the panic handling in `$cast()`
I've been a bit too quick in #1219, the `unwrap()` I left in `cast()` provokes a panic in one of the tests (which is known) and this isn't handled great when running the tests non-interactively (there's a big panic message, even though the tests still pass)

eitsupi:
Fix #1221


etiennebacher (approved):
Thanks!

PR #1228, opened 2024-09-06
feat: use `_` as big mark (thousand-separator) of shape as like Python Polars

eitsupi:
Ref: pola-rs/polars#8472


etiennebacher (approved):
Thanks!

PR #1230, opened 2024-09-11
Bump rust-polars to 0.43.1

etiennebacher:
Not going to address that very soon, just to ensure we do the bump on this branch

Changelog for 0.43.0: https://github.com/pola-rs/polars/releases/tag/rs-0.43.0
Changelog for 0.43.1: https://github.com/pola-rs/polars/releases/tag/rs-0.43.1


eitsupi:
There seem to be quite a few critical bugs in this version, so perhaps the next release will be made soon.
You may want to wait for that before working on it.

etiennebacher:
@eitsupi there are still a few errors in Rust in `arrow_interop/to_rust.rs` and `series.rs`. They are all due to failures in `Series::try_from()`. Could you take a look? I *think* those are due to https://github.com/pola-rs/polars/pull/18564/files.

Also, it would be great if you could double check my changes in Rust as I'm almost sure some things are not idiomatic. After that I can take care of the R side.

src/rust/src/rdataframe/mod.rs
```diff
@@ -327,7 +328,7 @@ impl RPolarsDataFrame {
 
     pub fn to_struct(&self, name: Robj) -> RResult<RPolarsSeries> {
         use pl::IntoSeries;
-        let name = robj_to!(Option, str, name)?.unwrap_or("");
+        let name = robj_to!(str, name)?.into();
```
eitsupi, line 331: Good catch.

eitsupi:
> They are all due to failures in `Series::try_from()`. Could you take a look?

I fixed it so that it would compile successfully. All the fixes seem to be due to the introduction of PlSmallStr.

> Also, it would be great if you could double check my changes in Rust as I'm almost sure some things are not idiomatic.

I think you did an excellent job. There were no areas that I felt particularly uncomfortable with.

Personally, I would prefer to use the `%||%` operator on the R side rather than the `NULL` exception handling on the Rust side, but since the original code is that way, I don't think it is a big problem (no such point in the `next` branch)

eitsupi:
Hmm, it seems that polars-parquet fails to compile for the arm64 targets.
I'm assuming it's an upstream issue, but I'm not sure what the cause is here, since the Python wheel (for arm64 macOS) build should have succeeded.

eitsupi:
> I guess so, let's see how it is when the remaining Rust errors are fixed.

Does this seem to be the only one giving a broadcast error in the test log?
https://github.com/pola-rs/r-polars/blob/77bba3cf53a0fbaee139514b21b8d8ed0a087f6d/tests/testthat/test-lazy_functions.R#L493-L496

In fact, this is not a big deal, since Python's DataFrame constructor also fails unless all the data lengths match; the `next` branch even documents that it fails.
https://github.com/pola-rs/r-polars/blob/28eae28b1f8b36db60545f8999a422169ce7165f/R/dataframe-frame.R#L21-L25

etiennebacher:
Thanks for the help!

I'll explore the test failures to see what comes from this broadcasting change. I don't have a linux or macos machine to test those compilation failures now but let's see if it comes up in the issue tracker upstream.

etiennebacher:
Pleasantly surprised by the few number of fixes to do, there are just the two failures you mentioned about broadcasting. Those do not always appear, for example this works:

```r
pl$DataFrame(
    start = c("2022-01-01", "2022-01-02"),
    end = c("2022-01-03")
)
```
while this fails:
```r
pl$DataFrame(
    start = c("2022-01-01", "2022-01-02", NA),
    end = as.Date(c("2022-01-03"))
)
```

I'll try to take a look in the next few days, unless you want to take care of that first.

eitsupi:
> I'll try to take a look in the next few days

Perhaps it might be here, and in the match arm that calls `to_series_then_lit()`, branch off into length 1 and other cases, and call `.first()` in the case of length 1?
https://github.com/pola-rs/r-polars/blob/77bba3cf53a0fbaee139514b21b8d8ed0a087f6d/src/rust/src/lazy/dsl.rs#L76-L176

This is a target to be completely rewritten in the `next` branch, so it will be a minimal modification and not worth much effort.

etiennebacher:
Compiling failure on macOS is due to https://github.com/rust-lang/rust/issues/129887, which should have been fixed in more recent nightly versions. I tried bumping the toolchain to 2024-09-19 to see if it works

etiennebacher:
@eitsupi I didn't add much since last time but can you check the fix for literal values?

src/rust/src/lazy/dsl.rs
```diff
@@ -87,8 +87,12 @@ impl RPolarsExpr {
...
             (_, rlen) if rlen != 1 => to_series_then_lit(robj),
             (Rtype::List, _) => to_series_then_lit(robj),
-            (_, _) if robj_inherits(&robj, ["POSIXct", "PTime", "Date"]) => {
-                to_series_then_lit(robj)
+            (_, rlen) if robj_inherits(&robj, ["POSIXct", "PTime", "Date"]) => {
+                if rlen == 1 {
+                    Ok(to_series_then_lit(robj)?.first())
+                } else {
+                    to_series_then_lit(robj)
+                }
```
eitsupi, line 95: In my opinion this is not enough, because the case of length 1 should be handled for all R objects, not only for POSIXct.
So the length determination must be done after the converted Series in the `to_series_then_lit` function, not here.

eitsupi (approved):
Let's merge them as they are now, since literal values are not likely to be used that often and I think a rewrite is needed to fundamentally solve the problem.

Issue #1231, opened 2024-09-15
nix build:  it requires rustc 1.80.0 or newer


artur-sannikov:
If I run `nix build github:pola-rs/r-polars`,

I get 

```
> error: package `r-polars v0.42.2 (/build/source/src/rust)` cannot be built because it 
requires rustc 1.80.0 or newer, while the currently active rustc version is 1.78.0-nightly
```

eitsupi:
flake files are not maintained for past few months.
Contributions are welcome, or we may remove these files.

PR #1232, opened 2024-09-15
ci: set GHA for updating flake.lock

detroyejr:
Fixes https://github.com/pola-rs/r-polars/issues/1231#issuecomment-2351481317

The nix flake lockfile needed an update to support the version of rust r-polars uses. 

Nix build should work after this.

### DIFF
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -3,6 +3,8 @@ on:
   workflow_dispatch: # allows manual triggering
   schedule:
     - cron: '0 0 * * 0' # runs weekly on Sunday at 00:00
+permissions:
+  contents: write
 
 jobs:
   lockfile:
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -1,0 +1,19 @@
+name: update-flake-lock
+on:
+  workflow_dispatch: # allows manual triggering
+  schedule:
+    - cron: '0 0 * * 0' # runs weekly on Sunday at 00:00
+
+jobs:
+  lockfile:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


eitsupi:
Thanks for the quick update!

Do you know if we can set up CI for updating nix flake automatically?

detroyejr:
Looks like that's possible.

Setup an action that can run on a scheduled basis. Opens up a pr that looks like [this](https://github.com/detroyejr/r-polars/pull/1).

.github/workflows/flake.yml
```diff
@@ -0,0 +1,19 @@
+name: update-flake-lock
+on:
+  workflow_dispatch: # allows manual triggering
+  schedule:
+    - cron: '0 0 * * 0' # runs weekly on Sunday at 00:00
```
detroyejr, line 5: If this is something you want to incorporate, let me know if you want to change the default.

eitsupi (approved):
Thanks, this looks good!
Let's merge and see what happens.

.github/workflows/flake.yml
```diff
@@ -0,0 +1,19 @@
...
+
```
eitsupi, line 6:
```suggestion
permissions:
  contents: write

```

detroyejr:
@eitsupi Just a quick note on this, it looks like the scheduled run didn't quite work. If I'm not mistaken, the token provisioned for actions can push to [branches](https://github.com/pola-rs/r-polars/tree/update_flake_lock_action), but [isn't able to open PR's](https://github.com/pola-rs/r-polars/actions/runs/10976693004/job/30477726181#step:4:416).

Everything else on the run looked okay so I think this will go be good to go when that's updated.

PR #1234, opened 2024-09-16
fix: ensure report rustc version if pre-built binary will be used

eitsupi:
R CMD check now checks to see if the rustc version is shown in the installation log, so even if rustc is not used, the version must be shown.
https://github.com/r-devel/r-svn/commit/6114d4126434c056b476cbc5db2657536c153d9a

I saw this warning at <https://github.com/eitsupi/r-glaredb/actions/runs/10879498825/job/30184319369>

```log
 ❯ checking Rust compilation ... WARNING
    No rustc version reported prior to compilation
```

In addition, remove the unnecessary flag added by #1212.

### DIFF
--- a/cleanup.win
+++ b/cleanup.win
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-. ./configure
+. ./cleanup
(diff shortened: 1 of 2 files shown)


etiennebacher:
Thanks

PR #1252, opened 2024-10-09
fix: Keeps Enum order when converting to R

### LINKED ISSUE #1251: Keep Enum order when convertir to R dataframe
When converting an enum series from Polars to R, the order is not always kept.

andyquinterom:
Closes #1251


etiennebacher:
Hi, thanks for opening the issue and the corresponding PR, can you add a test and a bullet point in NEWS?

andyquinterom:
@etiennebacher I added a test to validate many cases. 

```r
  # Generate 100 random data frames with random
  # orders for levels. They should always match
  for (i in 1:100) {
    expected_levels = sample(letters, length(letters))
    expected_values = sample(letters, length(letters))
    expect_identical(
      as_polars_series(expected_values)$
        cast(pl$Enum(expected_levels))$
        to_r(),
      factor(expected_values, levels = expected_levels)
    )
  }
  ```

**Comment by andyquinterom** [comment]
@etiennebacher Everything seems to pass except for R CMD check due to Non-api calls, but I think that's out of the scope of this pull request

File: tests/testthat/test-datatype.R [hunk 1]
```diff
@@ -224,12 +224,18 @@ test_that("contains_* functions for datatype work", {
...
 
 test_that("Enum", {
-  expect_identical(
-    as_polars_series(c("z", "z", "k", "a"))$
-      cast(pl$Enum(c("z", "k", "a")))$
-      to_r(),
-    factor(c("z", "z", "k", "a"))
-  )
+  # Generate 100 random data frames with random
+  # orders for levels. They should always match
```
eitsupi, line 228: Please use quickcheck or patrick instead of for loop.
https://www.etiennebacher.com/posts/2024-10-01-using-property-testing-in-r/

eitsupi (changes requested):

etiennebacher:
> Everything seems to pass except for R CMD check due to Non-api calls, but I think that's out of the scope of this pull request

Yes this is out of scope. I agree with @eitsupi about using property-based testing on this. I think something like this would do the job (but I haven't tested yet):

```r
test_that("reversing twice returns the original input", {
  for_all(
    values = any_atomic(any_na = TRUE),
	levels = any_atomic(any_na = TRUE),
    property = function(values, levels) {
	    expect_identical(
	      as_polars_series(values)$
	        cast(pl$Enum(levels))$
	        to_r(),
	      factor(values, levels = levels)
	    )
    }
  )
})
``` 
Please also add a bullet point in news with the PR number and your username (if you want).

DESCRIPTION
```diff
@@ -116,7 +117,7 @@ Collate:
     'sql.R'
     'vctrs.R'
     'zzz.R'
-Config/rextendr/version: 0.3.1
+Config/rextendr/version: 0.3.1.9000
```
eitsupi, line 120: Unrelated change.

NEWS.md
```diff
@@ -5,6 +5,7 @@
 ## Polars R Package 0.20.0
 
 - Updated rust-polars to 0.43.1 (#1230).
+- Maintains level order when converting Enums to factors (#1252, @andyquinterom)
```
eitsupi, line 8: I think you are adding it in the wrong place.

eitsupi (changes requested):
Thanks for the updates!

Could you bump the Rust lib version and please remove the lib-sums.tsv.
Please check the document for details.
https://github.com/pola-rs/r-polars/blob/8ecb1a358c5bf0c6c6e6c9527904bb798f07a101/DEVELOPMENT.md

NEWS.md
andyquinterom: Thanks for pointing this out

DESCRIPTION
andyquinterom: Done

andyquinterom:
> Thanks for the updates!
> 
> Could you bump the Rust lib version and please remove the lib-sums.tsv. Please check the document for details. https://github.com/pola-rs/r-polars/blob/8ecb1a358c5bf0c6c6e6c9527904bb798f07a101/DEVELOPMENT.md

I just ran the task and see that nothing has changed. Am I doing something wrong?

eitsupi:
We should update lib version by hand first.
https://github.com/pola-rs/r-polars/blob/main/src%2Frust%2FCargo.toml#L3-L3

etiennebacher (approved):
Thanks @andyquinterom

tests/testthat/test-datatype.R
```diff
@@ -224,11 +224,16 @@ test_that("contains_* functions for datatype work", {
 })
 
 test_that("Enum", {
-  expect_identical(
-    as_polars_series(c("z", "z", "k", "a"))$
-      cast(pl$Enum(c("z", "k", "a")))$
-      to_r(),
-    factor(c("z", "z", "k", "a"))
+  quickcheck::for_all(
+    factors = quickcheck::factor_(),
```
eitsupi, line 228: How about including `NA`?

src/rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "r-polars"
-version = "0.43.0"
+version = "0.44.0"
```
eitsupi, line 3:
```suggestion
version = "0.43.1"
```

DESCRIPTION
```diff
@@ -118,5 +119,5 @@ Collate:
     'zzz.R'
 Config/rextendr/version: 0.3.1
 VignetteBuilder: knitr
-Config/polars/LibVersion: 0.43.0
+Config/polars/LibVersion: 0.44.0
```
eitsupi, line 122:
```suggestion
Config/polars/LibVersion: 0.43.1
```

eitsupi (changes requested):
Almost looks good.
A few comments.

tests/testthat/test-datatype.R
eitsupi: I tried but there is an issue about that.
armcn/quickcheck#22

eitsupi (approved):
Thanks!

PR #1257, opened 2024-10-16
test: the latest nanoarrow supports utf8view type

eitsupi:
It seems that the test is starting to fail due to nanoarrow updates in R-multiverse.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Suggests:
     jsonlite,
     knitr,
     lubridate,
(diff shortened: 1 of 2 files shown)


eitsupi:
@shikokuchuo (cc @paleolimbot) This may be an interesting case arising from the lack of reverse dependency checking in the R-multiverse.

etiennebacher (approved):

eitsupi:
Sorry, I forgot to correct one part. It should now pass.

PR #1271, opened 2024-11-02
feat!: bump polars to 0.44.2

eitsupi:


eitsupi:
It seems that an error is occurring in the following location only on Windows.

https://github.com/pola-rs/r-polars/blob/a37b3c088a2012dbffcb630a091eaf031878e8a6/vignettes/userguide.Rmd#L229-L240

```log
Execution halted with the following contexts
   0: In R: in pl$read_csv():
   0: During function call [tools::buildVignettes(dir = ".", tangle = TRUE)]
      	could not parse `"George Read (American politician, born 1733)"
` as dtype `str` at column 'wikipedia_id
' (column number 36)
      The current offset in the file is 4916 bytes.
      You might want to try:
      - increasing `infer_schema_length` (e.g. `infer_schema_length=10000`),
      - specifying correct dtype with the `schema_overrides` argument
      - setting `ignore_errors` to `True`,
      - adding `"George Read (American politician, born 1733)"
` to the `null_values` list.
      Original error: ```invalid csv file
      Field `"George Read (American politician, born 1733)"
` is not properly escaped.```
```

The cell in the CSV in issue is now this, and a comma can be found there.

```csv
wikipedia_id
"George Read (American politician, born 1733)"
```

@etiennebacher Could you please check if you can reproduce the problem in Windows (R and Python)?

eitsupi:
I have tried the following and it works fine in Python🤔

```python
import ssl
import polars as pl
ssl._create_default_https_context = ssl._create_unverified_context
pl.read_csv("https://theunitedstates.io/congress-legislators/legislators-historical.csv")
```

Is it possible that there is a problem with the mechanism through which R downloads the csv file?

eitsupi:
I will open a follow-up issue and merge it in for now.

PR #1272, opened 2024-11-02
refactor!: remove `<Series>$compare()`

eitsupi:
Needed for #1271

This does not exist in Python Polars and can no longer be maintained since the Rust functions used internally have been removed.


R/series__series.R
```diff
@@ -475,57 +475,35 @@ Series_pow = function(exponent) {
...
-      o_len != 1 &&
-      s_len != 1
-  ) {
-    stop("in compare Series: not same length or either of length 1.")
-  }
-  .pr$Series$compare(self, as_polars_series(other), op) |>
-    unwrap(paste0("in $compare() with operator `", op, "`:"))
+#' @export
+`==.RPolarsSeries` = function(s1, s2) {
+  pl$select(pl$lit(s1)$eq(pl$lit(as_polars_series(s2))$cast(s1$dtype)))$to_series()
```
eitsupi, line 480: It should be noted that although cast is added to pass existing tests (presumably tested to mimic the behavior of Python Polars), this leads to some rather strange behavior. (pola-rs/polars#19603)

PR #1291, opened 2024-11-18
fix: Use `"curl"` to download CSV files on Windows

### LINKED ISSUE #1288: CSV reader issue on Windows with polars rs-0.44.2
It seems that an error is occurring in the following location only on Windows.

https://github.com/pola-rs/r-polars/blob/a37b3c088a2012dbffcb630a091eaf031878e8a6/vignettes/userguide.Rmd#L229-L240

```log
Execution halted with the following contexts
   0: In R: in pl$read_csv():
   0: During function call [tools::buildVignettes(dir = ".", tangle = TRUE)]
      	could not parse `"George Read (American politician, born 1733)"
` as dtype `str` at column 'wikipedia_id
' (column number 36)
      The current offset in the file is 4916 bytes.
      You might want to try:
      - increasing `infer_schema_length` (e.g. `infer_schema_length=10000`),
      - specifying correct dtype with the `schema_overrides` argument
      - setting `ignore_errors` to `True`,
      - adding `"George Read (American politician, born 1733)"
` to the `null_values` list.
      Original error: ```invalid csv file
      Field `"George Read (American politician, born 1733)"
` is not properly escaped.```
```

The cell in the CSV in issue is now this, and a comma can be found there.

```csv
wikipedia_id
"George Read (American politician, born 1733)"
```

@etiennebacher Could you please check if you can reproduce the problem in Windows (R and Python)?

Maybe related to this change: pola-rs/polars#19088

_Originally posted by @eitsupi in https://github.com/pola-rs/r-polars/issues/1271#issuecomment-2480923761_

etiennebacher:
Close #1288.

Might be a bit flaky, I used the first method that worked on this example.

Btw, we have a warning in GHA about macos-12 being deprecated, maybe we can just remove it?

### DIFF
--- a/R/io_csv.R
+++ b/R/io_csv.R
@@ -224,7 +224,15 @@ check_is_link = function(path, reuse_downloaded, raise_error = FALSE) {
         cache_temp_file[[actual_url]] = tempfile()
       }
       if (isFALSE(reuse_downloaded) || isFALSE(file.exists(cache_temp_file[[actual_url]]))) {
-        download.file(url = actual_url, destfile = cache_temp_file[[actual_url]])
+        # TODO: py-polars doesn't error with the URL in the issue. We should
+        # see how they handle URLs.
+        # https://github.com/pola-rs/r-polars/issues/1288
+        curl_available = nzchar(Sys.which("curl"))
+        if (tolower(Sys.info()[["sysname"]]) == "windows" && curl_available) {
+          download.file(url = actual_url, destfile = cache_temp_file[[actual_url]], method = "curl")
+        } else {
+          download.file(url = actual_url, destfile = cache_temp_file[[actual_url]])
+        }
         message(paste("tmp file placed in \n", cache_temp_file[[actual_url]]))
       }
 


R/io_csv.R
```diff
@@ -224,7 +224,15 @@ check_is_link = function(path, reuse_downloaded, raise_error = FALSE) {
         cache_temp_file[[actual_url]] = tempfile()
       }
       if (isFALSE(reuse_downloaded) || isFALSE(file.exists(cache_temp_file[[actual_url]]))) {
-        download.file(url = actual_url, destfile = cache_temp_file[[actual_url]])
+        # TODO: py-polars doesn't error with the URL in the issue. We should
+        # see how they handle URLs.
+        # https://github.com/pola-rs/r-polars/issues/1288
+        curl_available = nzchar(Sys.which("curl"))
```
etiennebacher, line 230: Recommended in the "Good practice" section of `?download.file()`

eitsupi (commented):
Thanks for looking into this. But from the CI results it looks like the problem is not solved?

My understanding is that curl is used by default on Windows since R 4.2, so I am not sure if it makes sense to specify explicitly.

eitsupi:
> we have a warning in GHA about macos-12 being deprecated, maybe we can just remove it?

Some tests are still being run for Intel mac, so I have simply replaced them with macos-13. (#1293)

etiennebacher:
Looks like #1292 is a better fix

PR #1316, opened 2025-01-23
ci: ensure install R on ubuntu runner, and other minor fixes to make ci passed

eitsupi:


eitsupi:
Currently, the CI runner is being updated and the URL in the document is being updated at the same time, and I don't think it will pass the test unless both are resolved. So I have updated the document here at the same time.

PR #1326, opened 2025-02-13
test: fix test case broken by the new version of `patrick`

eitsupi:


eitsupi:
I think we need a new release that includes this to keep polars alive in the R-multiverse Production repository, so I will merge it and create a new release immediately.

PR #1331, opened 2025-02-27
docs: Add warning about maintenance mode in README

etiennebacher:
Might be a bit late but I just realized that the warning message was only displayed in the changelog and releases page, and I doubt many people check those.

So even if it's for a few weeks only, I'd like to show that the package is not abandoned.


eitsupi (approved):
👍

PR #1337, opened 2025-03-13
fix: fix arguments that were erroneously passed outside of `unwrap()`

eitsupi:
It appears that `unlist()` is failing on R 4.5 due to incorrectly positioned arguments.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,11 @@
 
 ## polars (development version)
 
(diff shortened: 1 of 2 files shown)


eitsupi:
Without this fix, many operations will fail in the upcoming R 4.5 release so I am merging it.

etiennebacher:
Thanks!

PR #1354, opened 2025-03-29
ci: stop using `filter_rcmdcheck.R` to simplify maintenance

eitsupi:
There is no active development on this package, so there is no need to be aware of new NOTEs anymore.

Context: https://github.com/pola-rs/r-polars/issues/80#issuecomment-1475223652


eitsupi:
The expected failures have correctly occurred, so merging.

Issue #1336, opened 2025-03-12
Change the default branch / rename the package


eitsupi:
Related to #1152 and eitsupi/neo-r-polars#199

Currently, the repository has two main branches with completely different histories.

- `main` (Maintenance mode default branch, continued from https://github.com/sorhawell/minipolars)
- `next` (Active development branch, mirror of https://github.com/eitsupi/neo-r-polars)

In R 4.5 (will be released April-2025), the `polars` package built from the current `main` branch has been found not to pass the test and will need to be replaced in the near future by the `neopolars` package on the `next` branch.

So I want to change the default branch, rename the two branches, and rename the packages on the two branches. (Since the two packages can exchange data via the Arrow C stream interface, it would be useful if they could coexist.)

1. ~~Change the default branch to `next`.~~
2. ~~Rename current `main` to `mini`, and current `next` to `main`.~~
3. ~~Rename current `polars` to `minipolars`, rename current `neopolars` to `polars`.~~

Edit 2025-05-16:

1. ~~Rename current `main` to `v0`, and current `next` to `main`.~~
2. ~~Complete the rewriting of package.~~
3. ~~Change the default branch from `v0` to `main`~~
4. ~~Rename current `neopolars` to `polars`~~
5. ~~Release `polars` 1.0.0~~
6. ~~Rename current `polars` to `minipolars`~~

~~We can't do this immediately because the CI setup is still incomplete on the next branch and the package is not fully functional, but it's time to start making plans.~~

Edit 2025-06-17:

1. Rename current `main` to `v0`. (Done in 2025-06-01)
2. Move `v0`'s home to <https://github.com/rpolars/r-polars0> and rename current `polars` to `polars0`. (Done in 2025-06-07)
3. Change the default branch from `v0` to `next`. (Done in 2025-06-17)
4. Rename current `neopolars` to `polars`. (Done in 2025-06-23)
5. Complete the rewriting of package. (Done in 2025-07-17)
6. Release `polars` 1.0.0. (Done in 2025-07-21)
7. Rename current `next` to `main` and finish moving from <https://github.com/eitsupi/neo-r-polars> to here. (Done in 2025-07-21)

etiennebacher:
> Rename current polars to minipolars, rename current neopolars to polars.

I understand that the original implementation was `minipolars`, but I think the current `polars` should be renamed `old-polars` so that it's clear it's not maintained anymore.

[...]

PR #1358, opened 2025-03-29
chore: switch the maintainer of this package

eitsupi:
I expect the next release to be the last release of this package and would like to clarify support responsibilities by crediting me as the maintainer instead of @sorhawell who is no longer active.
(Just for the record, this was the case in the polars package on CRAN, but I have revert to the original because we gave up on the CRAN release.)
https://github.com/cran/polars/blob/a4763e6c99a7985d3272691535769d0fd529faf5/DESCRIPTION#L6-L10

Related to #1152 and #1336


etiennebacher (approved):
Ok for me

PR #1435, opened 2025-07-21
ci(dependabot): Use options for cargo deps updates

eitsupi:


eitsupi:
`cooldown` seems very recent feature and does not seem supported by the linter.
https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/

PR #1442, opened 2025-07-21
feat: Basic `[[` method for dataframe

### LINKED ISSUE #1441: Add `[[` method for `polars_data_frame`
I came across this (exploring why mirai's row-wise multiple map is currently failing for polars objects):

``` r
mtcars[[1]]
#>  [1] 21.0 21.0 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 17.8 16.4 17.3 15.2 10.4
#> [16] 10.4 14.7 32.4 30.4 33.9 21.5 15.5 15.2 13.3 19.2 27.3 26.0 30.4 15.8 19.7
#> [31] 15.0 21.4

data.table::as.data.table(mtcars)[[1]]
#>  [1] 21.0 21.0 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 17.8 16.4 17.3 15.2 10.4
#> [16] 10.4 14.7 32.4 30.4 33.9 21.5 15.5 15.2 13.3 19.2 27.3 26.0 30.4 15.8 19.7
#> [31] 15.0 21.4

polars::as_polars_df(mtcars)[[1]]
#> Error in polars::as_polars_df(mtcars)[[1]]: wrong arguments for subsetting an environment
```

<sup>Created on 2025-07-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

It would perhaps make sense to add a `[[` method for polars data frames?

(note also that in base R `mtcars[[1]]` is an order of magnitude faster than `mtcars[, 1, drop = TRUE]`)

eitsupi:
Close #1441

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 ### New features
 
-- polars DataFrame can be subset with `[[` (#1442).
+- `[[` can be used to extract a column from a polars DataFrame as a Series (#1442).
 
 ## polars 1.0.0
 


R/dataframe-s3-base.R
```diff
@@ -524,3 +525,32 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     x
   }
 }
+
+#' @export
+`[[.polars_data_frame` <- function(x, i, ...) {
+  dots <- list2(...)
+  if (length(dots) > 0L) {
+    warn(
+      c(`!` = "Subsetting a polars DataFrame with `[[` ignores other than the first argument `i`.")
```
etiennebacher:
```suggestion
      c(`!` = "Subsetting a polars DataFrame with `[[` ignores arguments other than the first one (`i`).")
```

R/dataframe-s3-base.R
```diff
@@ -524,3 +525,32 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
...
+    )
+  }
+
+  if (is_string(i)) {
+    tryCatch(
+      x$get_column(i),
+      error = function(e) NULL
+    )
+  } else if (is_scalar_integerish(i, finite = TRUE) && i >= 1L) {
```
shikokuchuo:
```suggestion
  } else if (is_scalar_integerish(i)) {
```
We can just allow the normal error if `i` is not within range? I know this opens up the use of negative values, but we don't need to restrict that 'feature' here?

R/dataframe-s3-base.R
```diff
@@ -524,3 +525,32 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
...
+    x$select(pl$nth(i - 1L))$to_series()
+  } else {
+    abort(
+      c(
+        `!` = "Can't subset a polars DataFrame with `[[`.",
+        i = sprintf(
+          "Subscript `%s` must be a string or a positive scalar integer.",
```
shikokuchuo:
```suggestion
          "Subscript `%s` must be a string or a scalar integer.",
```
Depending on my suggestion above.

R/dataframe-s3-base.R
```diff
@@ -524,3 +525,32 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     x
   }
 }
+
+#' @export
+`[[.polars_data_frame` <- function(x, i, ...) {
+  dots <- list2(...)
+  if (length(dots) > 0L) {
```
shikokuchuo:
```suggestion
  if (...length()) {
```
Leaves dots unevaluated.

R/dataframe-s3-base.R
```diff
@@ -524,3 +525,32 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     x
   }
 }
+
+#' @export
+`[[.polars_data_frame` <- function(x, i, ...) {
+  dots <- list2(...)
```
shikokuchuo:
```suggestion

```

R/dataframe-s3-base.R
eitsupi: Thanks, indeed, there doesn't appear to be any major issues with accepting negative values here, so that is allowed for now.

tests/testthat/test-dataframe-s3-base.R
```diff
@@ -287,3 +287,52 @@ test_that("Special cases of `[` behavior", {
   # TODO: test behavior of x[i, , drop = TRUE] when it is clarified
   # https://github.com/tidyverse/tibble/issues/1570
 })
+
+patrick::with_parameters_test_that(
+  "Column subsetting with `[[` works with {rlang::quo_text(value)}",
+  .cases = {
+    series_a <- pl$Series("a", 1L)
+    series_b <- pl$Series("b", 2L)
```
etiennebacher: `series_b` is unused

eitsupi:
```suggestion
```

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## polars (development version)
 
+### New features
+
+- polars DataFrame can be subset with `[[` (#1442).
```
etiennebacher: subsetted?

tests/testthat/_snaps/dataframe-s3-base.md
```diff
@@ -969,3 +969,126 @@
...
+
+    Code
+      dat[[value]]
+    Condition
+      Error in `x$select()`:
+      ! Evaluation failed in `$select()`.
+      Caused by error:
+      ! Evaluation failed in `$collect()`.
+      Caused by error:
+      ! Column(s) not found: nth
```
etiennebacher, line 1021: That's a weird error message, can we improve it? (same with value = 4)

eitsupi: This uses the error from the `nth` as is, so it would probably be better to improve the nth error message upstream.

etiennebacher: I'll open an issue

Edit: already reported in https://github.com/pola-rs/polars/issues/22705, we'll see if that's fixed in the next release

R/dataframe-s3-base.R
```diff
@@ -184,6 +184,7 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
 # Try to match `tibble` behavior as much as possible, following
 # https://tibble.tidyverse.org/articles/invariants.html#column-subsetting
 # TODO: add document
+# TODO: move the implementation to polars_subset or something similar
```
etiennebacher, line 187: I don't mind keeping all S3 methods in the same file, it avoids having methods scattered here and there.

etiennebacher (approved):
Thanks, LGTM

R/dataframe-s3-base.R
etiennebacher: Or maybe I misunderstood what you mean by `polars_subset`

eitsupi: Ah, sorry.
I think this is the same strategy as tibble, I think moving the actual implementation into its own function makes it easier to maintain, especially after S7 when we can use double dispatch for `[`.

etiennebacher: Ah ok, thanks for the clarification

Issue #1440, opened 2025-07-21
Add vignette map example using native serialization


shikokuchuo:
The custom functions vignette examples currently all involve converting to and from R objects.

With polars using mirai's serialization registry, this is not always going to be necessary.

It might be an idea to add some examples of how users can take advantage of this, e.g.:

```r
library(polars)
library(purrr)
library(mirai)
daemons(6)

dat <- as_polars_df(mtcars)
dat |>
  as.list() |>
  map(in_parallel(\(x) x$sort())) |>
  as_polars_df()
```

There's no conversion to R types anywhere in this (you can verify line by line that the result is a polars object).

Not the best example, so can I leave it with you to find a more natural one?

Also happy to contribute something, just let me know.

etiennebacher:
> The custom functions vignette examples currently all involve converting to and from R objects.
> With polars using mirai's serialization registry, this is not always going to be necessary.

> Lines 126-127, it's because scale() only works on R vectors that we need to do the conversion. 

> otherwise doing some lengthy calculation which actually uses the native format,

I think I'm missing something regarding native serialization. The main usecase is to be able to send only some of the Series to a daemon, as shown in the vignette, but not to apply polars functions on this Series. To me, if the custom function you want to apply can be written entirely with polars syntax, then there's no point using mirai or purrr for parallel processing. You should just write a custom function and use this in `with_columns()`, `select()`, etc. and polars will take care of optimizing the queries, for instance by doing common subplan elimination. 

It seems to me that using purrr or mirai for parallel processing should be a last resort that happens when your code cannot be translated into polars syntax, for instance because it comes from an external package. We took the example of `scale()`, which would be simple enough to translate, but this could be any function from any R package. If the function in question doesn't accept polars Series (like the vast majority of functions), then we have to convert to R types.

> One powerful use of daemon is to send lazyframes to other computers for aggregation

Would be great to have some usecases like this!

eitsupi:
I think it would be better to add the documentation to integration-mirai instead of to create a new the vignette.

Just as a note, if we load polars after creating daemons like this, we'll get a strange error because `mirai::register_serial` doesn't propagate to the already created daemons.
It's probably worth checking to see if the daemons are already created and raising a warning:

``` r
mirai::daemons(1)
#> [1] 1

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
[...]

PR #1451, opened 2025-07-24
fix:  Warn for existing mirai daemons when loaded

eitsupi:
Related to https://github.com/pola-rs/r-polars/issues/1440#issuecomment-3112254174

It will raise a warning before the weird errors occur.

``` r
mirai::daemons(1)
#> [1] 1

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)
#> Warning: ! Auto registered mirai serialization configs by polars does not affect
#>   existing daemons.
#> ℹ To apply the configs, recreating daemons is needed.

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
#> 'miraiError' chr Error in x$len(): Evaluation failed in `$len()`.
```

<sup>Created on 2025-07-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

If polars is loaded before the daemons are created, nothing will happen.

``` r
series <- polars::as_polars_series(1)$cast(polars::pl$Int128)

mirai::daemons(1)
#> [1] 1

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
#> shape: (1, 1)
#> ┌──────┐
#> │      │
#> │ ---  │
#> │ i128 │
#> ╞══════╡
#> │ 2    │
#> └──────┘
```

<sup>Created on 2025-07-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

``` r
mirai::daemons(1)
#> [1] 1
mirai::daemons(0)
#> [1] 0

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)

mirai::daemons(1)
#> [1] 1

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
#> shape: (1, 1)
#> ┌──────┐
#> │      │
#> │ ---  │
#> │ i128 │
#> ╞══════╡
#> │ 2    │
#> └──────┘
```

<sup>Created on 2025-07-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

@shikokuchuo Could you take a look at this?
Maybe it would be better to add such functionality to `register_serial`.

### DIFF
--- a/R/integration-mirai.R
+++ b/R/integration-mirai.R
@@ -5,6 +5,25 @@ register_mirai_serial <- function() {
   # If the mirai package is not installed, `asNamespace("mirai")` will throw an error,
   # so this function should only be called when the mirai package is loaded.
   if (exists("register_serial", envir = asNamespace("mirai"), mode = "function")) {
+    # If the daemons are already set, registering serialization configs
+    # will not affect existing daemons. So warn to the user.
+    # Safety: mirai::daemons_set is added in mirai 2.3.0,
+    # the same version that introduced the `register_serial` function.
+    if (mirai::daemons_set()) {
+      warn(
+        format_warning(
+          c(
+            `!` = sprintf(
+              "Auto registered %s serialization configs by %s does not affect existing daemons.",
+              format_pkg("mirai"),
+              format_pkg("polars")
+            ),
+            i = "To apply the configs, recreating daemons is needed."
+          )
+        )
+      )
+    }
+
     mirai::register_serial(
       c("polars_data_frame", "polars_lazy_frame", "polars_series"),
       sfunc = list(\(x) x$serialize(), \(x) x$serialize(), \(x) x$serialize()),


shikokuchuo:
I would avoid throwing a warning on a package load event, as this can be unexpected for the user (vs. an explicit event like attaching a package interactively).

It's tricky but the only failsafe solution seems to be to require users to load polars before calling `mirai::daemons()`. Note that any packages which import polars would automatically satisfy this.

eitsupi:
Thanks.
I agree that it's not ideal to have a warning when loading polars.

On the other hand, I think it's not good to have a non-intuitive error when you're saying things like "users can use it like a normal R object without worrying about serialization."
How about just a message instead of a warning?

shikokuchuo:
Agreed, a message is probably enough. And perhaps it's not too bad having this on load.

But maybe also expose a no-op function that can be called:
```r
register_serial <- function() invisible()
```
The warning would still be generated on load, but then at least it makes sense if it originates from there? Someone can call it to be safe. Just a thought - not sure if it makes sense...

R/integration-mirai.R
```diff
@@ -5,6 +5,25 @@ register_mirai_serial <- function() {
...
+      warn(
+        format_warning(
+          c(
+            `!` = sprintf(
+              "Automatically registered %s serialization configs by %s does not affect existing daemons.", # nolint: line_length_linter
+              format_pkg("mirai"),
+              format_pkg("polars")
+            ),
+            i = "To apply the configs, recreating daemons is needed."
+          )
```
etiennebacher: Can this be clarified a bit? To re-use your example:

```r
mirai::daemons(1)

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)

# <<<<<<<<<<< should I put `mirai::daemons(1)` again here?
# <<<<<<<<<<< or should I put `mirai::daemons(0)` and then `mirai::daemons(1)`?

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
```

Maybe something along those lines?

```
i `polars` was attached to the environment after using `mirai::daemons()`
i The automatic serialization made by mirai is not applied to existing daemons.
i Reset the daemons with <code>.
```

eitsupi:
Thanks both, how about this?

``` r
mirai::daemons(1)
#> [1] 1

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)
#> ℹ The polars package was loaded after mirai daemons were already created.
#> ℹ To apply the serialization configs registered by polars, recreate `daemons.`
#> • Run `mirai::daemons(0)` to reset daemon connections, then recreate daemons
#>   with `mirai::daemons()`.

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
#> 'miraiError' chr Error in x$len(): Evaluation failed in `$len()`.
```

<sup>Created on 2025-07-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/integration-mirai.R
```diff
@@ -5,6 +5,33 @@ register_mirai_serial <- function() {
...
+      inform(
+        format_message(
+          c(
+            i = sprintf(
+              "The %s package was loaded after %s daemons were already created.",
+              format_pkg("polars"),
+              format_pkg("mirai")
+            ),
+            i = sprintf(
+              "To apply the serialization configs registered by %s, recreate daemons.",
```
etiennebacher:
```suggestion
              "To apply the serialization config registered by %s, recreate daemons.",
```

R/integration-mirai.R
```diff
@@ -5,6 +5,33 @@ register_mirai_serial <- function() {
...
+              format_pkg("polars")
+            ),
+            `*` = sprintf(
+              "Run %s to reset daemon connections, then recreate daemons with %s.",
+              format_code("mirai::daemons(0)"),
+              format_fn("mirai::daemons")
+            )
```
etiennebacher: Should we also suggest loading the package earlier? Something like:
```
i daemons already created, etc.
i either load polars earlier in the script or reset daemons
i reset daemons with etc.
```

eitsupi:
> But maybe also expose a no-op function that can be called:
> 
> ```r
> register_serial <- function() invisible()
> ```
> 
> The warning would still be generated on load, but then at least it makes sense if it originates from there? Someone can call it to be safe. Just a thought - not sure if it makes sense...

Do you mean export the `register_mirai_serial` function?
It's definitely possible, but as you say this is done automatically when mirai loads so I don't think it makes sense.

tests/testthat/_snaps/integration-mirai.md
```diff
@@ -0,0 +1,7 @@
+# Warn if daemons already exist when registering mirai serialization configs
+
+    Code
+      register_mirai_serial()
+    Message <rlang_message>
+      i The polars package was loaded after mirai daemons were already created. i To apply the serialization configs registered by polars, recreate daemons. * Run `mirai::daemons(0)` to reset daemon connections, then recreate daemons with `mirai::daemons()`.
```
etiennebacher: Do you know why we don't have line breaks here?

etiennebacher (approved):
LGTM, thanks! Just some minor comments

tests/testthat/_snaps/integration-mirai.md
eitsupi: I don't know. There are line breaks when it actually gets printed to the console...

R/integration-mirai.R
```diff
@@ -5,6 +5,38 @@ register_mirai_serial <- function() {
...
+              "To apply the serialization config registered by %s, recreate daemons.",
+              format_pkg("polars")
+            ),
+            `*` = sprintf(
+              "Run %s to reset daemon connections, then recreate daemons with %s.",
+              format_code("mirai::daemons(0)"),
+              format_fn("mirai::daemons")
+            ),
+            i = sprintf(
+              "Suppress this message by loading %s before creating daemons (or loading %s.)",
```
etiennebacher:
```suggestion
              "Suppress this message by loading %s before creating daemons (or loading %s).",
```

eitsupi:
> No I didn't mean that, but then I'm not sure my suggestion made much sense either - I was thinking of a dummy function that just loads polars - so not forcing people to call `library(polars)`.

I see, but in that case `requireNamespace` etc. isn't enough?

``` r
requireNamespace("polars", quietly = TRUE)

mirai::daemons(1)
#> [1] 1

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
#> shape: (1, 1)
#> ┌──────┐
#> │      │
#> │ ---  │
#> │ i128 │
#> ╞══════╡
#> │ 2    │
#> └──────┘
```

<sup>Created on 2025-07-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

``` r
rlang::check_installed("polars")

mirai::daemons(1)
#> [1] 1

series <- polars::as_polars_series(1)$cast(polars::pl$Int128)

list(series) |>
  mirai::mirai_map(\(x) x * 2L) |>
  _[][[1]]
#> shape: (1, 1)
#> ┌──────┐
#> │      │
#> │ ---  │
#> │ i128 │
#> ╞══════╡
#> │ 2    │
#> └──────┘
```

<sup>Created on 2025-07-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

eitsupi:
Let's merge it for now and see what happens.

Issue #1397, opened 2025-06-18
Functions to implement after rewrite


etiennebacher:
Once the rewritten package is definitely moved here, we'll be able to focus on implementing relatively new functions:

- Expr:
  - [ ] index_of(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.index_of.html#polars.Expr.index_of
  - [ ] bin.reinterpret(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bin.reinterpret.html#polars.Expr.bin.reinterpret
  - [ ] cat.starts_with(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.cat.starts_with.html
  - [ ] cat.ends_with(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.cat.ends_with.html
  - [ ] cat.len_bytes(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.cat.len_bytes.html#polars.Expr.cat.len_bytes
  - [ ] cat.len_chars(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.cat.len_chars.html#polars.Expr.cat.len_chars
  - [ ] dt.millenium(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.dt.millennium.html#polars.Expr.dt.millennium
  - [ ] dt.replace(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.dt.replace.html#polars.Expr.dt.replace
  - [ ] meta.is_literal(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.meta.is_literal.html#polars.Expr.meta.is_literal
  - [ ] str.escape_regex(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.str.escape_regex.html#polars.Expr.str.escape_regex
  - [ ] str.find_many(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.str.find_many.html#polars.Expr.str.find_many
  - [ ] str.normalize(): https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.str.normalize.html#polars.Expr.str.normalize
- Datatype:
  - [ ] pl.dtype_of(): https://docs.pola.rs/api/python/dev/reference/api/polars.dtype_of.html 
[...]

PR #1474, opened 2025-08-03
release: version 1.1.0

eitsupi:


NEWS.md
```diff
@@ -31,16 +45,75 @@ This is an update that corresponds to Python Polars 1.32.0, which includes signi
...
   - The `...` argument (dynamic dots) of `<lazyframe>$drop_nulls()` and `<dataframe>$drop_nulls()`.
   - The `...` argument (dynamic dots) of `<lazyframe>$drop_nans()` and `<dataframe>$drop_nans()`.
-  - The `...` argument (dynamic dots) of `<lazyframe>$unnest()` and `<dataframe>$unnest()`.
-  - The `...` argument (dynamic dots) of `<lazyframe>$explode()` and `<dataframe>$explode()`.
   - The `...` argument (dynamic dots) of `<lazyframe>$unique()` and `<dataframe>$unique()` (#1463).
+  - The `...` argument (dynamic dots) of `<lazyframe>$drop()` and `<dataframe>$drop()`.
+  - The `...` argument (dynamic dots) of `<lazyframe>$explode()` and `<dataframe>$explode()`.
+  - The `...` argument (dynamic dots) of `<lazyframe>$unnest()` and `<dataframe>$unnest()`.
   - The `on`, `index`, and `values` arguments of `<dataframe>$pivot()`.
   - The `on` and `index` arguments of `<lazyframe>$unpivot()` and `<dataframe>$unpivot()`.
```
etiennebacher: Would rather have something like:

```
This behavior change is visible in the following functions:
  - `<dataframe>$to_dummies()`
  - etc.

etiennebacher (commented):
Thank you for all the work, that looks very nice!

Apart from a minor comment in NEWS, my main question is: why do we want to release so early? 1.0.0 was released only two weeks ago and I think we could add a few items from #1397 before the next version. For users, it would also be easier to follow.

eitsupi:
Since upstream polars may be released about once a week, I think the basic policy is to release downstream as soon as it catches up with the upstream. Otherwise it will quickly become outdated.

This time, I had to absorb a month's worth of differences in one PR, so I've been struggling to catch up for the past two weeks.
Ideally, we should always be synchronized with the upstream main, so the main branch is stable for a very short period of time.
We don't have to release it now, but the main branch may soon be updated to the development version of upstream, so even if you add new features to the main branch, we can't release as 1.1.0.

etiennebacher:
It seems that they slowed down a bit the release pace, which is good because releasing once a week is too much IMO (especially in the R world where CRAN packages should wait at least a couple of months between each release).

If we try to match the python polars releases, would it also make sense to follow their version numbers? It would be convenient to get python polars and R polars 1.32.0 for instance.

>  We don't have to release it now, but the main branch may soon be updated to the development version of upstream, so even if you add new features to the main branch, we can't release as 1.1.0.

I don't understand this point. If we don't release 1.1.0, what prevents adding new features on main and release 1.1.0 later?

NEWS.md
eitsupi: How about 288ca6149fb3b7ba8a00a8c8325951ea14918f3e?

eitsupi:
As an additional note, due to an upstream bug I was unable to merge a PR based on py-1.30.0 into the main branch, forcing me to wait until py-1.31.0.
https://github.com/eitsupi/neo-r-polars/pull/360#issuecomment-2906835136

Such huge differences increase the effort required for updates, so I am keenly aware of the need to frequently catch up with upstream changes in unreleased versions to find bugs.

etiennebacher:
> The advantage of having separate minor versions is that it's easier to justify feature releases on the R side.

Right, I hadn't thought about this. On the other hand, I don't see many cases where we would like to release a new feature in between two "standard" releases (i.e. releases to catch up with Python polars).

PR #1479, opened 2025-08-06
feat: Implement `$str$normalize()`

etiennebacher:
Part of #1397


etiennebacher:
@eitsupi FYI there's no protection to prevent merging before review, I could merge right now. I think we had this before, do you know why it's no longer active?

PR #1481, opened 2025-08-10
perf: Speed up when all selectors in dots are strings

### LINKED ISSUE #1477: R session crash due to `parse_into_selector()`
`parse_into_selector()` becomes quite slow when thousands of columns are passed to it, and sometimes causes the R session to crash. This is for instance visible in `$unpivot()`, where it may happen that we want to reshape thousands of columns into two or three only.

This crashes R:

```r
library(polars)
  
long_list <- rep(list(1), 3000)
names(long_list) <- paste0("x", 1:length(long_list))
long_list[["index"]] <- "hi"
dat <- pl$DataFrame(!!!long_list)

dat$unpivot(on = names(long_list), index = "index")
```

This worked fine in old r-polars:

``` r
library(polars0)

long_list <- rep(list(1), 3000)
names(long_list) <- paste0("x", 1:length(long_list))

long_list[["index"]] <- "hi"
dat <- pl$DataFrame(long_list)

dat$unpivot(on = names(long_list), index = "index")
#> shape: (3_001, 3)
#> ┌───────┬──────────┬───────┐
#> │ index ┆ variable ┆ value │
#> │ ---   ┆ ---      ┆ ---   │
#> │ str   ┆ str      ┆ str   │
#> ╞═══════╪══════════╪═══════╡
#> │ hi    ┆ x1       ┆ 1.0   │
#> │ hi    ┆ x2       ┆ 1.0   │
#> │ hi    ┆ x3       ┆ 1.0   │
#> │ hi    ┆ x4       ┆ 1.0   │
#> │ hi    ┆ x5       ┆ 1.0   │
#> │ …     ┆ …        ┆ …     │
#> │ hi    ┆ x2997    ┆ 1.0   │
#> │ hi    ┆ x2998    ┆ 1.0   │
#> │ hi    ┆ x2999    ┆ 1.0   │
#> │ hi    ┆ x3000    ┆ 1.0   │
#> │ hi    ┆ index    ┆ hi    │
#> └───────┴──────────┴───────┘
```

And in Python polars:

```python
import polars as pl
import polars.selectors as cs

large_dict = {f"key_{i}": 1 for i in range(3000)}
df = pl.from_dict(large_dict)

df.unpivot(large_dict.keys(), index="key_1")
```
```
shape: (3_000, 3)
┌───────┬──────────┬───────┐
│ key_1 ┆ variable ┆ value │
│ ---   ┆ ---      ┆ ---   │
│ i64   ┆ str      ┆ i64   │
╞═══════╪══════════╪═══════╡
│ 1     ┆ key_0    ┆ 1     │
│ 1     ┆ key_1    ┆ 1     │
│ 1     ┆ key_2    ┆ 1     │
│ 1     ┆ key_3    ┆ 1     │
│ 1     ┆ key_4    ┆ 1     │
│ …     ┆ …        ┆ …     │
│ 1     ┆ key_2995 ┆ 1     │
│ 1     ┆ key_2996 ┆ 1     │
│ 1     ┆ key_2997 ┆ 1     │
│ 1     ┆ key_2998 ┆ 1     │
│ 1     ┆ key_2999 ┆ 1     │
└───────┴──────────┴───────┘
```

Profiling on the same example with 2000 columns (to avoid crash) shows that `parse_into_selector()` takes most of the time, and it was introduced in 1.1.0 so it's likely the culprit:

Image

eitsupi:
Fix #1477

### DIFF
--- a/R/utils-parse-expr.R
+++ b/R/utils-parse-expr.R
@@ -38,8 +38,13 @@ parse_into_selector <- function(..., .strict = TRUE, .arg_name = "...") {
   dots <- list2(...)
 
   try_fetch(
-    lapply(dots, \(x) as_polars_selector(x, strict = .strict)) |>
-      Reduce(`|`, x = _),
+    # If all elements are single strings, treat as a single character vector for shortcut
+    if (is_list_of_string(dots)) {
+      as_polars_selector.character(as.character(dots), strict = .strict)
+    } else {
+      lapply(dots, \(x) as_polars_selector(x, strict = .strict)) |>
+        Reduce(`|`, x = _)
+    },
     error = function(cnd) {
       abort(
         format_error(
--- a/R/utils-parse-expr.R
+++ b/R/utils-parse-expr.R
@@ -39,7 +39,7 @@ parse_into_selector <- function(..., .strict = TRUE, .arg_name = "...") {
 
   try_fetch(
     # If all elements are single strings, treat as a single character vector for shortcut
-    if (is_list_of_string(dots)) {
+    if (length(dots) > 0L && is_list_of_string(dots)) {
       as_polars_selector.character(as.character(dots), strict = .strict)
     } else {
       lapply(dots, \(x) as_polars_selector(x, strict = .strict)) |>


etiennebacher (changes requested):
Thanks, have you already benchmarked this? Otherwise I can do it on the test data where I noticed that.

Also, can you add an item in NEWS?

eitsupi:
> Thanks, have you already benchmarked this?

Yes, I tested 9eb758e and found it to be about 1.3x faster than as_polars_selector in https://github.com/pola-rs/r-polars/issues/1477#issuecomment-3172462902 case, which means 100x faster than main.

This is a fairly nerdy internal change, so I don't think it necessarily needs to be included in the NEWS. If necessary, I will add it before the release.

PR #1489, opened 2025-08-10
feat: Implement `pl$linear_space()` and `pl$linear_spaces()`

etiennebacher:
Part of #1397


R/functions-range-linear_space.R
```diff
@@ -0,0 +1,102 @@
...
+  num_samples,
+  ...,
+  closed = c("both", "left", "none", "right"),
+  as_array = FALSE
+) {
+  wrap({
+    check_dots_empty0(...)
+    if (is_integerish(num_samples)) {
+      num_samples <- as.integer(num_samples)
+    }
```
etiennebacher, line 97: Passing `num_samples = 3` would error by default, but I think it's a better user experience to allow it.

tests/testthat/test-functions-range-linear_space.R
eitsupi, line 1: Could you use parameterized tests?

eitsupi (changes requested):

R/functions-range-linear_space.R
eitsupi: Could you add a comment about that?

etiennebacher: done in [`95f3d2a` (#1489)](https://github.com/pola-rs/r-polars/pull/1489/commits/95f3d2ac96a103f2982824c268d5e3fd4f2bcfde)

tests/testthat/test-functions-range-linear_space.R
etiennebacher: I've added some in [653d5da](https://github.com/pola-rs/r-polars/pull/1489/commits/653d5dae4932f989257fca6486d908be3885688b) for `closed` but I think adding more for the existing tests would make them less readable.

PR #1515, opened 2025-08-23
chore: Use mold as the linker if available

### LINKED ISSUE #1508: Use mold if available
I found that using <https://github.com/rui314/mold> as the linker improved the developer experience.
Without mold, I often struggled build error "collect2: fatal error: ld terminated with signal 9 [Killed]".

Add `-fuse-ld=mold` to this line.

https://github.com/pola-rs/r-polars/blob/83f362a18389334b246f542313a2816f8f21c73e/src/Makevars.in#L9

eitsupi:
Close #1508

### DIFF
--- a/configure
+++ b/configure
@@ -243,6 +243,7 @@ if [ "${DEBUG}" = "true" ]; then
   PROFILE=dev
 fi
 
+check_mold
 check_emscripten
 check_bin_lib
 check_cargo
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -197,6 +204,40 @@ EOF
...
 
+# Simplified mold check: use single env var USE_MOLD and append flags
+check_mold() {
+  if [ "${USE_MOLD}" != "true" ]; then
+    return
+  fi
+
+  if command -v mold >/dev/null 2>&1; then
+    add_additional_lib "-fuse-ld=mold"
+    export RUSTFLAGS="${RUSTFLAGS:-} -C link-args=-fuse-ld=mold"
```
eitsupi, line 215: I'm not sure this is useful here.

PR #1496, opened 2025-08-11
perf: Skip check when get the first value of the series

eitsupi:

PR #1517, opened 2025-08-23
perf: Rewrite `as_polars_expr.default` in Rust

eitsupi:
Supersedes #1496

Before:

```r
bench::mark(
  lit_vec_single = pl$lit(1L),
  lit_vec_long = pl$lit(1:100),
  lit_s_single = pl$lit(as_polars_series(1L)),
  lit_s_long = pl$lit(as_polars_series(1:100)),
  check = FALSE
)

#> # A tibble: 4 × 13
#>   expression          min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory              time             gc                
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>              <list>           <list>            
#> 1 lit_vec_single    378µs    467µs     2062.    39.7KB     8.49   972     4      471ms <NULL> <Rprofmem [40 × 3]> <bench_tm [976]> <tibble [976 × 3]>
#> 2 lit_vec_long      407µs    480µs     1990.    40.2KB     8.50   937     4      471ms <NULL> <Rprofmem [41 × 3]> <bench_tm [941]> <tibble [941 × 3]>
#> 3 lit_s_single      628µs    806µs     1211.    70.9KB     6.32   575     3      475ms <NULL> <Rprofmem [65 × 3]> <bench_tm [578]> <tibble [578 × 3]>
#> 4 lit_s_long        413µs    523µs     1834.    40.2KB     8.53   860     4      469ms <NULL> <Rprofmem [41 × 3]> <bench_tm [864]> <tibble [864 × 3]>
```

After:

```r
bench::mark(
  lit_vec_single = pl$lit(1L),
  lit_vec_long = pl$lit(1:100),
  lit_s_single = pl$lit(as_polars_series(1L)),
  lit_s_long = pl$lit(as_polars_series(1:100)),
  check = FALSE
)
#> # A tibble: 4 × 13
#>   expression          min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory              time               gc                  
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>              <list>             <list>              
#> 1 lit_vec_single    290µs    373µs     2447.    14.4KB     13.0  1130     6      462ms <NULL> <Rprofmem [17 × 3]> <bench_tm [1,136]> <tibble [1,136 × 3]>
#> 2 lit_vec_long      293µs    379µs     2425.    14.8KB     13.8  1057     6      436ms <NULL> <Rprofmem [18 × 3]> <bench_tm [1,063]> <tibble [1,063 × 3]>
#> 3 lit_s_single      274µs    328µs     2904.    14.4KB     12.8  1362     6      469ms <NULL> <Rprofmem [17 × 3]> <bench_tm [1,368]> <tibble [1,368 × 3]>
#> 4 lit_s_long        287µs    339µs     2797.    14.8KB     12.8  1310     6      468ms <NULL> <Rprofmem [18 × 3]> <bench_tm [1,316]> <tibble [1,316 × 3]>
```

### DIFF
--- a/src/rust/src/functions/lazy.rs
+++ b/src/rust/src/functions/lazy.rs
@@ -127,8 +127,7 @@ pub fn lit_null() -> Result<PlRExpr> {
 #[savvy]
 pub fn lit_from_series(value: &PlRSeries, keep_series: bool, keep_name: bool) -> Result<PlRExpr> {
     let s = value.series.clone();
-    let len = s.len();
-    if keep_series || len != 1 {
+    if keep_series || s.len() != 1 {
         Ok(dsl::lit(s).into())
     } else {
         // Safety: only called on length 1 series


src/rust/src/functions/lazy.rs
```diff
@@ -125,15 +125,21 @@ pub fn lit_null() -> Result<PlRExpr> {
...
+pub fn lit_from_series(value: &PlRSeries, keep_series: bool, keep_name: bool) -> Result<PlRExpr> {
     let s = value.series.clone();
-    let av = s.get(0).map_err(RPolarsErr::from)?.into_static();
-    Ok(dsl::lit(Scalar::new(s.dtype().clone(), av)).into())
+    let len = s.len();
+    if keep_series || len != 1 {
+        Ok(dsl::lit(s).into())
+    } else {
+        // Safety: only called on non-empty series
+        let av = unsafe { s.get_unchecked(0).into_static() };
```
etiennebacher: How do we know that? Seems to me we only checked that length is not 1 above.

eitsupi: In other words, the comment is confusing, but what I'm trying to say is that when that reach this line, the length of the Series is always 1, so it can't be 0.

etiennebacher: > what I'm trying to say is that when that reach this line, the length of the Series is always 1, so it can't be 0.

But I don't understand why that is. If `len` is 0 then we end in this branch, no? Maybe I missed a length check on the R side, but if that's the case, can you also mention this check in the comment?

eitsupi: `s.get_unchecked(0)` gets the 0th element of `s` (note that this is a 0-based index), but since `s` always has length 1 here, so there is always a 0th element.
The length of `s` will never be 0, so that operation is safe.

etiennebacher: >  s.get_unchecked(0) gets the 0th element of s (note that this is a 0-based index), but since s always has length 1 here, so there is always a 0th element.

I understand, but I disagree with the second part of the sentence. If `s.len() == 0` then we also end up in this `else` branch.

Sorry if I'm missing sth obvious but from the files changed here I don't see how this case is appropriately handled.

eitsupi: > If `s.len() == 0` then we also end up in this `else` branch.

I don't think it's possible to reach this line in any other case than `s.len() == 1`.
Even if it were possible to reach it, it would surely be discovered during testing due to our comprehensive testing.

https://github.com/pola-rs/r-polars/blob/60cfbf6592f60e5a343a1cc5fe6cb6bc343112e2/tests/testthat/test-as_polars_expr.R#L78-L80

etiennebacher: Ok, just wait before merging please, I want to take another look first.

etiennebacher: That was just a stupid error from my side, implementation is good, thanks.

eitsupi: To be honest, I was a little confused when I wrote it😅

We could swap the conditions and test `s.len == 1 && !keep_series`, but this isn't necessarily easier to understand too, so I've left it as is.

etiennebacher: I don't think swapping the conditions would make it clearer, only extra sleep would :smile:

PR #1524, opened 2025-08-30
test: Add tests for `pl$Partition*()` functions

etiennebacher:


tests/testthat/test-output-partition.R
eitsupi, line 1: Could you make these parametrized?

etiennebacher: I don't know how to do that since what we want to parametrize is the function name. I tried this:

```r
my_lf <- as_polars_lf(mtcars)

patrick::with_parameters_test_that(
  "pl$PartitionByKey() works",
  {
    out_path <- withr::local_tempdir()
    do.call(fun, list(pl$PartitionByKey(out_path, by = c("am", "cyl")), mkdir = TRUE))
    expect_snapshot(list.files(out_path, recursive = TRUE))
  },
  fun = list(my_lf$sink_csv, my_lf$sink_ipc, my_lf$sink_ndjson, my_lf$sink_parquet)
)
```
but this fails internally in `patrick`:

```r
Error in `purrr::pmap()`:
ℹ In index: 1.
Caused by error:
! Failed to parse glue component
Caused by error in `parse()`:
! <text>:1:1: unexpected ','
1: ,
    ^
```

This works with more classic functions like `mean()` so I guess the `$` messes things up.

eitsupi: Done in 3b02060

PR #1535, opened 2025-09-07
docs: Update docs related to deprecations in the new release

eitsupi:


R/expr-string.R
```diff
@@ -652,14 +655,22 @@ expr_str_starts_with <- function(prefix) {
 
 #' Parse string values as JSON.
 #'
+#' Parse string values as JSON.
+#' Throw errors if encounter invalid json strings.
```
etiennebacher, line 659: I think you need a `@description` here

eitsupi: Why? roxygen2 should automatically classify the second section as description.

etiennebacher: I thought the entire text from "Throw errors..." to "inspects the data at runtime" should be in the description. Currently, roxygen2 will put the next paragraphs in "Details".

If you wanted to put the two following paragraphs into "Details", can you add the `@details` tag so that it's more explicit?

etiennebacher: Also "Parse string values as JSON." is duplicated

eitsupi: > Also "Parse string values as JSON." is duplicated

What does this mean? I don't think anything has changed since before.
Check the Rd file, not this one.

etiennebacher: Sorry, my bad

PR #1557, opened 2025-09-23
feat: `<lazyframe>$sink_batches()`

eitsupi:


tests/testthat/_snaps/output-batches.md
```diff
@@ -0,0 +1,246 @@
...
+      └──────┴─────┴───────┴───────┴───┴─────┴─────┴──────┴──────┘
+      shape: (2, 11)
+      ┌──────┬─────┬───────┬───────┬───┬─────┬─────┬──────┬──────┐
+      │ mpg  ┆ cyl ┆ disp  ┆ hp    ┆ … ┆ vs  ┆ am  ┆ gear ┆ carb │
+      │ ---  ┆ --- ┆ ---   ┆ ---   ┆   ┆ --- ┆ --- ┆ ---  ┆ ---  │
+      │ f64  ┆ f64 ┆ f64   ┆ f64   ┆   ┆ f64 ┆ f64 ┆ f64  ┆ f64  │
+      ╞══════╪═════╪═══════╪═══════╪═══╪═════╪═════╪══════╪══════╡
+      │ 15.8 ┆ 8.0 ┆ 351.0 ┆ 264.0 ┆ … ┆ 0.0 ┆ 1.0 ┆ 5.0  ┆ 4.0  │
+      │ 15.0 ┆ 8.0 ┆ 301.0 ┆ 335.0 ┆ … ┆ 0.0 ┆ 1.0 ┆ 5.0  ┆ 8.0  │
+      └──────┴─────┴───────┴───────┴───┴─────┴─────┴──────┴──────┘
```
eitsupi: This result (when the callback function returns true in an in-memory engine) seems odd.
It's not terminating early.

cc @coastalwhite

coastalwhite: The in-memory engine can, by definition, not return early. It does every step in full before continuing.

eitsupi: Thanks for the quick reply! I understand.
Perhaps the Python Polars documentation may need some updating.

etiennebacher:
Thanks, I was quite busy yesterday so I couldn't explore this. I'm still confused about the use cases for this. The example with `print()` is a bit simplistic (same issue with the Python Polars docs), and even in the issues closed by this feature in Python Polars, I couldn't find clear use cases. Do you have any?

coastalwhite:
It is mostly to write custom sinks. Let's say you have a file format for which there is no sink or let's say you want an HTTP server you want to send back a batch when it is ready.

PR #1561, opened 2025-09-24
docs: Improve examples of `<lazyframe>$sink_batches()`

### LINKED ISSUE #1560: Add examples for `sink_batches()`
https://github.com/pola-rs/r-polars/pull/1557#issuecomment-3327089483

For example, a file format I often interact with in my work is `.dta` (from Stata), for which there's no Polars writer. It could be useful to show how `sink_batches()` allows to stream from parquet to multiple dta files. One downside of this example is that we'd pull an extra Suggests (`haven` for `write_dta()`), but I'm open to other examples.

etiennebacher:
Fixes #1560

### DIFF
--- a/R/output-batches.R
+++ b/R/output-batches.R
@@ -47,6 +47,7 @@
 #' # It is therefore recommended to set `chunk_size` manually.
 #'
 #' output_dir <- withr::local_tempdir()
+#' dir.create(output_dir)
 #' output_dir_idx <- 1
 #'
 #' lf$sink_batches(
@@ -64,6 +65,7 @@
 #'
 #' # The number of rows in each chunk can be adjusted with `chunk_size`.
 #' output_dir <- withr::local_tempdir()
+#' dir.create(output_dir)
 #' output_dir_idx <- 1
 #'
 #' lf$sink_batches(
@@ -81,6 +83,7 @@
 #' # To avoid manually creating paths and incrementing `output_dir_idx` in the
 #' # anonymous function, we can use function factories:
 #' output_dir <- withr::local_tempdir()
+#' dir.create(output_dir)
 #' writer_factory <- function(dir) {
 #'   i <- 0
 #'   function(df) {
--- a/R/output-batches.R
+++ b/R/output-batches.R
@@ -35,12 +35,16 @@
 #'   chunk_size = 10
 #' )
 #'
-#'
 #' # One usecase for this function is to export larger-than-RAM data to file
 #' # formats for which polars doesn't provide a writer out of the box.
 #' # The example below writes a LazyFrame by batches to a CSV file for the
 #' # sake of the example, but one could replace `write.csv()` by
 #' # `haven::write_dta()`, `saveRDS()`, or other functions.
+#' #
+#' # Note that if `chunk_size` is missing, then Polars tries to compute it
+#' # automatically. However, depending on the characteristics of the data (for
+#' # instance very long string elements), this can lead to out-of-memory errors.
+#' # It is therefore recommended to set `chunk_size` manually.
 #'
 #' output_dir <- withr::local_tempdir()
 #' output_dir_idx <- 1
--- a/R/output-batches.R
+++ b/R/output-batches.R
@@ -51,7 +51,6 @@
 #'
 #' lf$sink_batches(
 #'   \(df) {
-#'      # `df` is one chunk of the original LazyFrame.
 #'      dest <- paste0(output_dir, "/file_", output_dir_idx, ".csv")
 #'      cat(sprintf("Writing %s rows to %s\n", nrow(df), dest))
 #'      write.csv(as.data.frame(df), dest)
--- a/R/output-batches.R
+++ b/R/output-batches.R
@@ -88,19 +88,14 @@
 #'   i <- 0
 #'   function(df) {
 #'     i <<- i + 1
+#'     cat(sprintf("Write %s rows\n", nrow(df)))
 #'     as.data.frame(df) |>
 #'       write.csv(file.path(dir, sprintf("%03d.csv", i)), row.names = FALSE)
 #'   }
 #' }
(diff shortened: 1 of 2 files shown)


R/output-batches.R
```diff
@@ -19,17 +19,68 @@
...
-#' lf$sort("cyl")$sink_batches(
+#' output_dir <- withr::local_tempdir()
+#' output_dir_idx <- 1
+#'
+#' lf$sink_batches(
+#'   \(df) {
+#'      # `df` is one chunk of the original LazyFrame.
+#'      dest <- paste0(output_dir, "/file_", output_dir_idx, ".csv")
+#'      cat(sprintf("Writing %s rows to %s\n", nrow(df), dest))
+#'      write.csv(as.data.frame(df), dest)
```
etiennebacher: Weirdly, this also works completely fine with `write.csv(df, dest)` while `df` is a `DataFrame`. I don't really know why but `haven::write_dta(df, dest)` errors for instance.

R/output-batches.R
eitsupi, line 1: Could you please leave the original example as is, since it's useful for verifying operation without I/O?

eitsupi (changes requested):

R/output-batches.R
```diff
@@ -19,17 +19,68 @@
 #' - `<lazyframe>$sink_batches()` returns `NULL` invisibly.
 #' - `<lazyframe>$lazy_sink_batches()` returns a new [LazyFrame].
 #' @examples
+#' # One usecase for this function is to export larger-than-RAM data to file
+#' # formats for which polars doesn't provide a writer out of the box.
+#' # The example below writes a LazyFrame by batches to a CSV file for the
+#' # sake of the example, but one could replace `write.csv()` by
+#' # `haven::write_dta()`, `saveRDS()`, or other functions.
```
eitsupi, line 44: My guess is that Polars will try to use as much RAM as possible, so if we really want to write data larger than RAM, I'm worried you'll hit OOM the moment we export a polars DataFrame to R.
In other words, I think specifying the chunk size option is essential in such a real-world use case.

etiennebacher: Good point, I suppose the `chunk_size` parameter is the same as in `Config.set_streaming_chunk_size()`:

> By default, the chunk size is determined by the schema and size of the thread pool. For some datasets (esp. when you have large string elements) this can be too optimistic and lead to Out of Memory errors.

https://docs.pola.rs/api/python/dev/reference/api/polars.Config.set_streaming_chunk_size.html#polars.Config.set_streaming_chunk_size

R/output-batches.R
etiennebacher: Done in [11e18b6](https://github.com/pola-rs/r-polars/pull/1561/commits/11e18b60e59183a8ed3991d07a3f0ca4e59ed858) but the output might be wrong for now, see: https://github.com/pola-rs/polars/issues/24620

R/output-batches.R
etiennebacher: Done in [453925d](https://github.com/pola-rs/r-polars/pull/1561/commits/453925d1278896f197174ce03de659490a772e48)

R/output-batches.R
```diff
@@ -20,16 +20,89 @@
...
+#'   }
+#' )
+#'
+#' list.files(output_dir)
+#'
+#' pl$read_csv(output_dir)
+#'
+#' # The number of rows in each chunk can be adjusted with `chunk_size`.
+#' output_dir <- withr::local_tempdir()
+#' dir.create(output_dir)
```
eitsupi: If I understand correctly, `withr::local_tempdir()` should have already created the directory.

R/output-batches.R
```diff
@@ -20,16 +20,89 @@
...
+#'       write.csv(file.path(dir, sprintf("%03d.csv", i)), row.names = FALSE)
+#'   }
+#' }
+#' writer <- writer_factory(output_dir)
+#'
+#' lf$sink_batches(
+#'   \(df) {
+#'      cat(sprintf("Write %s rows\n", nrow(df)))
+#'      writer(df)
+#'   },
```
eitsupi: Thinking about it again, if there is a `cat()` in the `writer` function, we can simply pass `writer` here.

R/output-batches.R
etiennebacher: I thought so too but the examples failed without explicitly creating it (e.g. https://github.com/pola-rs/r-polars/actions/runs/17989365812/job/51175564349?pr=1561, in the first commit). I could also reproduce locally with `devtools::run_examples()`.

I think `withr::with_tempdir()` would automatically create the dir, I'll try this later.

R/output-batches.R
etiennebacher: Didn't see that you already reported it..

eitsupi: Ah yes, I was playing around with the examples and noticed something odd, which is why there are no snapshot tests (which I'll add later as they seem to have been fixed in upstream).

R/output-batches.R
eitsupi: Interesting. Perhaps the directory was automatically deleted after the first call?
https://withr.r-lib.org/reference/with_tempfile.html

I imagine we could disable automatic deletion by specifying `clean = FALSE`, but as you say `with_tempdir()` might be the better option here.

R/output-batches.R
etiennebacher: [c5bb821](https://github.com/pola-rs/r-polars/pull/1561/commits/c5bb82180d75955aedea355f28230a480fec17f8)

R/output-batches.R
etiennebacher: [cbb6123](https://github.com/pola-rs/r-polars/pull/1561/commits/cbb6123f903c12c8d4d9336064d18b8a2959dd0e)

eitsupi:
Thanks!

PR #1562, opened 2025-09-25
feat: `lazy_sink_*` methods and support the `engine` argument for `sink_*` methods

### LINKED ISSUE #1364: lazy versions of `sink_*` methods
Moved from eitsupi/neo-r-polars#260

---

Recent update (pola-rs/polars#21733) has made significant change to Rust's sink functions.

Previously, sink functions were methods that were executed at the end of the query, like `collect`, but now they are methods that must be executed before `collect`.

This behavior is masked in Python Polars (and current R neopolars), but we can check the behavior by explicitly specifying `lazy=True` in Python Polars.

```python
>>> import polars as pl
>>> lf = pl.LazyFrame({"a": [1, 2,]})
>>> lf.sink_ipc("test.arrow", lazy=True)
<LazyFrame at 0x7F2812DBF770>
>>> lf.sink_ipc("test.arrow", lazy=True).explain()
'SINK (file)\n  DF ["a"]; PROJECT */1 COLUMNS'
```

We can see that `SINK (file)` is added to the LazyFrame via the `sink_ipc` method.
(Note that using `collect` on this LazyFrame returns an empty DataFrame and generates a file as a side effect.)

For now, R Polars returns exactly the same LazyFrame as the input, but this behavior will need to change as Rust and Python now return mutated LazyFrames.
And we need to consider whether we should support something like the lazy argument in Python.

Personally, as I had already argued for several other functions, I believe that:

1. I do not want to add an additional argument simply to omit one simple function call, such as `$collect()`.
2. I think it is confusing that the return values have different meanings depending on the argument values.

So, I feel it is clearer and better to always return a mutated LazyFrame, as with the Rust API.
This is simple and unless we call `$collect()`, the operation should be inexpensive, so there is no risk of failure.

```r
# Before
lf$sink_ipc("test.arrow")
```

```r
# After
lf$sink_ipc("test.arrow")$collect()
```

The con is obviously that the default behavior is different from the Python API...

eitsupi:
Close #1364

### DIFF
--- a/R/output-csv.R
+++ b/R/output-csv.R
@@ -47,7 +47,7 @@
 #' tmpf <- tempfile(fileext = ".csv")
 #' as_polars_lf(mtcars)$sink_csv(tmpf)
 #'
-#' # Save a query for streaming end-to-end
+#' # Create a query that can be run in streaming end-to-end
 #' tmpf2 <- tempfile(fileext = ".csv")
 #' lf <- pl$scan_csv(tmpf)$select(pl$col("cyl") * 2)$lazy_sink_csv(tmpf2)
 #' lf$explain() |>
--- a/R/output-csv.R
+++ b/R/output-csv.R
@@ -82,6 +82,77 @@ lazyframe__sink_csv <- function(
   retries = 2,
   sync_on_close = c("none", "data", "all"),
   mkdir = FALSE,
+  engine = c("auto", "in-memory", "streaming"),
+  collapse_joins = deprecated()
+) {
+  wrap({
+    check_dots_empty0(...)
+
+    self$lazy_sink_csv(
+      path = path,
+      include_bom = include_bom,
+      include_header = include_header,
+      separator = separator,
+      line_terminator = line_terminator,
+      quote_char = quote_char,
+      batch_size = batch_size,
+      datetime_format = datetime_format,
+      date_format = date_format,
+      time_format = time_format,
+      float_scientific = float_scientific,
+      float_precision = float_precision,
+      decimal_comma = decimal_comma,
+      null_value = null_value,
+      quote_style = quote_style,
+      maintain_order = maintain_order,
+      type_coercion = type_coercion,
+      `_type_check` = `_type_check`,
+      predicate_pushdown = predicate_pushdown,
+      projection_pushdown = projection_pushdown,
+      simplify_expression = simplify_expression,
+      slice_pushdown = slice_pushdown,
+      no_optimization = no_optimization,
+      storage_options = storage_options,
+      retries = retries,
+      sync_on_close = sync_on_close,
+      mkdir = mkdir,
+      collapse_joins = collapse_joins
+    )$collect(engine = engine)
+  })
+  # TODO: support `optimizations` argument
+  invisible(NULL)
+}
+
+#' @rdname lazyframe__sink_csv
+lazyframe__lazy_sink_csv <- function(
+  path,
+  ...,
+  include_bom = FALSE,
+  include_header = TRUE,
+  separator = ",",
+  line_terminator = "\n",
+  quote_char = '"',
+  batch_size = 1024,
+  datetime_format = NULL,
+  date_format = NULL,
+  time_format = NULL,
+  float_scientific = NULL,
+  float_precision = NULL,
+  decimal_comma = FALSE,
+  null_value = "",
+  quote_style = c("necessary", "always", "never", "non_numeric"),
+  maintain_order = TRUE,
+  type_coercion = TRUE,
+  `_type_check` = TRUE,
+  predicate_pushdown = TRUE,
+  projection_pushdown = TRUE,
+  simplify_expression = TRUE,
+  slice_pushdown = TRUE,
+  no_optimization = FALSE,
+  storage_options = NULL,
+  retries = 2,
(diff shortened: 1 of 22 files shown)


R/output-parquet.R
```diff
@@ -48,7 +50,9 @@
 #' * `"all"`: syncs the file contents and metadata.
 #' @param mkdir Recursively create all the directories in the path.
 #'
-#' @return `NULL` invisibly.
+#' @return
+#' - `<lazyframe>$sink_*()` returns `NULL` invisibly.
+#' - `<lazyframe>$lazy_sink_*()` returns a new [LazyFrame].
```
etiennebacher, line 60: Can we explicitly mention that the first one directly writes the file(s) while the second one will write the file(s) only after `collect()`?

(maybe this info would fit better in `@description`)

eitsupi: How about 36efc2d?

etiennebacher: Actually I think it's better to put this in `@description`, something like:

```
#' `r lifecycle::badge("experimental")`
#'
#' This allows streaming results that are larger than RAM to be written to disk:
#' * `<lazyframe>$lazy_sink_*()` don't write directly to the output file(s) but need
#'    a `$collect()` call. 
#' * `<lazyframe>$sink_*()` write directly to the output file(s) (they are shortcuts for 
#'   `<lazyframe>$lazy_sink_*()$collect()`).
```

What do you think?

This section would also need to be updated once we have `pl$collect_all()` since it seems to me the `lazy_sink*()` are mostly useful in `pl$collect_all()`.

eitsupi: Nice, how about c841802?

R/output-csv.R
```diff
@@ -43,17 +43,93 @@
 #'   integer, then quotes will be used even if they aren`t strictly necessary.
 #'
 #' @examples
-#' # sink table 'mtcars' from mem to CSV
+#' # Sink table 'mtcars' from mem to CSV
 #' tmpf <- tempfile(fileext = ".csv")
 #' as_polars_lf(mtcars)$sink_csv(tmpf)
 #'
-#' # stream a query end-to-end
+#' # Save a query for streaming end-to-end
```
etiennebacher:
```suggestion
#' # Create a query that can be run in streaming end-to-end
```

R/output-parquet.R
```diff
@@ -48,20 +55,79 @@
...
 #'
 #' @examples
-#' # sink table 'mtcars' from mem to parquet
+#' # Sink table 'mtcars' from mem to parquet
 #' tmpf <- tempfile()
 #' as_polars_lf(mtcars)$sink_parquet(tmpf)
 #'
-#' # stream a query end-to-end
+#' # Save a query for streaming end-to-end
```
etiennebacher:
```suggestion
#' # Create a query that can be run in streaming end-to-end
```

etiennebacher (approved):
Thank you, LGTM, just a couple of places to reword I think.

etiennebacher (changes requested):
Sorry, actually it looks like we don't have any tests for any of the `$lazy_sink_*()` functions? Can you add some?

tests/testthat/_snaps/output-parquet.md
```diff
@@ -1,3 +1,21 @@
...
+
+---
+
+    Code
+      lf$collect()
+    Output
+      shape: (0, 0)
+      ┌┐
+      ╞╡
+      └┘
```
etiennebacher, line 17: In python polars, only the shape is printed and not the empty frame, do you have the same thing?

eitsupi: I don't think so.

```
In [1]: import polars as pl

In [2]: pl.DataFrame({"x": [1]}).lazy().sink_parquet("test.parquet", lazy=True)
Out[2]: <LazyFrame at 0x7F38A8D80170>

In [3]: pl.DataFrame({"x": [1]}).lazy().sink_parquet("test.parquet", lazy=True).collect()
Out[3]: 
shape: (0, 0)
┌┐
╞╡
└┘

In [4]: pl.__version__
Out[4]: '1.33.0'
```

etiennebacher: I get this on 1.34.0-beta1:
```
>>> import polars as pl
>>> pl.LazyFrame({"x": [1]}).sink_parquet("foo.parquet", lazy=True).collect()
shape: (0, 0)

>>>
```
Not a big deal anyway since we don't control the output but I thought it wouldn't print anything to the console.

eitsupi: Anyway, I don't think it has anything to do with this PR.

eitsupi:
```
In [1]: import polars as pl

In [2]: pl.__version__
Out[2]: '1.34.0-beta.1'

In [3]: pl.DataFrame({"x": [1]}).lazy().sink_parquet("test.parquet", lazy=True).collect()
Out[3]: 
shape: (0, 0)
┌┐
╞╡
└┘
```

eitsupi: As an aside, 1.34.0b2, released to PyPI an hour ago, appears to be broken: the polars module no longer works after installing it.

PR #1563, opened 2025-09-25
feat: Implement `DataTypeExpr` wrappers

etiennebacher:
https://github.com/pola-rs/r-polars/issues/1454

### DIFF
--- a/src/rust/src/expr/datatype.rs
+++ b/src/rust/src/expr/datatype.rs
@@ -1,8 +1,8 @@
-use polars::prelude::{DataType, DataTypeExpr, PlSmallStr, Schema};
-use savvy::{Result, savvy};
-
+use crate::expr::selector::PlRSelector;
 use crate::prelude::Wrap;
 use crate::{PlRDataType, PlRExpr, RPolarsErr, prelude::*};
+use polars::prelude::{DataTypeExpr, Schema};
+use savvy::{ListSexp, NumericScalar, Result, Sexp, savvy};
 
 #[savvy]
 #[repr(transparent)]
@@ -12,15 +12,15 @@ pub struct PlRDataTypeExpr {
 }
 
 impl From<DataTypeExpr> for PlRDataTypeExpr {
-    fn from(expr: DataTypeExpr) -> Result<Self> {
+    fn from(expr: DataTypeExpr) -> Self {
         PlRDataTypeExpr { inner: expr }
     }
 }
 
 #[savvy]
 impl PlRDataTypeExpr {
-    pub fn from_dtype(datatype: Wrap<PlRDataType>) -> Result<Self> {
-        DataTypeExpr::Literal(datatype.0).into()
+    pub fn from_dtype(datatype: &PlRDataType) -> Result<Self> {
+        Ok(DataTypeExpr::Literal(datatype.dt.clone()).into())
     }
 
     pub fn of_expr(expr: PlRExpr) -> Result<Self> {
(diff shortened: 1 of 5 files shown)


src/rust/src/expr/datatype.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    pub fn of_expr(expr: PlRExpr) -> Result<Self> {
+        Ok(DataTypeExpr::OfExpr(Box::new(expr.inner)).into())
+    }
+
+    pub fn self_dtype() -> Result<Self> {
+        Ok(DataTypeExpr::SelfDtype.into())
+    }
+
+    pub fn collect_dtype<'py>(&self, schema: ListSexp) -> Result<Sexp> {
```
eitsupi:
```suggestion
    pub fn collect_dtype(&self, schema: ListSexp) -> Result<PlRDataType> {
```

src/rust/src/expr/datatype.rs
```diff
@@ -0,0 +1,142 @@
...
+        let schema = <Wrap<Schema>>::try_from(schema)?.0;
+        let dtype = self
+            .clone()
+            .inner
+            .into_datatype(&schema)
+            .map_err(RPolarsErr::from)?;
+
+        let value: Sexp = PlRDataType::from(dtype.clone()).try_into()?;
+        Ok(value)
```
eitsupi:
```suggestion
        self
            .clone()
            .inner
            .into_datatype(&schema)
            .map(Into::into)
            .map_err(RPolarsErr::from)
            .map_err(Into::into)
```

eitsupi (changes requested):

etiennebacher:
@eitsupi I won't add all DataTypeExpr functions here, both so that it's easier to review and because I don't know how much time I'll have in the next days. I added all functions on the Rust side, but only `$matches()` on the R side so that we can focus on whether the "infrastructure" is fine. We can then complete it in future PRs.

R/datatype-expr-s3-base.R
```diff
@@ -0,0 +1,6 @@
+#' @export
+print.polars_datatype_expr <- function(x, ...) {
+  # x$`_datatype_expr`$display() |>
+  #   writeLines()
+  invisible(x)
+}
```
etiennebacher, line 6: This is temporary, just so that it won't error when I do `pl$dtype_of("a")` in the console.

R/functions-datatype.R
```diff
@@ -0,0 +1,18 @@
...
+pl__dtype_of <- function(col_or_expr) {
+  wrap({
+    if (is_string(col_or_expr)) {
+      e <- pl$col(col_or_expr)
+    } else if (is_polars_expr(col_or_expr)) {
+      e <- col_or_expr
+    } else {
+      abort("`col_or_expr` must be a column name or a Polars expression.")
+    }
+    `PlRDataTypeExpr`$`of_expr`(e$`_rexpr`)
```
etiennebacher: I couldn't find out a better way here. Usually, functions are marked as `#[pyfunction]` in py-polars so we can just call the function name without `PlRDataTypeExpr$`, but here this is part of the `impl PlRDataTypeExpr`.

Here's how it's called in py-polars: https://github.com/pola-rs/polars/blob/485618ad33eb056028ede4dff809eb993436b529/py-polars/polars/datatype_expr/datatype_expr.py#L152

R/functions-datatype.R
```diff
@@ -0,0 +1,18 @@
...
+#' expression][polars_expr].
+pl__dtype_of <- function(col_or_expr) {
+  wrap({
+    if (is_string(col_or_expr)) {
+      e <- pl$col(col_or_expr)
+    } else if (is_polars_expr(col_or_expr)) {
+      e <- col_or_expr
+    } else {
+      abort("`col_or_expr` must be a column name or a Polars expression.")
+    }
```
eitsupi: Please do not this.
Use S3 method instead.

eitsupi (commented):
I honestly don't see how this is useful.
It seems to be a work in progress, so can I postpone the merge?

etiennebacher:
> I honestly don't see how this is useful. It seems to be a work in progress, so can I postpone the merge?

My point was just to limit the size of the PR to avoid the massive ones that were hard to review done at the beginning of the rewrite.

In my mind, there's basically two steps to implementing this: i) add the "infrastructure" i.e. all the changes related to adding a new namespace (`R/generated-s3methods-dollar.R`, `zzz.R`, etc.), and ii) add the functions, docs, and tests in R, which is a bit tedious to do but easy to add and review. I just wanted to split these two components so that it's easier to review, but if you don't mind then I can add all methods in this PR.

eitsupi:
I think it's good to keep the changes to a minimum, but my point is not that, but that the changes made look unfinished.

src/rust/src/expr/datatype.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+#[savvy]
+impl PlRDataTypeExpr {
+    pub fn from_dtype(datatype: &PlRDataType) -> Result<Self> {
+        Ok(DataTypeExpr::Literal(datatype.dt.clone()).into())
+    }
+
+    pub fn of_expr(expr: PlRExpr) -> Result<Self> {
```
eitsupi:
```suggestion
    pub fn of_expr(expr: &PlRExpr) -> Result<Self> {
```

src/rust/src/expr/datatype.rs
```diff
@@ -0,0 +1,140 @@
+use crate::expr::selector::PlRSelector;
+use crate::prelude::Wrap;
+use crate::{PlRDataType, PlRExpr, RPolarsErr, prelude::*};
+use polars::prelude::{DataTypeExpr, Schema};
+use savvy::{ListSexp, NumericScalar, Result, savvy};
```
eitsupi:
```suggestion
use crate::{expr::selector::PlRSelector, PlRDataType, PlRExpr, RPolarsErr, prelude::*};
use savvy::{ListSexp, NumericScalar, Result, savvy};
```

PR #1580, opened 2025-10-05
feat: Implement `polars_envvars()`

### LINKED ISSUE #1575: Add `pl$Config()`
Or at least document the envvars used by polars, e.g. `POLARS_FMT_MAX_ROWS` used for option `set_tbl_rows`

https://github.com/pola-rs/polars/blob/main/py-polars/src/polars/config.py#L1289-L1322

etiennebacher:
Fixes #1575 

Mostly copy-pasted from the implementation in old r-polars.


R/polars_envvars.R
```diff
@@ -0,0 +1,139 @@
...
+#' # We can temporarily allow for wider columns with `withr::with_envvar()`:
+#' withr::with_envvar(
+#'   list(POLARS_FMT_STR_LEN = "50"),
+#'   print(df)
+#' )
+#'
+#' # Or we could set it permanently with `Sys.setenv(POLARS_FMT_STR_LEN = "50")`.
+polars_envvars <- function() {
+  envvars <- rbind(
+    # c("POLARS_AUTO_STRUCTIFY", ""),
```
etiennebacher: This was already commented out in the old r-polars. I see that this is deprecated so I guess we don't need to include it here?

R/polars_envvars.R
```diff
@@ -0,0 +1,139 @@
...
+    c("POLARS_WARN_UNSTABLE", "0")
+  ) |>
+    as.data.frame()
+
+  out <- vector("list", length(envvars))
+  for (i in seq_len(nrow(envvars))) {
+    e <- envvars[[1]][i]
+    out[[e]] = Sys.getenv(e, unset = envvars[[2]][i])
+  }
+  structure(out, class = "polars_envvars")
```
eitsupi: The class name should be updated to avoid conflicts with polars0.
See `polars_info()`.

R/polars_envvars.R
eitsupi: Ah yes, this stopped working in Python Polars at some point, so I think they also wiped it from the documentation.

R/polars_envvars.R
etiennebacher: Thanks, done in [cf404d7](https://github.com/pola-rs/r-polars/pull/1580/commits/cf404d7a71d571227c30dcf5b183f5a26ff2fca2)

etiennebacher:
> Given that the environment variables are all strings, I suspect a named string vector would suffice here rather than using a list.

I like that `polars_options()`, `polars_envvars()` and `polars_info()` all return a list. This prevents gotchas because `[` would return a list in some cases and a vector in other cases.

R/polars_envvars.R
```diff
@@ -0,0 +1,138 @@
...
+    c("POLARS_FMT_TABLE_ROUNDED_CORNERS", "0"),
+    c("POLARS_MAX_THREADS", thread_pool_size()),
+    c("POLARS_STREAMING_CHUNK_SIZE", "variable"),
+    c("POLARS_TABLE_WIDTH", "variable"),
+    c("POLARS_VERBOSE", "0"),
+    c("POLARS_WARN_UNSTABLE", "0")
+  ) |>
+    as.data.frame()
+
+  out <- vector("list", length(envvars))
```
eitsupi: This seems like a pretty roundabout way to do this when building a list, could you please fix that?

etiennebacher: Done in [866c898](https://github.com/pola-rs/r-polars/pull/1580/commits/866c898d4ce986c65d03a85dc4be2291b5d58173)

PR #1584, opened 2025-10-07
release: version 1.4.0

eitsupi:


etiennebacher:
Just for the future, I would have liked to check tidypolars with this version before releasing. Shouldn't be a big deal for this release I think.

etiennebacher:
But also thanks for all the efforts to maintain the infra and take care of the releases

eitsupi:
Just to be clear, I edited the milestone and set the date a few days ago (as I think is the case with most releases these days).

PR #1591, opened 2025-10-10
feat: More info via `polars_info()`

eitsupi:


R/polars_info.R
```diff
@@ -44,8 +52,12 @@ print.polars_info_list <- function(x, ...) {
     cat("\n")
   }
 
-  cat("Polars R package version : ", format(x$versions$r_package), "\n", sep = "")
-  cat("Rust Polars crate version: ", format(x$versions$rust_crate), "\n", sep = "")
+  cat("Polars R package version   : ", format(x$versions$r_package), "\n", sep = "")
+  cat("Rust Polars crate version  : ", format(x$versions$rust_crate), "\n", sep = "")
+  cat("\n")
```
etiennebacher, line 57:
```suggestion
```
Makes sense to me to have all these versions in a single block

eitsupi: Yeah, I'm struggling with this too.

For now, I think it makes more sense to display `py_version` separately since it's at a different level in the List.

I considered putting `py_version` under `versions`, but decided against it since this package doesn't actually include Python Polars.

etiennebacher: That's ok for me

PR #1598, opened 2025-10-14
feat: Implement `pl$collect_all()`

etiennebacher:
Part of #1397

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "r-polars"
-version = "1.5.0-rc.2"
+version = "1.6.0-rc.1"
 edition = "2024"
 rust-version = "1.89.0"
 publish = false
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,7 +52,7 @@ VignetteBuilder:
 Config/Needs/dev: devtools, lifecycle, readr, glue, RcppTOML, smvr
 Config/Needs/lint: fs, lintr
 Config/Needs/website: etiennebacher/altdoc, future.apply
-Config/polars/lib-version: 1.5.0-rc.2
+Config/polars/lib-version: 1.6.0-rc.1
 Config/testthat/edition: 3
 Config/testthat/parallel: true
 Config/testthat/start-first: lazyframe-frame, *-s3-base, polars_options,
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -2248,7 +2248,7 @@ checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
 
 [[package]]
 name = "r-polars"
-version = "1.5.0-rc.2"
+version = "1.6.0-rc.1"
 dependencies = [
  "ciborium",
  "either",
--- a/tools/lib-sums.tsv
+++ b/tools/lib-sums.tsv
@@ -1,8 +1,0 @@
-url	sha256sum
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-aarch64-apple-darwin.tar.gz	d379ccc7de3c468d70c0c73f8340a7927fc15c8fc5d0940ec61b850bd0dde0f7
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-aarch64-unknown-linux-gnu.tar.gz	74b40b7b023f4c50156f112147dd091d6f65d8a88094f1a6a86882041160e90d
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-aarch64-unknown-linux-musl.tar.gz	f379432eb12c435fef9d6cfb67f616e82345d661d1c7b8a426e783bde987b7ec
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-x86_64-apple-darwin.tar.gz	353f59f34e5c0001fc926fb2401fed751d8f4fb6a5e4bc3ab23f2131806ace3e
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-x86_64-pc-windows-gnu.tar.gz	a812b2c5608c127a7e58f72a094f97790c998f7faeeb38b4dcb03701d7ba18e3
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-x86_64-unknown-linux-gnu.tar.gz	ef15067afe4bde56e83ab6948202102314fd7d8e260e1ab1110b8236f7365ee1
-https://github.com/pola-rs/r-polars/releases/download/lib-v1.5.0-rc.2/libr_polars-1.5.0-rc.2-x86_64-unknown-linux-musl.tar.gz	bea18fd92682e9bf026175c94c474d4607f6313eb18ab220a00758d2e9c67063


R/functions-lazy.R
```diff
@@ -121,3 +121,60 @@ pl__arg_sort_by <- function(
...
+    optflags <- list(
+      type_coercion = type_coercion,
+      predicate_pushdown = predicate_pushdown,
+      projection_pushdown = projection_pushdown,
+      simplify_expression = simplify_expression,
+      slice_pushdown = slice_pushdown,
+      comm_subplan_elim = comm_subplan_elim,
+      comm_subexpr_elim = comm_subexpr_elim,
+      cluster_with_columns = cluster_with_columns
+    )
```
etiennebacher, line 169: All those args are deprecated in `pl.collect_all()` but also in `lazyframe.collect()`, `lazyframe.explain()`, etc. so I thought we could keep those arguments for now in this PR, and then do another PR where we deprecate those arguments everywhere at once.

src/rust/src/conversion/mod.rs
```diff
@@ -1044,3 +1045,35 @@ impl TryFrom<&str> for Wrap<UnicodeForm> {
...
+    fn try_from(list: ListSexp) -> Result<Self, savvy::Error> {
+        let opts = PlROptFlags::empty()?;
+        let _ = list.names_iter().map(|x| {
+            let elem_name: &str = x;
+            let elem_value = list.get(x).unwrap().into_typed();
+            let elem_value = match elem_value {
+                TypedSexp::Logical(s) => {
+                    let vec = s.to_vec().to_owned();
+                    *vec.first().unwrap()
+                }
```
etiennebacher: I guess we should check on the R side that `optflags` is a list of logical values.

src/rust/src/lazyframe/optflags.rs
```diff
@@ -0,0 +1,192 @@
+use super::PlROptFlags;
+use parking_lot::RwLock;
+use polars::prelude::OptFlags;
+use savvy::{Result, Sexp, savvy};
+
+#[savvy]
+impl PlROptFlags {
```
etiennebacher: This is done with macros in Python Polars: https://github.com/pola-rs/polars/blob/a143eb0d7077ee9da2ce209a19c21d7f82228081/crates/polars-python/src/lazyframe/optflags.rs

But I couldn't make it work with `savvy` so I just added all the methods below.

R/functions-lazy.R
eitsupi: Nice, but this function is new one, so could you omit these arguments as like `<lf>$sink_batches()`?
https://github.com/pola-rs/r-polars/blob/1089d3e90e587cafa387af3f28e11ad696e74063/R/output-batches.R#L125

etiennebacher: The difference here is that in `.sink_batches()`, the optimizations are applied in `.with_optimizations()` after calling the Rust function `sink_batches()`: https://github.com/pola-rs/polars/blob/a6fa669f4464aa18bd73fa8b6440c9816af64de2/py-polars/src/polars/lazyframe/frame.py#L3739

For `.collect_all()`, the optflags have to be passed to the Rust function, so we need to pass a list of optimizations anyway: https://github.com/pola-rs/polars/blob/a6fa669f4464aa18bd73fa8b6440c9816af64de2/py-polars/src/polars/functions/lazy.py#L2108

 I could hardcode the list in the body of `lazyframe__collect_all()`, but I don't really see the point of preventing the user from setting these arguments since we don't know exactly when we'll remove them in favor of a single argument `optimizations`.

R/000-wrappers.R
```diff
@@ -4336,6 +4341,227 @@ class(`PlRLazyGroupBy`) <- c("PlRLazyGroupBy__bundle", "savvy_polars__sealed")
...
 }
 
+### wrapper functions for PlROptFlags
+
+`PlROptFlags_copy` <- function(self) {
+  function() {
+    .savvy_wrap_PlROptFlags(.Call(savvy_PlROptFlags_copy__impl, `self`))
+  }
+}
+
```
etiennebacher: I don't know if those functions should be exported by `savvy` to have them on the R side, it seems they are just defined in `optflags.rs` in Python polars: https://github.com/search?q=repo%3Apola-rs%2Fpolars+get_comm_subplan_elim&type=code

We should probably not export them, they're not really useful on the R side.

src/rust/src/conversion/mod.rs
```diff
@@ -1044,3 +1045,35 @@ impl TryFrom<&str> for Wrap<UnicodeForm> {
...
+            match elem_name {
+                "predicate_pushdown" => opts.set_predicate_pushdown(elem_value),
+                "projection_pushdown" => opts.set_projection_pushdown(elem_value),
+                "simplify_expression" => opts.set_simplify_expression(elem_value),
+                "slice_pushdown" => opts.set_slice_pushdown(elem_value),
+                "comm_subplan_elim" => opts.set_comm_subplan_elim(elem_value),
+                "comm_subexpr_elim" => opts.set_comm_subexpr_elim(elem_value),
+                "cluster_with_columns" => opts.set_cluster_with_columns(elem_value),
+                "check_order_observe" => opts.set_check_order_observe(elem_value),
+                _ => unreachable!(),
```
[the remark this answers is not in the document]
etiennebacher: I don't this is an issue since we create the list on the R side, so we know that it cannot contain unknown args.

R/functions-lazy.R
eitsupi: I don't really get what you're trying to say.
Unlike Python Polars, these arguments didn't exist here in the past, why add something that's not recommended?
Please hard code internally if needed and do not expose these.

etiennebacher: My point was that if we hardcode the values internally, then the user won't be able to change them until we implement the `optflags` argument in the future. This is why I wanted to add those arguments in the meantime, even if they are deprecated. If we implement `optflags` soon after this PR, then this doesn't matter.

I'll update the PR to hardcode these values internally later today.

src/rust/Cargo.toml
```diff
@@ -41,6 +41,7 @@ state = "0.6"
 strum = "0.27"
 strum_macros = "0.27"
 thiserror = "2"
+parking_lot = "0.12.5"
```
eitsupi:
```suggestion
parking_lot = "0.12"
```

src/rust/src/conversion/mod.rs
```diff
@@ -1044,3 +1045,35 @@ impl TryFrom<&str> for Wrap<UnicodeForm> {
         Ok(Wrap(parsed))
     }
 }
+
+impl TryFrom<ListSexp> for Wrap<PlROptFlags> {
```
eitsupi, line 1071: Is this needed?

R/functions-lazy.R
eitsupi: > My point was that if we hardcode the values internally, then the user won't be able to change them until we implement the `optflags` argument in the future.

Of course I understand that.
I think something similar happened in the past with the engine argument in `<lazyframe>$sink_*`, but I don't think anyone complained.

R/000-wrappers.R
etiennebacher: What do you think about this?

R/functions-lazy.R
etiennebacher: Done in [d7d51bc](https://github.com/pola-rs/r-polars/pull/1598/commits/d7d51bcf5f26ef24826c7738d7b2a12beba961cb)

src/rust/Cargo.toml
etiennebacher: [9100db9](https://github.com/pola-rs/r-polars/pull/1598/commits/9100db9ad316eb1b5e0c7dedecbede3ee280fe3c)

src/rust/src/conversion/mod.rs
etiennebacher: We need to pass a list with all the optimizations since those are used in `LazyFrame::collect_all_with_engine`. Even if the values are hardcoded for now, when we support `optimizations` in the future we'll need this conversion anyway.

eitsupi: Wouldn't this be unnecessary if `PlROptFlags` was constructed on the R side?
(You're adding a bunch of functions for that, right?)

R/000-wrappers.R
eitsupi: I agree that it's not necessary.

src/rust/src/conversion/mod.rs
etiennebacher: Now that the `PlROptFlags` are not exported anymore by `savvy`, we definitely need this function to convert the R list to `PlROptFlags`.

eitsupi (approved):
~~Thank you, it looks good, but I'd like to fix #1625 and release that first, and then I'll merge this.~~

Edit: #1625 was due to my misunderstanding; the problem was on the Python Polars side... Therefore, it seems there's no need to release version 1.5.1.

Issue #1625, opened 2025-11-01
Windows binary has the different DSL hash


eitsupi:
See #1621

```
── Failed tests ────────────────────────────────────────────────────────────────
Error ('test-integration-reticulate.R:40:3'): LazyFrame roundtrip via py-polars
<polars.exceptions.ComputeError/polars.exceptions.PolarsError/python.builtin.Exception/python.builtin.BaseException/python.builtin.object/error/condition>
Error in `py_call_impl(callable, call_args$unnamed, call_args$named)`: polars.exceptions.ComputeError: deserialization failed

given DSL_SCHEMA_HASH: 6286bad7b59c6dffbabcb7eda0d3a1c386cc651ad8b2479b2e4e4686199e04e1 is not compatible with this Polars version which uses DSL_SCHEMA_HASH: 45389c69048be32a61fcfa8990e6bd529f3d369c229e6004318f7e15baa0b597
error: can't deserialize DSL with incompatible schema
Run `reticulate::py_last_error()` for details.
```

https://github.com/pola-rs/r-polars/actions/runs/18998000214/job/54260599368?pr=1621#step:11:620

```sh
$ wget \
    https://raw.githubusercontent.com/pola-rs/polars/df69276daf5d195c8feb71eef82cbe9804e0f47f/crates/polars-plan/dsl-schema-hashes.json \
    --quiet
$ sha256sum dsl-schema-hashes.json
6286bad7b59c6dffbabcb7eda0d3a1c386cc651ad8b2479b2e4e4686199e04e1  dsl-schema-hashes.json
$ sed -e 's/$/\r/g' dsl-schema-hashes.json | head -c -1 >crlf.json
$ sha256sum crlf.json
45389c69048be32a61fcfa8990e6bd529f3d369c229e6004318f7e15baa0b597  crlf.json
```

Git for Windows had rewritten the files, resulting in different DSL versions.

PR #1627, opened 2025-11-01
ci: Do not build polars with the autocrlf setting of git for windows

eitsupi:
Related to #1625


eitsupi:
I was wondering why the problem wasn't being solved despite trying various things, but I think this is a problem with Python Polars, not here.
I'm closing this.

PR #1633, opened 2025-11-03
refactor: Define `QueryOptFlags` S7 class

### LINKED ISSUE #1613: Add S7 to the dependencies
Since ggplot2 has added S7 as a dependency and best practices are starting to emerge, I would like to add S7 as a dependency and prepare for the transition.
First, I want to migrate from a simple list-based S3 class like `polars_info_list` to S7.

eitsupi:
Close #1613

This is more of a proof of concept, but I think it works well.


src/rust/src/conversion/mod.rs
```diff
@@ -1068,30 +1068,62 @@ impl TryFrom<StringSexp> for Wrap<ScanSources> {
...
+        match obj.get_class() {
+            Some(class) => {
+                if !class.contains(&"polars::QueryOptFlags") {
+                    return Err("`polars::QueryOptFlags` must be passed.".into());
+                }
+            }
+            _ => {
+                return Err("`polars::QueryOptFlags` must be passed.".into());
+            }
+        }
```
eitsupi: I'm not sure if this is useful.

etiennebacher: Isn't this already covered by `check_is_S7(optflags, QueryOptFlags)`?

eitsupi: Yes, that's right. So I don't think it's particularly useful, but since it's a PoC I left it in for now.
We can remove it before merging.

R/utils-s7.R
etiennebacher, line 1: I think it would be useful to have a benchmark for this (in the linked issue too)

eitsupi: Since I'm on mobile, I can only share the results I remember:

- In the original implementation with validation, `eager_opt_flags()` took three times as long as `QueryOptFlags()` (700us vs. 230us).
- Without validation, there was almost no difference (240us vs. 230us).

This can be a big problem for DataFrame queries that call `collect` frequently, so further improvement is needed.

eitsupi: I tried:

```r
bench::mark(
  bare = pl$QueryOptFlags(),
  func = eager_opt_flags(),
  check = FALSE
)
```

In 3f588c6 

```
# A tibble: 2 × 13
  expression      min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory     time       gc      
  <bch:expr> <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>     <list>     <list>  
1 bare          231µs   286µs     3354.        0B     19.5  1549     9      462ms <NULL> <Rprofmem> <bench_tm> <tibble>
2 func          729µs   954µs      924.    77.2KB     12.9   431     6      467ms <NULL> <Rprofmem> <bench_tm> <tibble>
```

In c4c1cf0

```
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory     time       gc      
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>     <list>     <list>  
1 bare          225µs    257µs     3669.        0B     21.7  1691    10      461ms <NULL> <Rprofmem> <bench_tm> <tibble>
2 func          278µs    332µs     2885.     254KB     20.3  1280     9      444ms <NULL> <Rprofmem> <bench_tm> <tibble>
```

eitsupi:
```r
bench::mark(
  collect = as_polars_lf(NULL)$collect(),
  collect_all = pl$collect_all(list(as_polars_lf(NULL))),
  check = FALSE
)
```

In 3f588c6

```
# A tibble: 2 × 13
  expression       min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory              time             gc      
  <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>              <list>           <list>  
1 collect     974.02µs   1.35ms      706.    9.43KB     6.25   339     3      480ms <NULL> <Rprofmem [23 × 3]> <bench_tm [342]> <tibble>
2 collect_all   1.21ms   1.62ms      596.    6.83KB    10.7    279     5      468ms <NULL> <Rprofmem [17 × 3]> <bench_tm [284]> <tibble>
```

In c4c1cf0

```
# A tibble: 2 × 13
  expression       min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory                 time       gc      
  <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>                 <list>     <list>  
1 collect       1.14ms    1.5ms      645.    1.02MB     6.30   307     3      476ms <NULL> <Rprofmem [1,922 × 3]> <bench_tm> <tibble>
2 collect_all   1.38ms   2.23ms      385.   75.18KB     8.65   178     4      462ms <NULL> <Rprofmem [153 × 3]>   <bench_tm> <tibble>
```

eitsupi:
I was concerned about the performance degradation caused by validation, but it seems that the overhead of at least one validation is negligible.
I will take further measures (to avoid the validation step) in subsequent work (to implement the `optimizations` argument to `<lazyframe>$collect` and call it every time with DataFrame methods).

```r
bench::mark(
  collect = as_polars_lf(NULL)$collect(),
  collect_all = pl$collect_all(list(as_polars_lf(NULL))),
  check = FALSE
)
```

In 5eba02cd371e8247daabe2f78c4292c6108395c7

```
# A tibble: 2 × 13
  expression       min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory               time       gc      
  <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>               <list>     <list>  
1 collect       1.38ms   1.89ms      497.   850.5KB     2.04   244     1      491ms <NULL> <Rprofmem>           <bench_tm> <tibble>
2 collect_all   1.48ms   2.23ms      415.    69.4KB     2.04   203     1      489ms <NULL> <Rprofmem [144 × 3]> <bench_tm> <tibble
```

In current main (6cd8df81059d798f53d1702f676254494f39ee0b)

```
# A tibble: 2 × 13
  expression       min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory              time       gc      
  <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>              <list>     <list>  
1 collect       1.66ms   2.25ms      418.    9.43KB     4.16   201     2      481ms <NULL> <Rprofmem [23 × 3]> <bench_tm> <tibble>
2 collect_all    1.2ms   1.72ms      552.    6.83KB     4.11   269     2      487ms <NULL> <Rprofmem [17 × 3]> <bench_tm> <tibble>
```

tests/testthat/test-lazyframe-opt_flags.R
```diff
@@ -0,0 +1,14 @@
...
+  expect_snapshot(
+    pl$QueryOptFlags(TRUE),
+    error = TRUE,
+    cnd_class = TRUE
+  )
+
+  opt_flags <- pl$QueryOptFlags()
+
+  expect_snapshot(opt_flags)
+  expect_snapshot(eager_opt_flags())
```
etiennebacher, line 16: Can you add a test that this gives a proper error?
```r
expect_snapshot(
  opt_flags@type_coercion <- 1,
  error = TRUE
)
```

R/lazyframe-opt_flags.R
```diff
@@ -0,0 +1,138 @@
...
+
+# TODO: This should be a constructor
+# https://github.com/RConsortium/S7/issues/522
+eager_opt_flags <- function() {
+  QueryOptFlags()$no_optimizations() |>
+    set_props_uncheck(
+      eager = TRUE,
+      simplify_expression = TRUE
+    )
+}
```
etiennebacher, line 148: I find it confusing to have two different syntaxes for `pl$QueryOptFlags()` and `eager_opt_flags()`, can this be harmonized?

etiennebacher (changes requested):

R/lazyframe-opt_flags.R
eitsupi: `eager_opt_flags()` is for internal use only.
Not exported.

etiennebacher: Ah ok, my bad, thanks

R/lazyframe-opt_flags.R
```diff
@@ -0,0 +1,138 @@
+# nolint start: object_length_linter
+
+# The env for storing QueryOptFlags methods
+polars_query_opt_flags__methods <- new.env(parent = emptyenv())
+
+QueryOptFlags <- new_class(
```
etiennebacher, line 34: Can you add documentation (including examples) for this, so that we can directly refer to it when we start using `pl$QueryOptFlags()` in other places?

eitsupi: How about 2354c10?
I can't provide a proper example because there is no function that takes this as an argument yet.

etiennebacher: I don't understand why it's marked as internal. I thought the usecase would be something like this:

```r
lf$collect(optimizations = pl$QueryOptFlags())
```

since most of the individual optimization parameters are deprecated in Python Polars: https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.collect.html. In which case this function needs to be exported.

That's also why I had in mind to have examples like this:

```r
opts <- pl$QueryOptFlags()
opts

opts <- pl$QueryOptFlags(predicate_pushdown = FALSE)
opts
```

Did you want to keep this function internal only?

eitsupi: Sorry for the confusion. I'm not saying that this will be internal in the future, I'm just marking it as internal for now because it's only used internally as of the time this PR is merged (That's why the title of the PullRequest is "refactor" instead of "feat.").

Of course, in the future we'll need to make this public API along with deprecating various arguments.

etiennebacher: Thanks for the explanation, makes sense. Can you still add the examples, just to show the basic behavior of this function? 

This way, we'll just have to remove the `@keywords internal` when we want to use this function elsewhere in the package.

etiennebacher: Sorry, forget about it, didn't notice there were already examples

etiennebacher (approved):
Thanks, LGTM!

PR #1634, opened 2025-11-09
perf: Skip validation step for attributes of S7 objects on the R side

eitsupi:
Follow up for #1633

S7 objects are validated when they are created, so there is usually no need to validate them again on the R side.
I have modified the Rust side so that it does not panic even if an invalid S7 object is passed, and an appropriate error will be raised.


etiennebacher:
>  I have modified the Rust side so that it does not panic even if an invalid S7 object is passed, and an appropriate error will be raised.

Can this be covered in a test?

eitsupi:
> Can this be covered in a test?

At this time, I haven't added tests because there is no function that takes `QueryOptFlags` as an argument, but we can add it later.

eitsupi:
#1635

PR #1635, opened 2025-11-09
feat: `optimizations` arguments

eitsupi:


R/lazyframe-frame.R
```diff
@@ -268,58 +298,40 @@ lazyframe__group_by <- function(..., .maintain_order = FALSE) {
...
-  projection_pushdown = TRUE,
-  simplify_expression = TRUE,
-  slice_pushdown = TRUE,
-  comm_subplan_elim = TRUE,
-  comm_subexpr_elim = TRUE,
-  cluster_with_columns = TRUE,
-  no_optimization = FALSE,
   engine = c("auto", "in-memory", "streaming"),
-  `_check_order` = TRUE,
-  `_eager` = FALSE,
```
eitsupi, line 283: I'd like to remove these _internal use only_ arguments, because there are too many arguments and it's hard to read.

IIUC, these are clearly stated in the documentation as being for internal use and are not official APIs, so I don't think removing these are breaking changes.
(Of course, I will clearly state the deletion in the NEWS.)

@etiennebacher Any thoughts?

etiennebacher: Agreed on the principle because they were marked internal from the beginning I think.

However, I think we call `collect(_eager = TRUE)` in a bunch of DataFrame methods. Wouldn't those break if we remove this argument?

eitsupi: > However, I think we call `collect(_eager = TRUE)` in a bunch of DataFrame methods. Wouldn't those break if we remove this argument?

They're all going to be replaced with something like `collect(optimizations = eager_opt_flags())` (Same as Python Polars).

eitsupi: > They're all going to be replaced with something like `collect(optimizations = eager_opt_flags())` (Same as Python Polars).

b209e26

etiennebacher: > They're all going to be replaced with something like `collect(optimizations = eager_opt_flags())` (Same as Python Polars).

Ah I hadn't noticed, I'm fine with removing these internal args then. Thanks!

eitsupi:
I'll add some tests tomorrow...

R/functions-lazy.R
```diff
@@ -148,19 +148,18 @@ pl__arg_sort_by <- function(
 pl__collect_all <- function(
   lazy_frames,
   ...,
-  engine = c("auto", "in-memory", "streaming")
+  engine = c("auto", "in-memory", "streaming"),
+  optimizations = QueryOptFlags()
```
etiennebacher: Shouldn't this be `pl$QueryOptFlags()`? It will appear in the "Usage" section so it might be confusing to have places where we use `pl$QueryOptFlags()` and other where we use `QueryOptFlags()`.

NEWS.md
```diff
@@ -2,9 +2,46 @@
...
+
+  Functions affected are those that gained the `optimizations` argument.
+  See the New features section below for details.
+  Also, for the experimental `<lazyframe>$sink_*` methods,
+  the above arguments are removed instead of being deprecated.
+
 ### New features
 
-- `pl$collect_all()` to efficiently collect a list of LazyFrames (#1598).
+- The following functions gain the experimental `optimizations` argument
```
etiennebacher, line 34: `<lazyframe>$explain()` and `<lazyframe>$profile()` also gain the arg `engine`.

NEWS.md
```diff
@@ -2,9 +2,46 @@
...
+  - `<lazyframe>$collect()`
+  - `<lazyframe>$explain()`
+  - `<lazyframe>$profile()`
+  - `<lazyframe>$to_dot()`
+  - `<lazyframe>$sink_batches()`
+  - `<lazyframe>$sink_csv()`
+  - `<lazyframe>$sink_ipc()`
+  - `<lazyframe>$sink_parquet()`
+  - `<lazyframe>$sink_ndjson()`
+  - `pl$collect_all()`
```
etiennebacher: I don't think `pl$collect_all()` should be in the list since it's an entirely new function in this version (and is listed below).

NEWS.md
```diff
@@ -2,9 +2,46 @@
...
+  - `cluster_with_columns`
+  - `no_optimization`
+  - `_type_check` (removed)
+  - `_check_order` (removed)
+  - `_eager` (removed)
+
+  Functions affected are those that gained the `optimizations` argument.
+  See the New features section below for details.
+  Also, for the experimental `<lazyframe>$sink_*` methods,
+  the above arguments are removed instead of being deprecated.
```
etiennebacher: I'm not a big fan of this. I know those are marked experimental, but I'd rather have a deprecation warning in those functions too. There are already so many deprecations in this PR that the marginal gain of removing those arguments entirely without a deprecation warning is very small IMO. In other words, it doesn't cost much for us to add a deprecation warning to the args in those functions, but it might prevent code from breaking.

Also, this would break the behavior in `tidypolars` and I'm not sure I could address before 1.6.0.

etiennebacher (changes requested):
Looks great, thanks! 

I don't think these args should be removed without deprecation in the sink functions (see the comment). Apart from that, just a few minor remarks.

etiennebacher (approved):
Thanks!

PR #1636, opened 2025-11-09
chore: Bump polars to py-1.35.2

etiennebacher:
https://github.com/pola-rs/polars/releases/tag/py-1.35.2

There a few important bug fixes  (at least important enough to make a patch version). I put our version to 1.5.1 because I imagine we could do a patch release too, what do you think @eitsupi ?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,7 +53,7 @@ VignetteBuilder:
 Config/Needs/dev: devtools, lifecycle, readr, glue, RcppTOML, smvr
 Config/Needs/lint: fs, lintr
 Config/Needs/website: etiennebacher/altdoc, future.apply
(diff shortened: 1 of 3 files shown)


eitsupi:
> I think we can make a new release, unless maybe you wanted to make more progress on the S7 front before making a new one?

The release process is quite a hassle, so I simply wanted to postpone it if it didn't seem necessary.
If it is a patch release (1.5.1), feature additions in the main branch will be ignored.

The one I'm really hoping for is pola-rs/polars#25123 to be released, and if it included in the patch version I'd imagine I would have made a patch release here, but they didn't cherry-pick it (which means we can't use Polars Cloud from Windows, which is quite a surprise).

PR #1637, opened 2025-11-09
feat: Implement `<expr>$is_close()`

etiennebacher:


tests/testthat/test-expr-expr.R
```diff
@@ -2935,3 +2935,68 @@ test_that("index_of works", {
 test_that("Deprecated shrink_dtype", {
   expect_snapshot(pl$col("foo")$shrink_dtype(), cnd_class = TRUE)
 })
+
+test_that("is_close works", {
+  df <- pl$DataFrame(a = c(1.5, 2.0, NaN), b = c(1.55, 2.2, NaN))
+
+  # abs_tol works
+  expect_equal(
```
eitsupi, line 2943: Could you use `expect_query_equal`?

etiennebacher: We use `expect_equal()` everywhere when testing expressions, `expect_query_equal()` is when we test methods that exist for DataFrame and LazyFrame so I don't see why it's needed here.

eitsupi: My bad, thanks.

My concern was the streaming mode, but so far it's been fine.

PR #1642, opened 2025-11-14
test: Update snapshots with the latest testthat and suppress some stdouts

eitsupi:


tests/testthat/_snaps/lazyframe-opt_flags.md
```diff
@@ -97,7 +97,7 @@
 
     Code
       opt_flags@type_coercion <- 1L
-    Condition <simpleError>
+    Condition <rlang_error>
```
eitsupi, line 100: It seems ~~a bug~~ an intentional change of testthat (r-lib/testthat#2286).

tests/testthat/test-expr-datetime.R
```diff
@@ -302,7 +302,7 @@ test_that("hour minute", {
   )
 })
 
-
+# TODO: rewrite tests
```
etiennebacher, line 305: What needs to change?

PR #1644, opened 2025-11-15
refactor: Simplify most `<lazygroupby>` methods 

### LINKED ISSUE #1639: Refactor `dataframe-group-by-general.R`
https://github.com/pola-rs/r-polars/pull/1638#pullrequestreview-3460121205

etiennebacher:
Fixes #1639 

~We need `lazygroupby__max` to be loaded before `groupby_max` (for example), otherwise loading the package would fail with `"lazygroupby_max not found"`. This is why the "Collate" field is required in DESCRIPTION (it is generated by `roxygen2` based on the `@include` tag I added in one of the docs).~


etiennebacher:
Yes indeed, done

PR #1655, opened 2025-11-18
feat: Implement `<expr>$arr$agg()` and `<expr>$list$agg()`

etiennebacher:
Part of #1397

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@
 
 ### Other changes
 
-- `<expr>$list$eval()` now properly errors (as documented) if the input is not a Polars expression (1655).
+- `<expr>$list$eval()` now properly errors (as documented) if the input is not a Polars expression (#1655).
 
 ## polars 1.6.0
 


R/expr-array.R
```diff
@@ -461,3 +461,34 @@ expr_arr_eval <- function(expr, ..., as_list = FALSE) {
...
+#'   null_count = pl$col("a")$arr$agg(pl$element()$null_count())
+#' )
+#'
+#' # The column "no_nulls" has dtype arr(u32) because the expression doesn't
+#' # guarantee to return a scalar.
+#' df$with_columns(
+#'   no_nulls = pl$col("a")$arr$agg(pl$element()$drop_nulls())
+#' )
+expr_arr_agg <- function(expr) {
+  self$`_rexpr`$arr_agg(as_polars_expr(expr, as_lit = TRUE)$`_rexpr`) |>
```
eitsupi: In Python Polars, only Expr input is accepted here, so it would be more correct to translate it without executing `as_polars_expr`.
What do you think?

etiennebacher: Indeed, fixed.

The same mistake is done in `list$eval()` and `arr$eval()`. I added `arr$eval()` recently so I fixed it in this PR too it but I think changing `list$eval()` would be a breaking change.

eitsupi: In the document, only Expr is specified as the input, so it should be okay to change `expr_list_eval` as well.
https://github.com/pola-rs/r-polars/blob/ff513541733faedd1f91ecf0b48c6950ff1a7eb9/R/expr-list.R#L485-L486

etiennebacher: Ok, done

PR #1662, opened 2025-11-21
chore: Bump polars to 2025-11-21 unreleased version

eitsupi:
Includes big changes.

- [x] pola-rs/polars#25016
- [x] pola-rs/polars/pull/25302, pola-rs/polars#25303, pola-rs/polars#25308

### DIFF
--- a/tests/testthat/test-lazyframe-frame.R
+++ b/tests/testthat/test-lazyframe-frame.R
@@ -422,6 +422,14 @@ test_that("unique works", {
       ham = rep("b", 3)
     )
   )
+
+  # Expr input
+  expect_query_equal(
+    .input$unique(pl$col("a")$str$to_lowercase(), maintain_order = TRUE),
+    .input = pl$DataFrame(a = c("A", "a", "A", "a")),
+    pl$DataFrame(a = "A")
+  )
+
   expect_query_error(
     .input$unique("foobar", maintain_order = TRUE),
     df,


NEWS.md
```diff
@@ -2,8 +2,22 @@
 
 ## polars (development version)
 
+This is an update that corresponds to Python Polars 1.36.0, which includes significant internal changes.
```
etiennebacher, line 5: Not completely related to this PR but given the amount of big changes in 1.36.0, I guess there will be a 1.36.1 shortly after and we should target that for our next release.

NEWS.md
```diff
@@ -2,8 +2,22 @@
...
+- `<expr>$index_of()` now strictly checks types. Therefore, operations like `$index_of(NA)`
+  that worked previously will error if `NA` is converted to Bool type and the target column cannot be converted from Bool type.
+  Use `NULL` or `vctrs::unspecified(1)` instead of `NA` for such operations (#1662).
+
 ### New features
 
+- New method `<lazyframe>$pivot()` and `<dataframe>$pivot()`'s new argument `on_columns`
+  (#1662, [pola-rs/polars#25016](https://github.com/pola-rs/polars/pull/25016)).
+- `<lazyframe>$unique()` and `<dataframe>$unique()`'s `...` (dynamic dots) polars expressions
+  (#1662, [pola-rs/polars#25099](https://github.com/pola-rs/polars/pull/25099))
```
etiennebacher:
```suggestion
- `<lazyframe>$unique()` and `<dataframe>$unique()` now allow polars expressions in `...`
  (#1662, [pola-rs/polars#25099](https://github.com/pola-rs/polars/pull/25099))
```

NEWS.md
```diff
@@ -14,11 +28,27 @@
 - `<expr>$rolling_kurtosis()` (#1665).
 - `pl$explain_all()` to show a single optimized query plan from several input LazyFrames (#1666).
 - `<expr>$bin$reinterpret()` (#1664).
+- `<expr>$mode()` gains the `maintain_order` argument (#1662).
+- `pl$scan_parquet()` and `pl$read_parquet()` gain the `missing_columns` argument
+  to specify how to handle missing columns in the scanned files (#1662).
```
etiennebacher: I think that's duplicated with the bullet point about deprecation of `allow_missing_columns`

NEWS.md
```diff
@@ -14,11 +28,27 @@
...
+### Bug fixes
 
 - `<expr>$list$eval()` now properly errors (as documented) if the input is not a Polars expression (#1655).
 
+### Other changes
+
+- The experimental partitioning scheme class
+  (constructed with `pl$PartitionByKey()`, `pl$PartitionMaxSize()`, or `pl$PartitionParted()`)
+  is rewriten as S7 class (#1662).
+- `pl$element()` is rewriten in Rust (#1662, [pola-rs/polars#24885](https://github.com/pola-rs/polars/pull/24885)).
```
etiennebacher:
```suggestion
- `pl$element()` is rewritten in Rust (#1662, [pola-rs/polars#24885](https://github.com/pola-rs/polars/pull/24885)).
```

NEWS.md
etiennebacher, line 1: Can you add some tests for the new functions and the new arguments?

R/expr-expr.R
```diff
@@ -3948,19 +3955,46 @@ expr__drop_nulls <- function() {
...
+#' Returns a column with a separate row for every list element
+#'
+#' @inherit as_polars_expr return
+#' @inheritParams expr__explode
+#'
+#' @examples
+#' df <- pl$DataFrame(a = list(c(1, 2, 3), c(4, 5, 6)))
+#' df$select(pl$col("a")$list$explode())
+expr_list_explode <- expr__explode
+
```
etiennebacher: Shouldn't those be placed in `expr-array.R` and `expr-list.R`?

Also, for the array method, Python Polars uses `arr_explode()` and not `explode()`: https://github.com/pola-rs/polars/blob/main/py-polars/src/polars/expr/array.py#L762-L800
Isn't this a problem here?

R/expr-expr.R
```diff
@@ -4906,12 +4940,34 @@ expr__bitwise_xor <- function() {
...
-#'   null = pl$col("a")$index_of(NA),
+#'   null = pl$col("a")$index_of(NULL),
 #'   fiftyfive = pl$col("a")$index_of(55),
 #' )
 expr__index_of <- function(element) {
-  self$`_rexpr`$index_of(as_polars_expr(element, as_lit = TRUE)$`_rexpr`) |>
-    wrap()
+  wrap({
+    # TODO: @2.0 remove this
+    if (identical(element, NA)) {
```
etiennebacher, line 4935: Can you add a comment that this is not the same as `is.na(element)`?

R/generated-polars-version.R
```diff
@@ -1,6 +1,6 @@
 # Generated by dev script: do not edit by hand
 
-PY_VERSION <- "1.35.2"
+PY_VERSION <- "1.35.1"
```
etiennebacher, line 3: I'm probably missing something but why do we have this change?

NEWS.md
eitsupi: Definitely. I'm just putting a tentative value here, and I'll wait and see for a few days after 1.36.0.

R/generated-polars-version.R
eitsupi: 1.35.2 was not released from the main branch, and the version of the main branch (i.e. the branch where this Rust Polars commit was) remains 1.35.1.

R/expr-expr.R
eitsupi: Good catch.

This was done because when moving between files, `expr-array.R` is loaded before `expr-expr.R`.

As for `arr_explode`, it is outside the scope of this PR, so I think we should merge it as is and fix it later.

etiennebacher: > As for arr_explode, it is outside the scope of this PR, so I think we should merge it as is and fix it later.

But we already have `expr_arr_explode()`, it is deleted in this PR.

R/generated-polars-version.R
etiennebacher: Ok, thanks

R/expr-expr.R
eitsupi: > But we already have `expr_arr_explode()`, it is deleted in this PR.

What does that mean? I don't think it's been deleted.
Please check carefully.

eitsupi: What I have done here is simply combine `expr__explode`, `expr_arr_explode`, and `expr_list_explode` into one definition, since their contents are currently exactly the same.

etiennebacher: Right, I got confused between the function name and the call to the Rust function, all good.

NEWS.md
eitsupi: I've added some tests, and I'd like to merge this PR now since it's a blocker for all other work and tests can be added later.

etiennebacher (approved):
This looks great, I'll try to adapt `tidypolars` with this (in particular the sink functions and the lazy pivot) before the next release.

Thanks!

PR #1664, opened 2025-11-22
feat: Implement `<expr>$bin$reinterpret()`

etiennebacher:
Part of #1397 

This compiles properly but the output is not as expected. Using the example here: https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.bin.reinterpret.html#polars.Expr.bin.reinterpret

I have this:

```r
df <- pl$DataFrame(
  x = as_polars_series(c("\\x05\\x00\\x00\\x00", "\\x10\\x00\\x01\\x00"))$cast(pl$Binary)
)

df$with_columns(
  bin2int = pl$col("x")$bin$reinterpret(
    dtype = pl$Int32,
    endianness = "little"
  )
)

# shape: (2, 2)
# ┌─────────────────────┬─────────┐
# │ x                   ┆ bin2int │
# │ ---                 ┆ ---     │
# │ binary              ┆ i32     │
# ╞═════════════════════╪═════════╡
# │ b"\x05\x00\x00\x00" ┆ null    │
# │ b"\x10\x00\x01\x00" ┆ null    │
# └─────────────────────┴─────────┘
```

while the output in `bin2int` should be `5, 35`.

@eitsupi any idea why?

### DIFF
--- a/altdoc/mkdocs.yml
+++ b/altdoc/mkdocs.yml
@@ -462,6 +462,7 @@ nav:
             - bin_decode: man/expr_bin_decode.md
             - bin_encode: man/expr_bin_encode.md
             - bin_ends_with: man/expr_bin_ends_with.md
+            - bin_reinterpret: man/expr_bin_reinterpret.md
             - bin_size: man/expr_bin_size.md
             - bin_starts_with: man/expr_bin_starts_with.md
         - Categories:


eitsupi:
> which is not the case in Python Polars.

When I sent the DataFrame created with `as_polars_series(c("\\x05\\x00\\x00\\x00", "\\x10\\x00\\x01\\x00"))$cast(pl$Binary)` on R to Python, it became null, so I couldn't understand what was going on.

R/expr-binary.R
```diff
@@ -182,3 +182,34 @@ expr_bin_size <- function(unit = c("b", "kb", "mb", "gb", "tb")) {
...
+#' df$with_columns(
+#'   bin2int = pl$col("x")$bin$reinterpret(
+#'     dtype = pl$UInt8,
+#'     endianness = "little"
+#'   )
+#' )
+expr_bin_reinterpret <- function(..., dtype, endianness = c("little", "big")) {
+  wrap({
+    check_dots_empty0(...)
+    dtype <- as_polars_dtype_expr(dtype)
```
eitsupi, line 211: Would it be safer to check the class of the dtype here?

tests/testthat/test-expr-binary.R
```diff
@@ -67,3 +67,24 @@ test_that("bin$size()", {
     error = TRUE
   )
 })
+
+test_that("bin$reinterpret()", {
+  df <- pl$DataFrame(x = blob::as_blob(c(5L, 35L)))
```
eitsupi, line 72: Maybe skip if not installed is needed.

eitsupi (changes requested):

tests/testthat/test-expr-binary.R
etiennebacher: This is actually not necessary anymore as of `testthat` 3.3.0. See the third bullet point here: https://tidyverse.org/blog/2025/11/testthat-3-3-0/#other-new-features

R/expr-binary.R
etiennebacher: I don't understand, the second snapshot test shows a proper error message when `dtype` isn't of the right type. Do you have another example to test in mind?

tests/testthat/test-expr-binary.R
eitsupi: I've read it but I'm not sure how it actually works.

R/expr-binary.R
eitsupi: Ok, thanks.

PR #1686, opened 2025-12-08
chore: bump polars to post py-1.36.0 unreleased version

eitsupi:


tests/testthat/test-lazyframe-frame.R
```diff
@@ -2213,7 +2211,7 @@ test_that("rolling: arg 'offset' works", {
     )$select("sum_a", "min_a", "max_a"),
     .input = df,
     pl$DataFrame(
-      sum_a = c(2, 2, 2, NA, NA, NA),
+      sum_a = c(2, 2, 2, 0, 0, 0),
```
eitsupi, line 2214: @etiennebacher Do you think this is a bug?

tests/testthat/test-expr-expr.R
```diff
@@ -2552,8 +2552,8 @@ test_that("rolling, arg offset", {
   expect_equal(
     out,
     pl$DataFrame(
-      sum_a_offset1 = c(11, 11, 11, 2, NA, NA),
-      sum_a_offset2 = c(2, 2, 2, NA, NA, NA)
+      sum_a_offset1 = c(11, 11, 11, 2, 0, 0),
+      sum_a_offset2 = c(2, 2, 2, 0, 0, 0)
```
eitsupi, line 2556: @etiennebacher Do you think this is a bug?

eitsupi:
Ah, maybe an intended change?
pola-rs/polars#25494

> This also fixes an undocumented issue whereby the rolling group sum kernel returns None on empty groups instead of 0.

tests/testthat/test-lazyframe-frame.R
etiennebacher: I remember seeing this in one of the recent rolling issues, I'll try to find it

etiennebacher: There you go: https://github.com/pola-rs/polars/pull/25494

> This also fixes an undocumented issue whereby the rolling group sum kernel returns None on empty groups instead of 0.

If I remember correctly, it's the same reason why `sum([]) == 0`.

eitsupi: Thanks, I may find that https://github.com/pola-rs/r-polars/pull/1686#issuecomment-3632522653

etiennebacher: Well you found it already 😄

eitsupi: > If I remember correctly, it's the same reason why `sum([]) == 0`.

Oh, I get it. Thanks for your quick reply!

PR #1688, opened 2025-12-09
ci(lint-r): Add jarl check

eitsupi:


eitsupi:
Oops...I forgot to that we have to disable some rules as like for lintr...

PR #1709, opened 2026-01-08
feat: `as_nanoarrow_array_stream(<lazyframe>)`

eitsupi:
We can now output an Arrow C stream without performing heavy computation.

```r
lf <- pl$scan_parquet(
  "http://github.com/duckdb/duckdb-data/releases/download/v1.0/lineitemsf1.snappy.parquet"
)$select(
  revenue = (polars::pl$col("l_extendedprice") * polars::pl$col("l_discount"))$sum()
)

# Traditional way
tictoc::tic()
lf$collect(engine = "streaming") |>
  nanoarrow::as_nanoarrow_array_stream()
tictoc::toc()
#> <nanoarrow_array_stream struct<revenue: double>>
#>  $ get_schema:function ()  
#>  $ get_next  :function (schema = x$get_schema(), validate = TRUE)  
#>  $ release   :function ()  
#> 1.055 sec elapsed

# New way
tictoc::tic()
lf |>
  nanoarrow::as_nanoarrow_array_stream()
tictoc::toc()
#> <nanoarrow_array_stream struct<revenue: double>>
#>  $ get_schema:function ()  
#>  $ get_next  :function (schema = x$get_schema(), validate = TRUE)  
#>  $ release   :function ()  
#> 0.018 sec elapsed
```

Same as pola-rs/polars#25994, based on pola-rs/polars#25918

### DIFF
--- a/R/lazyframe-s3-nanoarrow.R
+++ b/R/lazyframe-s3-nanoarrow.R
@@ -1,3 +1,5 @@
+# nolint start: object_name_linter
+
 #' @param maintain_order `r lifecycle::badge("experimental")`
 #'   Maintain the order in which data is processed. Setting this to `FALSE` will be slightly faster.
 #' @param chunk_size `r lifecycle::badge("experimental")`
@@ -46,3 +48,5 @@ as_nanoarrow_array_stream.polars_lazy_frame <- function(
   )
   stream
 }
+
+# nolint end


R/lazyframe-s3-nanoarrow.R
```diff
@@ -0,0 +1,28 @@
+#' @rdname s3-as_nanoarrow_array_stream
+as_nanoarrow_array_stream.polars_lazy_frame <- function(
+  x,
+  ...,
+  schema = NULL,
+  polars_compat_level = c("newest", "oldest"),
+  maintain_order = FALSE,
+  chunk_size = NULL
```
eitsupi, line 17: Since this is an S3 method, there is a risk of argument name conflicts with other methods.
Perhaps it would be better to pass in a single S7 class object.

etiennebacher: I don't understand how S3 dispatch can lead to conflict, can you give me an example where another method (fake for now) would lead to one?

eitsupi: That is, if another function uses an argument with the same name for a different purpose, which is of course unlikely.

```r
func.foo <- function(x, ..., arg = TRUE) {}
func.bar <- function(x, ..., arg = 1) {}

func(unknown, arg = TRUE) # If unkown is bar may cause error
```

NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 ## polars (development version)
 
+### New features
+
+- New S3 method `nanoarrow::as_nanoarrow_array_stream(<lazyframe>)` for export LazyFrame via Arrow C stream interface
```
etiennebacher:
```suggestion
- New S3 method `nanoarrow::as_nanoarrow_array_stream(<lazyframe>)` to export LazyFrame via Arrow C stream interface
```

R/lazyframe-s3-nanoarrow.R
etiennebacher: Ah ok  I see, thanks

NEWS.md
etiennebacher: Duplicate of copilot comment

Issue #1597, opened 2025-10-13
Factor columns seem to lose their physical representation during conversion


mbanand:
Factor columns seem to lose the correspondence to physical representation during conversion to or from polars dataframe.

```r
library(polars)
library(tidypolars)
library(dplyr)

df <- tibble(x = c(1, 2, 3), y = factor(c("a", "b", "c"), 
                                    levels = c("b", "c", "a"))) 
str(df$y)

#> Factor w/ 3 levels "b","c","a": 3 1 2
```

But:
```r
df |> as_polars_df() |> pull(y) |> str()

#> Factor w/ 3 levels "a","b","c": 1 2 3
```

The same thing happens when converting from polars dataframe to R also:
```r
df <- tibble(x = 1:3, y = c("a", "b", "c")) |> as_polars_lf()
df <- df |>
  mutate(across(y, \(x) x$cast(pl$Enum(c("b", "a", "c"))),
                .names = "{.col}_fac")) |>
  # Add column to check the levels
  mutate(across(y_fac, \(x) x$to_physical(), .names = "{.col}_o"))
df |> compute()

#> 
shape: (3, 4)
┌─────┬─────┬───────┬─────────┐
│ x   ┆ y   ┆ y_fac ┆ y_fac_o │
│ --- ┆ --- ┆ ---   ┆ ---     │
│ i32 ┆ str ┆ enum  ┆ u8      │
╞═════╪═════╪═══════╪═════════╡
│ 1   ┆ a   ┆ a     ┆ 1       │
│ 2   ┆ b   ┆ b     ┆ 0       │
│ 3   ┆ c   ┆ c     ┆ 2       │
└─────┴─────┴───────┴─────────┘
```

shows the correct levels, but:
```r
df |> pull(y_fac) |> str()

#> Factor w/ 3 levels "a","b","c": 1 2 3
```

loses it. 

I'm not sure if this is a bug, or a minor feature enhancement, and it's not a big deal, but it would be nice to maintain the definition of the levels during conversion.

Issue #1146, opened 2024-06-22
Polars Categolical/Enum -> R factor conversion via Rust


eitsupi:
Currently the conversion to the factor type is performed by calling the `factor()` function in R.
https://github.com/pola-rs/r-polars/blob/6eac27a0766d2b6ca92a72c1c7fa76eaeb58bb98/src/rust/src/conversion_s_to_r.rs#L113-L118

but I think it would be possible to perform the conversion on Rust by porting the process from the arrow package.
https://github.com/apache/arrow/blob/26724ab787d12854f109d01571638b099f2bc8a5/r/src/array_to_vector.cpp#L563-L717

andyquinterom:
Is there any progress on this? I might be able to take the responsability

eitsupi:
Thanks. Probably no one is working on this at this time.

BTW, we are shifting our focus to the next branch (under development in https://github.com/eitsupi/neo-r-polars) on adding new features, so it might be better to have them work on that.
(Currently the main branch is stuck on updating Rust Polars #1271)

PR #1723, opened 2026-01-27
feat!: Convert R factors to Polars Enum and keep all levels when converting Polars Enum to R factors

etiennebacher:
I think we should improve the R -> Polars and Polars -> R conversions for factors.

### Categorical variables are weird in Polars

With the Categorical refactor from a few months ago in upstream Polars, there's a single pool for all categories, which leads to this kind of situation where levels of all columns are mixed together:

``` r
library(polars)

dat <- pl$DataFrame(
  x = factor(c("a", "b"), levels = c("a", "b", "c")),
  y = factor(c("d", "e"), levels = c("d", "e"))
)

dat$select(pl$col("x")$cat$get_categories())
#> shape: (4, 1)
#> ┌─────┐
#> │ x   │
#> │ --- │
#> │ str │
#> ╞═════╡
#> │ a   │
#> │ b   │
#> │ d   │
#> │ e   │
#> └─────┘
```
I think R factors are much closer to Enums than to Categoricals (honestly I'm not even sure that Polars Categoricals are useful anymore):

```r
dat <- pl$DataFrame(
  x = factor(c("a", "b"), levels = c("a", "b", "c")),
  y = factor(c("d", "e"), levels = c("d", "e"))
)$cast(x = pl$Enum(c("a", "b", "c")), y = pl$Enum(c("d", "e")))

dat$select(pl$col("x")$cat$get_categories())
#> shape: (3, 1)
#> ┌─────┐
#> │ x   │
#> │ --- │
#> │ str │
#> ╞═════╡
#> │ a   │
#> │ b   │
#> │ c   │
#> └─────┘
```

-> **R factors should be converted to Polars enums**

### Converting a Categorical to R loses the levels information

This happens a lot in `tidypolars` tests: I start from `iris` (in DataFrame format), filter out all rows that match a certain Categorical value, and convert to `data.frame`. But then this loses the information about the level that was filtered out while the same operation in `tidyverse` keeps the levels info.

In this example, we don't even need to filter since not all levels are represented in the data:
```r
dat <- pl$DataFrame(
  x = factor(c("a", "b"), levels = c("a", "b", "c")),
  y = factor(c("d", "e"), levels = c("d", "e"))
)

dat |>
  as.data.frame() |>
  dplyr::pull(x) |>
  levels()
#> [1] "a" "b"

data.frame(x = factor(c("a", "b"), levels = c("a", "b", "c"))) |>
  dplyr::pull(x) |>
  levels()
#> [1] "a" "b" "c"
```

It would be nice to keep all the levels here.

-> **Polars enums should keep all the levels information when converted to R factors**

---

@eitsupi This PR is incomplete, I didn't update the tests for now because I wanted to see if you're ok with it first (and this would be a breaking change). There are about ~60 failures but it seems they would all be solved by an extra `$cast()` or a snapshot update.

Almost all changes have been done by Claude, but it looks okay to me.

Same examples with this PR:

``` r
library(polars)

dat <- pl$DataFrame(
  x = factor(c("a", "b"), levels = c("a", "b", "c")),
  y = factor(c("d", "e"), levels = c("d", "e"))
)

dat
#> shape: (2, 2)
#> ┌──────┬──────┐
#> │ x    ┆ y    │
#> │ ---  ┆ ---  │
#> │ enum ┆ enum │  # ----> now "enum", not "cat" anymore
#> ╞══════╪══════╡
#> │ a    ┆ d    │
#> │ b    ┆ e    │
#> └──────┴──────┘

dat$select(pl$col("x")$cat$get_categories())
#> shape: (3, 1)
#> ┌─────┐
#> │ x   │
#> │ --- │
#> │ str │
#> ╞═════╡
#> │ a   │
#> │ b   │
#> │ c   │
#> └─────┘

dat |>
  as.data.frame() |>
  dplyr::pull(x) |>
  levels()
#> [1] "a" "b" "c"
```

Related to #1597 and #1146

### DIFF
--- a/R/as_polars_series.R
+++ b/R/as_polars_series.R
@@ -253,14 +253,20 @@ as_polars_series.raw <- function(x, name = NULL, ...) {
     wrap()
 }
 
-#' @param factor_as_enum if `FALSE` (default), R factors are converted to Polars'
-#' [Categorical][pl__Categorical]. If `TRUE`, they are converted to Polars'
-#' [Enum][pl__Enum].
+#' @param factor_as_enum `r lifecycle::badge("experimental")` if `FALSE`
+#' (default), R factors are converted to Polars [Categorical][pl__Categorical].
+#' If `TRUE`, they are converted to Polars [Enum][pl__Enum].
 #' @rdname as_polars_series
 #' @export
 as_polars_series.factor <- function(x, name = NULL, ..., factor_as_enum = FALSE) {
+  if (missing(factor_as_enum)) {
+    opt <- getOption("polars.factor_as_enum")
+    if (!is.null(opt)) {
+      factor_as_enum <- opt
+    }
+  }
   opt <- getOption("polars.factor_as_enum")
-  factor_as_enum <- opt %||% factor_as_enum
+
   PlRSeries$new_str(name %||% "", as.character(x))$cast(
     if (isTRUE(factor_as_enum)) pl$Enum(levels(x))$`_dt` else pl$Categorical()$`_dt`,
     strict = TRUE
--- a/R/datatypes-classes.R
+++ b/R/datatypes-classes.R
@@ -223,9 +223,6 @@ pl__Categorical <- function(ordering = deprecated()) {
 #' Should not contain `NA` values and all values should be unique.
 pl__Enum <- function(categories) {
   wrap({
-    if (is_polars_series(categories)) {
-      return(PlRDataType$new_enum(categories$`_s`))
-    }
     check_character(categories, allow_na = FALSE)
 
     if (anyDuplicated(categories)) {
--- a/man/as_polars_series.Rd
+++ b/man/as_polars_series.Rd
@@ -117,9 +117,9 @@ When not specified, name is set to an empty string.}
 
 \item{...}{Additional arguments passed to the methods.}
 
-\item{factor_as_enum}{if \code{FALSE} (default), R factors are converted to Polars'
-\link[=pl__Categorical]{Categorical}. If \code{TRUE}, they are converted to Polars'
-\link[=pl__Enum]{Enum}.}
+\item{factor_as_enum}{\ifelse{html}{\href{https://lifecycle.r-lib.org/articles/stages.html#experimental}{\figure{lifecycle-experimental.svg}{options: alt='[Experimental]'}}}{\strong{[Experimental]}} if \code{FALSE}
+(default), R factors are converted to Polars \link[=pl__Categorical]{Categorical}.
+If \code{TRUE}, they are converted to Polars \link[=pl__Enum]{Enum}.}
 
 \item{strict}{A logical value to indicate whether throwing an error when
 the input \link{list}'s elements have different data types.
--- a/tests/testthat/_snaps/as_polars_series.md
+++ b/tests/testthat/_snaps/as_polars_series.md
@@ -438,6 +438,18 @@
       	"2"
       ]
 
+---
+
+    Code
+      as_polars_series(factor(1:2), factor_as_enum = FALSE)
+    Output
+      shape: (2,)
+      Series: '' [cat]
+      [
+      	"1"
+      	"2"
+      ]
+
 # as_polars_series.default throws an error
 
     Code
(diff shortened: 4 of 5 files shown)


R/as_polars_series.R
```diff
@@ -253,11 +253,16 @@ as_polars_series.raw <- function(x, name = NULL, ...) {
...
 
+#' @param factor_as_enum if `FALSE` (default), R factors are converted to Polars'
+#' [Categorical][pl__Categorical]. If `TRUE`, they are converted to Polars'
+#' [Enum][pl__Enum].
 #' @rdname as_polars_series
 #' @export
-as_polars_series.factor <- function(x, name = NULL, ...) {
+as_polars_series.factor <- function(x, name = NULL, ..., factor_as_enum = FALSE) {
+  opt <- getOption("polars.factor_as_enum")
+  factor_as_enum <- opt %||% factor_as_enum
```
eitsupi: What does this intended to?

R/datatypes-classes.R
```diff
@@ -223,6 +223,9 @@ pl__Categorical <- function(ordering = deprecated()) {
 #' Should not contain `NA` values and all values should be unique.
 pl__Enum <- function(categories) {
   wrap({
+    if (is_polars_series(categories)) {
+      return(PlRDataType$new_enum(categories$`_s`))
+    }
```
eitsupi: Please do not.

eitsupi (changes requested):

R/as_polars_series.R
```diff
@@ -253,11 +253,16 @@ as_polars_series.raw <- function(x, name = NULL, ...) {
     wrap()
 }
 
+#' @param factor_as_enum if `FALSE` (default), R factors are converted to Polars'
+#' [Categorical][pl__Categorical]. If `TRUE`, they are converted to Polars'
+#' [Enum][pl__Enum].
```
eitsupi: Please mark as experimental.

R/as_polars_series.R
etiennebacher: The idea is to allow one to set `options(polars.factor_as_enum = TRUE)` but I think the logic is wrong. We should use the global option if `factor_as_enum` is missing in the call.

R/as_polars_series.R
etiennebacher: done

R/datatypes-classes.R
etiennebacher: done

R/as_polars_series.R
etiennebacher: Note that if we use `use_option_if_missing()` here, we get duplicated messages if we convert a `data.frame` with several factors to a polars dataframe.

```r
  factor_as_enum <- use_option_if_missing(
    factor_as_enum,
    missing(factor_as_enum),
    FALSE,
    "polars."
  )
```

```r
options(polars.factor_as_enum = TRUE)

data.frame(x = factor(1:2), y = factor(1:2)) |> 
  as_polars_df()
```
```
`factor_as_enum` is overridden by the option "polars.factor_as_enum" with `TRUE`
`factor_as_enum` is overridden by the option "polars.factor_as_enum" with `TRUE`
shape: (2, 2)
┌──────┬──────┐
│ x    ┆ y    │
│ ---  ┆ ---  │
│ enum ┆ enum │
╞══════╪══════╡
│ 1    ┆ 1    │
│ 2    ┆ 2    │
└──────┴──────┘
```

eitsupi: > Note that if we use use_option_if_missing() here, we get duplicated messages if we convert a data.frame with several factors to a polars dataframe.

I think you need to handle the argument here to avoid that.

https://github.com/pola-rs/r-polars/blob/28fbc068ee066e2677c3e786ffd702894ad2ff1e/R/as_polars_df.R#L178-L209

etiennebacher: Do you mean I should add `factor_as_enum` after `...` in the function definition, or that I should just call `getOption("polars.factor_as_enum")` in `as_polars_df.list`?

eitsupi: Thinking about it again, I feel that overriding with options is dangerous as it could potentially destroy behavior everywhere.
Could you keep it as it cannot be used except by explicit specification?

PR #1724, opened 2026-02-03
feat: Bump polars to 1.38.0

etiennebacher:
Nowhere near ready, just fixed compilation errors to see if tests pass.


eitsupi:
As an experiment, I'll let Claude Code do the version bumping.
I'll be working on a different branch, so if it works, I might not need to do this here.

PR #1726, opened 2026-02-08
chore: bump polars to py-1.38.1

eitsupi:
Supersede #1724

## Summary

Bump polars from py-1.37.1 to py-1.38.1.

- Update polars rev hash in `Cargo.toml`
- Adapt to upstream API changes (sink options, IPC scan options, retry config, etc.)
- Deprecate `retries` and `file_cache_ttl` params (converted to `storage_options` entries)
- Deprecate `Expr$flatten()` in favor of `Expr$list$explode()`
- Expose `expand_patterns` in `cs$by_name()` selector
- Remove `per_partition_sort_by` from deprecated partition classes

### Follow-up (separate PR)

New features available in py-1.38.1 but not yet exposed:

- `Expr$bin$get()`
- `str$split()` literal/strict + regex support
- `Expr$quantile()` list support
- `scan_lines()` / `read_lines()`
- sink CSV/NDJSON compression
- `record_batch_statistics` (IPC)
- `DataFrame(height=...)`

### Known issue: WASM build failure

The WASM (emscripten) build is failing due to an upstream change in polars. [pola-rs/polars#26353](https://github.com/pola-rs/polars/pull/26353) added an unconditional `tokio` dependency to `polars-plan`, which pulls in `mio` — a crate that does not support the `wasm32-unknown-emscripten` target. This cannot be worked around in r-polars because removing the `lazy` feature for emscripten would make almost all functionality non-operational. An upstream fix is required to make `tokio` optional in `polars-plan`.

## Test plan

- [x] `task test-all` passes (FAIL 0 / PASS 3988)
- [x] `task build-documents` succeeds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/testthat/test-expr-expr.R
+++ b/tests/testthat/test-expr-expr.R
@@ -1853,6 +1853,11 @@ test_that("rolling_*_by: arg 'closed'", {
     error = TRUE
   )
 
+  expect_equal(
+    df$select(pl$col("a")$rolling_rank_by("date", window_size = "2d", closed = "right")),
+    pl$DataFrame(a = c(1, 2, 2, 2, 2, 2))
+  )
+
   expect_snapshot(
     df$select(pl$col("a")$rolling_min_by("date", window_size = "2d", closed = "foo")),
     error = TRUE


tests/testthat/test-expr-expr.R
```diff
@@ -1844,11 +1843,16 @@ test_that("rolling_*_by: arg 'closed'", {
         closed = "left",
         interpolation = "linear"
       ),
-      rank = pl$col("a")$rolling_rank_by("date", window_size = "2d", closed = "left"),
```
etiennebacher, line 1847: This is now checked for an error but we no longer check the output of `rolling_rank_by()` anymore, can you modify the example to check the values with `closed = "right"` for instance?

etiennebacher (approved):
Thanks, this looks like a good task for claude indeed

eitsupi:
Even Claude Code took several hours, so if I had done it manually, I probably would have wasted my weekend.

Issue #1729, opened 2026-02-08
New features from py-polars 1.38.0 and 1.38.1


etiennebacher:
Extracted from #1726:

- [x] Expr$bin$get() (#1731)
- [x] str$split() literal/strict + regex support (#1730)
- ~Expr$quantile() list support~ 
- [x]  scan_lines() / read_lines() (#1732)
- [x] sink CSV/NDJSON compression (#1733, #1735)
- ~record_batch_statistics (IPC)~
- ~DataFrame(height=...)~

I can work on that.

etiennebacher:
@eitsupi fyi we already have several missing arguments in `pl$DataFrame` compared to the Python version (`nan_to_null`, etc.) so I'm dropping the new arg `height` from the TODO list in the OP and opening a new issue for this instead: https://github.com/pola-rs/r-polars/issues/1734

PR #1730, opened 2026-02-08
feat: Add arguments `literal` and `strict` in `<expr>$str$split()`

etiennebacher:
Part of #1729


R/expr-string.R
```diff
@@ -895,13 +899,59 @@ expr_str_count_matches <- function(pattern, ..., literal = FALSE) {
...
+    if (isTRUE(inclusive)) {
+      return(
+        self$`_rexpr`$str_split_inclusive(by) |>
+          wrap()
+      )
+    } else {
+      return(
+        self$`_rexpr`$str_split(by) |>
+          wrap()
+      )
```
eitsupi: returns seem redundant.

Issue #1405, opened 2025-07-12
Implement Auto-completion again


eitsupi:
Auto-completion was present in v0, but not in the 1.0.0 pre-release version. This needs to be re-implemented.

etiennebacher:
I can try to take a look in the coming week but no promises. I also don't really know if the autocompletion mechanism will be the same in Positron as in RStudio.

Edit: https://github.com/posit-dev/positron/discussions/12836

etiennebacher:
@eitsupi should we add a package startup message to mention `polars_activate_code_completion()`, as we did in the old r-polars?

eitsupi:
Since it currently only works with RStudio and there haven't been any requests for it so far, I don't feel it's necessary to display a startup message.

PR #1768, opened 2026-04-05
feat: Add autocompletion in RStudio

etiennebacher:
Related to #1405

This is mostly adapted from the code in the [first r-polars](https://github.com/rpolars/r-polars0). Differences:

- I removed the  "native" mode because it is hard to handle and to have feature parity with the RStudio implementation, and I doubt many people would use it
- the old implementation suggested column names, e.g. to insert `pl$col("x")` directly but 1) this would be annoying when there are more than just a few columns, and 2) it could insert expressions where `...` don't expect them.

 RStudio, you can now see:

- suggested DataFrame/LazyFrame functions and args
- suggested args for expr

It is not possible to test this given that it only works in RStudio.

[Screencast from 2026-04-05 18-25-18.webm](https://github.com/user-attachments/assets/c30bbac3-a7fe-4936-9170-7e723d007841)

It seems it's not possible to have custom autocompletion in Positron for now (see link to discussion in #1405).


R/autocompletion.R
```diff
@@ -0,0 +1,281 @@
+#' Polars code completion
+#'
+#' This only works in RStudio.
```
eitsupi, line 5: Could you add the experimental badge?

R/autocompletion.R
```diff
@@ -0,0 +1,281 @@
...
+}
+
+#' @export
+#' @rdname polars_code_completion_activate
+polars_code_completion_deactivate <- function() {
+  if (!is_rstudio()) {
+    return(invisible(NULL))
+  }
+  .rs_complete$deactivate()
+  invisible(NULL)
```
eitsupi: Can this be simplified?

R/autocompletion.R
```diff
@@ -0,0 +1,281 @@
...
+    return(invisible())
+  }
+
+  local(envir = rs, {
+    # 1 - autocompletion for function arguments (inside call).
+    # For instance, "pl$col('x')$str$count_matches(<TAB>)" should show the
+    # two args of str$count_matches().
+
+    # save original function here
+    .rs.getCompletionsFunction_polars_orig <- .rs.getCompletionsFunction
```
eitsupi: Is there any way to change the name of this object?

R/autocompletion.R
```diff
@@ -0,0 +1,281 @@
...
+      numCommas,
+      envir = parent.frame(),
+      object = .rs.resolveObjectFromFunctionCall(functionCall, envir)
+    ) {
+      # if Rstudio failed to immediately resolve the object, do a full
+      # evaluation of the entire lhs line/section which creates the calling
+      # function
+      col_results <- NULL
+      if (isFALSE(object)) {
+        lhs <- polars:::.rs_complete$eval_lhs_string(string, envir)
```
eitsupi, line 161: `polars:::` is needed?

eitsupi (approved):

R/autocompletion.R
etiennebacher: Not sure what you mean, I don't see how to simplify further.

R/autocompletion.R
etiennebacher: Yes, but what name would you prefer?

R/autocompletion.R
etiennebacher: Yes, this is needed since we don't export `.rs_complete` and doing `<TAB>` needs to trigger this function. I can't explain much more than that, but I tried to remove the `polars:::` and I get this error on `<TAB>`:

``` 
Error in .rs.getCompletionsDollar(token, string[[1]], functionCall, envir,  : 
  object '.rs_complete' not found
```

R/autocompletion.R
eitsupi: Done in 3ce68bf

R/autocompletion.R
eitsupi: Done in 3ce68bf

R/autocompletion.R
```diff
@@ -0,0 +1,278 @@
...
+      "polars_chained_then",
+      "polars_sql_context"
+    )
+  )
+}
+
+# decide if some function/method recursively has the polars namespace as parent
+# environment.
+.rs_complete$is_polars_function <- function(x) {
+  is_function(x) && env_inherits(environment(x), ns_env("polars"))
```
eitsupi: If we were to create a package with the index u64 under a different name in the future, I think we shouldn't hardcode the package name. However, this seemed like the easiest way to do it right now.

eitsupi:
@etiennebacher I've refactored it a bit; could you take a look?

etiennebacher:
Thanks @eitsupi, this looks good. I've fixed the tests and also hidden the message in some cases, do you agree with that?

Also, what does the "bk" mean in `.rs.getCompletionsFunction.bk.polars`?

eitsupi:
> Also, what does the "bk" mean in `.rs.getCompletionsFunction.bk.polars`?

Oh, I meant to "back up".
"bak" might be a more common term.

etiennebacher:
Never saw this abbreviation :smile: do you mind if I just put `.rs.getCompletionsFunction.backup.polars`?

eitsupi:
Of course, that's fine.

I think I probably forgot to update the tests, which is why they're not passing CI...

etiennebacher:
I'll take care of the tests

tests/testthat/test-autocompletion.R
```diff
@@ -4,10 +4,15 @@ test_that("verbose arg works", {
...
-  expect_message(
-    polars_code_completion_deactivate(),
-    "Polars code completion"
-  )
+
+  if (inside_rstudio()) {
+    expect_message(
+      polars_code_completion_deactivate(),
+      "Polars code completion"
+    )
```
eitsupi, line 12: Shouldn't this be a mock test?

etiennebacher: The thing is that we cannot set  `inside_rstudio()` to `TRUE` at the beginning of the `test_that()` block because  the activation would fail. And if we do this:

```r
  local_mocked_bindings(inside_rstudio = function(...) TRUE)
  expect_message(
    polars_code_completion_deactivate(),
    "Polars code completion"
  )
```

then we're in a case where we're mimicking that we're in RStudio but `.rs_complete$is_activated()` is `FALSE` and therefore the message will not be shown.

PR #1771, opened 2026-04-10
chore: Bump `rlang` requirement to >= 1.2.0

etiennebacher:
Some standalone type checks have been removed since they are available in `rlang` 1.2.0. Without this change, `task build-documents` would always generate changes in the standalone type checks file if you have `rlang` >= 1.2.0 installed.


etiennebacher:
I don't understand these snapshot failures, they're not related to this PR since they also appear in others

eitsupi (approved):
Thanks!

PR #1777, opened 2026-04-18
chore: bump polars to py-1.40.1

etiennebacher:


etiennebacher:
@eitsupi just FYI I won't be able to finish this one before a few days (at least) so feel free to push to this branch if you want, otherwise I'll just finish it later.

eitsupi:
Thank you. I have some free time today, so let's take a look.

eitsupi:
We should switch to py-1.40.1...

etiennebacher:
Thanks for your help on this!

PR #1789, opened 2026-05-26
chore: bump polars to py-1.41.2

etiennebacher:


eitsupi:
Thanks for working on this!

Recent versions of Polars have become too bloated for my machine to debug build, so I couldn't work with them.
It also seems like there are resource shortages in CI.

I think we'll need to update all the infrastructure and switch to release builds...

eitsupi (approved):

etiennebacher:
Yeah it's really painful to recompile locally for me too, especially because I often run out of space. However we rarely hit this issue in CI and it would be quite  annoying to use release builds in CI, we can try the action they use in polars first: https://github.com/pola-rs/polars/pull/27769

eitsupi:
I know about it, but it hasn't been updated in years, so I stopped using it.

Also, it only works on Linux, but I think it's not sufficient these days, even on macOS, where there isn't enough space.

PR #1804, opened 2026-06-27
chore: bump polars to py-1.42.0

eitsupi:
This PR updates the upstream Polars dependency from py-1.41.2 to py-1.42.0
(94 commits upstream).

## Changes

- **Cargo.toml**: Update rev to `5e9cf5c0`, mirror upstream `[patch.crates-io]`
  changes (`object_store` branch `feat_with_dns_resolver`, add `tikv-jemalloc-sys`
  patch for jemalloc Makefile bug fix)
- **Deprecation**: Emit warning when `empty_as_null` is not explicitly set in
  `$explode()` — the default will change from `TRUE` to `FALSE` in Polars 2.0
  ([pola-rs/polars#28040](https://github.com/pola-rs/polars/pull/28040)).
  Affects: `<expr>$explode()`, `<expr>$list$explode()`, `<expr>$arr$explode()`,
  `<lazyframe>$explode()`, `<dataframe>$explode()`
- **DSL schema hash**: Updated snapshot for new expressions added upstream
  (`is_sorted`, etc.)

## Follow-up tasks (separate PRs)

- Route `polars_warn!` to R's `warning()` system
  ([#1641](https://github.com/pola-rs/r-polars/issues/1641))
- New features deferred: `Expr.is_sorted` ([pola-rs/polars#26708](https://github.com/pola-rs/polars/pull/26708)), `DataFrame.is_sorted()` ([pola-rs/polars#27870](https://github.com/pola-rs/polars/pull/27870)), strict mode, out-of-core spilling

PR #1805, opened 2026-06-27
feat: route `polars_warn!` to R `warning()` system

### LINKED ISSUE #1641: Convert Rust warnings to R warnings
Like Python Polars: https://github.com/pola-rs/polars/blob/82f97416776accc17eaa40e123d291cbe048c2ef/crates/polars-python/src/exceptions.rs

eitsupi:
Close #1641

## Summary

Rust's `polars_warn!` macro output was previously going to stderr, bypassing R's condition system. This PR routes those warnings into R's `warning()` system with proper condition classes.

### How it works

1. **Thread-safe queue** (`src/rust/src/warn.rs`): A `parking_lot::Mutex<Vec<QueuedWarning>>` accumulates warnings from any thread (including rayon worker threads). Direct R API calls from non-main threads would be UB, so warnings are only queued there.

2. **`#[savvy_init]` hook**: Registers our queue function with `polars_error::set_warning_function()` at package load time.

3. **`wrap()` drains the queue** (`R/utils-wrap.R`): After every Polars operation, `drain_warnings()` is called. Each warning is emitted as an rlang condition with class `c("<category>", "polars_warning")`.

4. **On error, queue is also drained** to prevent warning leakage between operations.

### Condition classes

| Polars variant | R condition class |
|---|---|
| `Deprecation` | `polars_deprecation_warning` |
| `UserWarning` | `polars_user_warning` |
| `CategoricalRemappingWarning` | `polars_categorical_remapping_warning` |
| `MapWithoutReturnDtypeWarning` | `polars_map_without_return_dtype_warning` |

### Performance

`drain_warnings()` with an empty queue costs ~0.42 µs (NULL fast path, no R allocations).

To measure the end-to-end impact, `wrap()` was benchmarked before and after the change using a realistic lazy query on a 100k-row DataFrame (filter + group_by + agg + sort + collect, triggering ~6 `wrap()` calls per iteration), run N=300 times in alternating order (with/without drain) after 10 warmup iterations. `assignInNamespace()` was used to patch the installed `wrap` function at runtime to remove the drain call, so both variants ran against the same installed binary.

| | median | sd |
|---|---|---|
| With `drain_warnings()` | 19.000 ms | 2.157 ms |
| Without `drain_warnings()` | 19.000 ms | 1.657 ms |

The theoretical overhead of 6 × 0.42 µs ≈ 2.5 µs is ~800× smaller than the measurement noise (±2 ms), so no performance regression is detectable.


tests/testthat/test-expr-expr.R
```diff
@@ -2146,7 +2156,7 @@ test_that("clip", {
   # clip() works with temporal
   df <- pl$DataFrame(foo = as.Date(c("2020-01-01", "2020-01-02")))
   expect_equal(
-    df$select(clipped = pl$col("foo")$clip(lower_bound = pl$lit("2020-01-02"))),
+    df$select(clipped = pl$col("foo")$clip(lower_bound = pl$lit(as.Date("2020-01-02")))),
```
etiennebacher, line 2159: Why is this change necessary? I don't think it was throwing a warning before and I don't see API changes in this PR.

tests/testthat/test-series-to_r_vector.R
```diff
@@ -295,7 +295,10 @@ patrick::with_parameters_test_that(
       "2212-01-01T12:34:56.123456789"
     )
 
-    series_naive_time <- as_polars_series(chr_vec)$cast(pl$Datetime(time_unit))
+    series_naive_time <- as_polars_series(chr_vec)$str$to_datetime(
+      time_unit = time_unit,
+      format = "%Y-%m-%dT%H:%M:%S%.f"
+    )
```
etiennebacher, line 301: Same

tests/testthat/test-expr-expr.R
```diff
@@ -569,6 +569,16 @@ test_that("cast", {
     pl$DataFrame(big = NA_integer_)
   )
 
+  # casting String to Date/Datetime emits a Polars deprecation warning routed to R
+  expect_snapshot(
+    as_polars_series(c("2020-01-01", "2021-06-15"))$cast(pl$Date),
+    cnd_class = TRUE
+  )
```
etiennebacher, line 576: Does this mean that `suppressWarnings(classes = "PolarsDeprecationWarning")` would only hide this warning? I think this could be tested with:

```r
f <- function() {
  warning("hi")
  c("2020-01-01", "2021-06-15")
}

expect_snapshot(
    suppressWarnings(
       as_polars_series(f())$cast(pl$Date),
       classes = "PolarsDeprecationWarning"
    )
 )
```

which should only show the `"hi"` warning.

etiennebacher (commented):
Thanks! (for the other PRs too)

A few minor comments.

tests/testthat/test-expr-expr.R
eitsupi: I believe this is causing a warning on the main branch (Since #1804).

```
  > test-expr-expr.R: Deprecation: Casting from String to Date is deprecated and will be removed in Polars 2.0.
  > test-expr-expr.R: Use `str.to_date()` instead.
  > test-series-to_r_vector.R: Deprecation: Casting from String to DateTime is deprecated and will be removed in Polars 2.0.
  > test-series-to_r_vector.R: Use `str.to_datetime()` instead.
  > test-series-to_r_vector.R: Deprecation: Casting from String to DateTime is deprecated and will be removed in Polars 2.0.
  > test-series-to_r_vector.R: Use `str.to_datetime()` instead.
  > test-series-to_r_vector.R: Deprecation: Casting from String to DateTime is deprecated and will be removed in Polars 2.0.
  > test-series-to_r_vector.R: Use `str.to_datetime()` instead.
  [ FAIL 0 | WARN 0 | SKIP 2 | PASS 4075 ]
```

https://github.com/pola-rs/r-polars/actions/runs/28286301205/job/83810738792#step:11:210

tests/testthat/test-series-to_r_vector.R
eitsupi: Ditto.

tests/testthat/test-expr-expr.R
eitsupi: Definitely.

``` r
library(polars)

f <- function() {
  warning("hi")
  c("2020-01-01", "2021-06-15")
}

suppressWarnings(
  as_polars_series(f())$cast(pl$Date),
  classes = "PolarsDeprecationWarning"
)
#> Warning in f(): hi
#> shape: (2, 1)
#> ┌────────────┐
#> │            │
#> │ ---        │
#> │ date       │
#> ╞════════════╡
#> │ 2020-01-01 │
#> │ 2021-06-15 │
#> └────────────┘
```

<sup>Created on 2026-06-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher: I thought it would be good to add a test for this but it would become outdated every time a deprecation message is removed upstream, so maybe it's just not worth it.

src/rust/src/warn.rs
```diff
@@ -0,0 +1,61 @@
...
+
+static POLARS_WARNING_QUEUE: Mutex<Vec<QueuedWarning>> = Mutex::new(Vec::new());
+
+fn polars_warning_function(msg: &str, variant: PolarsWarning) {
+    let category = match variant {
+        PolarsWarning::Deprecation => "PolarsDeprecationWarning",
+        PolarsWarning::UserWarning => "PolarsUserWarning",
+        PolarsWarning::CategoricalRemappingWarning => "PolarsCategoricalRemappingWarning",
+        PolarsWarning::MapWithoutReturnDtypeWarning => "PolarsMapWithoutReturnDtypeWarning",
+    };
```
etiennebacher, line 19: Since those values end up as warning classes on the R side, shouldn't they be like `polars_deprecation_warning` for consistency with our other polars classes?

PR #1808, opened 2026-06-28
feat: add Ctrl+C cancellation for collect/profile/list_to_struct

### LINKED ISSUE #1453: Catch KeyboardInterrupts
Like this
https://github.com/pola-rs/polars/blob/03c772c68f5d907e7135aa0989a4a1d00e715d3e/crates/polars-python/src/utils.rs#L98-L114

eitsupi:
Close #1453

## Summary

Add Ctrl+C query cancellation for `collect()`, `profile()`, and `list_to_struct()`. Pressing Ctrl+C during a long-running Polars query now cancels the computation and returns control to R cleanly, raising an interrupt condition (catchable via `tryCatch(..., interrupt = function(e) ...)`) instead of hanging or crashing.

## Approach

polars-error already ships a panic-based SIGINT mechanism (`polars_error::signals`). We wire it up without adding any new Rust dependencies:

1. **At startup** (`on_startup.rs`): call `register_polars_keyboard_interrupt_hook()`, which registers a SIGINT handler via `signal-hook`. `signal-hook` *chains* the old handler, so R's own SIGINT handler also runs — `R_interrupts_pending` is set automatically.

2. **In the worker thread** (`wrap_with_cancel`): wrap the Polars computation with `catch_keyboard_interrupt()`. When Polars calls `try_raise_keyboard_interrupt()` at an internal computation boundary after a SIGINT, the resulting panic is caught. An `Arc<AtomicBool>` flag is set and a `PolarsError` is returned — no R APIs are called from the worker thread.

3. **Back on the main R thread**: after `concurrent_handler` returns, check the cancellation flag. If set, call `Rf_onintr()` inside `savvy::unwind_protect()`. This converts the R longjmp to a Rust `Err`, lets Rust run all destructors normally, and then savvy's entry point calls `R_ContinueUnwind` to re-raise the interrupt to R.

### Why `Rf_onintr()` instead of `R_CheckUserInterrupt()`

Investigation of [yutannihilation/savvy#314](https://github.com/yutannihilation/savvy/issues/314) showed that `R_CheckUserInterrupt()` is unreliable in Positron and some other IDEs (fails after the first interrupt). Since `catch_keyboard_interrupt()` already confirms the interrupt has occurred, calling `Rf_onintr()` directly is more appropriate. `Rf_onintr` is accessible via a raw `extern "C"` declaration — no savvy-ffi PR required, as it is already exported by `<Rinternals.h>`.

### Why `savvy::unwind_protect` is required

Calling `Rf_onintr()` directly from within a live Rust frame would cause `longjmp` to skip Rust destructors (undefined behavior per the Rust Reference). Wrapping the call in `savvy::unwind_protect` converts the longjmp to a Rust `Err`, ensuring proper stack unwinding before savvy re-raises the interrupt via `R_ContinueUnwind`.

### Cancellation granularity: streaming engine vs. in-memory engine

Cancellation is cooperative: it fires at Polars' internal `try_raise_keyboard_interrupt()` checkpoints. The **streaming engine** (`collect(engine = "streaming")`) places a checkpoint at every async task poll via `polars-async`, giving very fine-grained cancellation — heavy computations are reliably interrupted. The **in-memory engine** (default) only checks at executor node boundaries (start of join, sort, group_by, filter, etc.), so single large expressions such as `int_range(0, 2e10)` cannot be cancelled mid-computation. In practice, workloads heavy enough to need cancellation are typically run with the streaming engine.

### Panic hook interaction with savvy

savvy installs its own panic hook (which prints "panic occured!") at the start of every `#[savvy]` wrapper, pushing the polars suppression hook below it. This caused every `__POLARS_KEYBOARD_INTERRUPT` panic in the streaming executor's tasks to produce noisy "panic occured!" output. The fix is to install a scoped suppression hook manually inside `wrap_with_cancel` using `std::panic::take_hook()` + `std::panic::set_hook()`. This silences `__POLARS_KEYBOARD_INTERRUPT` panics without calling `register_polars_keyboard_interrupt_hook()`, which would also add another SIGINT handler on every call and cause unbounded handler accumulation. savvy restores its snapshot of `orig_hook` when the outer `#[savvy]` function returns, so no explicit cleanup is needed.

### `R_interrupts_pending` is not used

`R_interrupts_pending` lives in `<R_ext/GraphicsDevice.h>`, a header not intended as a package API. duckdb-r also avoids it. Since `signal-hook` chains R's SIGINT handler, `R_interrupts_pending` is set automatically without us touching it.

## Changes

| File | Change |
|------|--------|
| `src/rust/src/on_startup.rs` | New file: consolidates `#[savvy_init]` startup (polars-python pattern); registers SIGINT hook and warning function |
| `src/rust/src/warn.rs` | Remove `#[savvy_init]` (moved to `on_startup.rs`); expose `warning_function` as `pub(crate)` |
| `src/rust/src/r_threads.rs` | Add `Rf_onintr` extern declaration; add `wrap_with_cancel()` helper (re-registers polars suppression hook to sit above savvy's hook); remove stale extendr TODO comment |
| `src/rust/src/lazyframe/general.rs` | Rewrite `collect()` and `profile()` with cancellation support and proper `#[cfg]` WASM guards |
| `src/rust/src/series/general.rs` | Split `list_to_struct()` into nested / top-level-with-UDF (cancellable) / direct-call paths |
| `src/rust/api.h`, `src/init.c` | Regenerated to reference `savvy_on_startup__ffi` |

## Limitations

- **In-memory engine, single large expressions**: `try_raise_keyboard_interrupt()` checkpoints are at executor node boundaries. A single expression with no internal checkpoint (e.g. `int_range(0, 2e10)`) cannot be cancelled mid-computation; cancellation takes effect at the next node boundary after the SIGINT. The streaming engine does not have this limitation.
- **`list_to_struct()` without a name generator**: the direct (no-UDF) call path is not wrapped, as it runs synchronously on the R thread and is typically fast.
- **Nested execution** (calling `collect()` from within a Polars UDF): cancellation is not added to this path; it follows the original behavior.
- **DLL reload**: the SIGINT hook is process-global. Repeated `unloadNamespace()` / `loadNamespace()` cycles are not tested (typical for R packages; expected to be fine in practice).

## Testing

```r
library(polars)

# Streaming engine — fine-grained cancellation, reliably interrupted
n <- 3000L
lf <- pl$LazyFrame(a = 1:n, b = rnorm(n))
lf$join(lf, how = "cross")$collect(engine = "streaming")


# Catchable as an interrupt condition
tryCatch(
  lf$join(lf, how = "cross")$collect(engine = "streaming"),
  interrupt = function(e) message("Caught interrupt!")
)

# Verify normal queries still work after cancellation
pl$LazyFrame(a = 1:5)$collect()
```


src/rust/src/r_threads.rs
```diff
@@ -186,3 +182,49 @@ where
...
+    let prev_hook = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |info| {
+        let is_kbi = info
+            .payload()
+            .downcast_ref::<&str>()
+            .is_some_and(|s| s.contains("__POLARS_KEYBOARD_INTERRUPT"))
+            || info
+                .payload()
+                .downcast_ref::<String>()
+                .is_some_and(|s| s.contains("__POLARS_KEYBOARD_INTERRUPT"));
```
eitsupi, line 217: Ideally, this string should be exposed from the upstream to avoid hardcoding.
https://github.com/pola-rs/polars/blob/507e4fc79c0c78162ab34d17c0499ec03850b0cc/crates/polars-error/src/signals.rs#L10-L11

etiennebacher:
I don't know how to review this, I have no idea if it is correct or not. Should we wait for some feedback in the `savvy` issue?

eitsupi:
A simple functionality check as described would be sufficient.
If a review proves difficult, I'm considering simply merging and releasing it.
From what I've tested, there seem to be no problems.

etiennebacher:
I'll try to build this branch locally and do a few quick checks but I also don't have many usecases to test.

etiennebacher:
I can never trigger the `interrupt` part of `tryCatch()` and I also don't observe a cancellation behavior on code that takes a few seconds and a lot of memory to run (adapted from the example above). Did you manage to see this behavior with the example above?

eitsupi:
Thanks for checking.
Of course, I am testing this with my own examples.
Please note, as stated, that the in-memory engine has very limited cancellation capabilities.

You might also need to run an equivalent query in Python to isolate the issue. In that case, however, you will need to pay close attention to whether the cancellation occurred while the Rust process was running or after control had returned to the Python side.

etiennebacher:
Ok I managed to get the message once, but I don't have a lot of time or usecases to test this more extensively, at least it doesn't seem to have downsides.

Thank you!

etiennebacher (approved):

Issue #1024, opened 2024-04-11
Polars Expression plugins for R


eitsupi:
We needs:

1. Mechanism for registering subnamespaces from outside the package something like <https://docs.pola.rs/py-polars/html/reference/api.html>
2. Rust crate something like <https://github.com/pola-rs/pyo3-polars>

eitsupi:
Note: Serialization and deserialization of R objects that may be needed are already defined here (I don't know if this is sufficient)

https://github.com/pola-rs/r-polars/blob/1ea820b7791a1747fe477cb4d7d7172dcf619137/src/rust/src/rbackground.rs#L77-L131

eitsupi:
> 1. Mechanism for registering subnamespaces from outside the package something like [docs.pola.rs/py-polars/html/reference/api.html](https://docs.pola.rs/py-polars/html/reference/api.html)

I was able to make this work in an implementation that I am rewriting from scratch using py-polars as a reference.
https://github.com/eitsupi/neo-r-polars/blob/afac2ae8020e4dbe3d02f7515653a574283b577a/man/polars_api_register_series_namespace.Rd#L20-L44

The current concern is performance degradation due to frequent for loops (basically each call to a single method).
I believe the current implementation of r-polars registers all active bindings and methods when the package is installed, but it registers methods each time an R class instance is built, which would degrade performance (Of course, if it's acceptable, no problem)
https://github.com/eitsupi/neo-r-polars/blob/afac2ae8020e4dbe3d02f7515653a574283b577a/R/series-series.R#L7-L31

eitsupi:
I have looked into this and it appears that this is accomplished by connecting to a dynamic library via the libloading crate.
https://docs.rs/libloading/latest/libloading/
https://github.com/pola-rs/polars/blob/5cad69e5d4af47e75ae0abbf88dc2bafbc8f66d2/crates/polars-plan/src/dsl/function_expr/plugin.rs#L5

In the case of R packages, it is the static libraries, not the dynamic libraries, that are built by rustc.
Dynamic libraries are built by R.

We need to find a way to generate the proper expected C ABI on the plugin side, but this is obviously beyond my knowledge.

etiennebacher:
> In the case of R packages, it is the static libraries, not the dynamic libraries, that are built by rustc. Dynamic libraries are built by R.
> 
> We need to find a way to generate the proper expected C ABI on the plugin side, but this is obviously beyond my knowledge.

[...]

PR #1821, opened 2026-07-24
feat: add register_plugin_function

Shoeboxam:
Towards #1024. Allows for compiled plugin functions to be registered in the DSL. 

About pickles (gross): The implementation aims to be perfectly symmetric with Python to allow for plugin reuse. Unfortunately the core Polars library is opinionated about raw bytes being pickles, so the plugin ecosystem expects pickles. To maintain parity and allow plugin interop, the Rust side carries extra complexity to convert R objects into pickles. I've added a non-pickling kwargs_raw argument to avoid this code path for un-opinionated R-native plugins. Let me know if you'd rather the extra arg or the pickling be stripped out. On the other hand, there's no need for a dedicated r-polars plugin crate analogous to pyo3-polars... pyo3-polars is sufficiently general already.

Motivation: I would find this useful to extend OpenDP to support Polars in R.
Disclosure: I used AI to help me write the code, but I tested and checked it myself.


eitsupi (commented):
Thank you for working on this.
Are you trying to load a static library built for Python as an extension?

It seems there are compatibility issues with other codebases because it uses base R functions instead of rlang functions, but I can fix that later if necessary.

Before that, I'd like to know how you expect this to work and how we can test it locally.

Shoeboxam:
No, I'm trying to load a dynamic library built via pyo3-polars. The plugin ABI is language neutral, but I didn't realize pyo3-polars might leave behind a python link requirement that the compiler doesn't eliminate.

I'll handle the rlang swaps, that's minor, I just didn't anticipate it. 

Once I get the plugin build sorted, I'll add an example plugin that gets built and tested in CI.

Shoeboxam:
I think my angle of approach will be to open a PR that makes python a default feature of pyo3-polars. Sure, the crate has an unfortunate name, but it seems less painful than splitting and less maintenance than forking.

eitsupi:
So, in that case, are you saying that r-polars cannot be built without libpython?

If that's the case, I'm not sure if r-polars will still be useful.
It might be better to write a wrapper package that calls Python polars using reticulate.

Shoeboxam:
I agree, requiring python would defeat the purpose. 

Three approaches:
1. make pyo3 an optional feature in pyo3-python. Simplest, but the crate name is a bit unfortunate.
2. split the pyo3-python crate into two crates. Take a dep on the non-pyo3 crate. Re-export the same APIs through pyo3-python for backcompat, r-polars takes a dep on the split crate.
3. fork the crate and strip it down. needless maintenance work.

I was thinking I'd try approach 1, but after tinkering with it I'm switching to 2. I'm open to change, of course!

PR #1827, opened 2026-08-01
ci(release-lib): use ubuntu-24.04-arm runner for aarch64-unknown-linux-musl

eitsupi:
The 22.04 runner seems unable to complete the build, so I'll try the 24.04 runner. If that doesn't work, we'll have no choice but to exclude the arm64 musl build from the release.
https://github.com/pola-rs/r-polars/actions/runs/30681205454/job/91333266191


eitsupi:
Close in favor of #1829

PR #1828, opened 2026-08-01
chore: bump polars to `py-1.43.2`

etiennebacher:
This also removes a warning when compiling:

>    warning: patch `color-backtrace v0.7.2 (https://github.com/orlp/color-backtrace?rev=bb62ccf1e9eb1f6b7af5f16acff1fd7151a876dd#bb62ccf1)` was not used in the crate graph
>
>    help: Check that the patched package version and available features are compatible
         with the dependency requirements. If the patch has a different version from
         what is locked in the Cargo.lock file, run `cargo update` to use the new
         version. This may also occur with an optional dependency that is not enabled.


eitsupi (approved):
Thanks!

PR #1824, opened 2026-07-28
docs: fix some typos

jeis4wpi:
docs: fix some typos

Does this help?

PR #1837, opened 2026-08-15
build: use mini-roxygen for Rd file generator

eitsupi:
Changing the Rd generator from roxygen2 to mini-roxygen CLI (roxy).

mini-roxygen is very fast because it doesn't require compiling the Rust lib, and I think it's useful in cases like #1824.


etiennebacher:
Thanks, that looks very convenient.

I know this is a draft for now, but I just want to mention I'd like to actually review this before this is merged, but I'm not available until end of next week.

eitsupi:
Thank you. It's not urgent, so it's fine.
This PR was simply to see if the Rd files generated by mini-roxygen actually passes the R CMD check.

I'm planning to release mini-roxygen 0.1.0 as early as today and make it available for installation the CLI from GitHub.

etiennebacher:
I don't find `mini-roxygen`, was it already released?

eitsupi:
Sorry, I ended up fixing some bugs, so I haven't been able to release it yet (it's still only on my local machine).
I'm planning to release it sometime this week.

etiennebacher:
No problem :+1:  just ping me on this PR when it's released

eitsupi:
Released on GitHub and crates.io
https://github.com/eitsupi/mini-roxygen

Taskfile.yml
```diff
@@ -17,7 +17,7 @@ vars:
   MANIFEST: src/rust/Cargo.toml
   CARGO_LOCK: src/rust/Cargo.lock
   R_SOURCE: R/*
-  R_TESTS: tests/*
+  R_TESTS_SOURCE: tests/*
```
eitsupi, line 20: It seems `R_TESTS` is a reserved environment variable in R.
I changed the name because it caused problems due to go-task/task#2993.

man/dataframe__clone.Rd
```diff
@@ -31,21 +31,21 @@ rlang::env_label(df3)
 
 # Make a function to take a DataFrame, add an attribute, and return a
 # DataFrame:
-give_attr <- function(data) {
+give_attr <- function(data) \{
```
etiennebacher, line 34: Apparently this kind of changes doesn't affect the rendered docs, but I hope this also doesn't affect the `altdoc` process. Would it be hard to get rid of these changes in `mini-roxygen`?

mini-roxygen.toml
```diff
@@ -0,0 +1,4 @@
+# Ensure update this table if upstream polars add new compat levels
+[inline-r.substitutions]
+"(pl$CompatLevel$oldest:pl$CompatLevel$newest) |> format_code()" = '\code{0} and \code{1}'
```
etiennebacher, line 3:
```suggestion
"(pl$CompatLevel$oldest:pl$CompatLevel$newest) |> format_code()" = '\code{0} to \code{1}'
```
This is a range of versions, right?

etiennebacher (commented):
Looks very useful, thanks!

man/dataframe__clone.Rd
eitsupi: Here, the priority is on achieving equivalent values ​​(Rd) rather than producing exactly the same output.
I believe the actual processing takes place here.
https://github.com/eitsupi/r-documentation-rs/blob/2ea5cfaff05299ed41bf424febd0e55c4e678eca/crates/rd-writer/src/escape.rs#L79-L81

If you parse it, you'll see that the presence or absence of escapes has absolutely no effect on interpretation.
The conditions under which escapes can be removed must depend on the meaning within the R code, which would necessitate interpreting the R code, and it seems that the decision to simply escape is made because it is simpler and faster.

mini-roxygen.toml
eitsupi: This is simply using the original expression accurately.

```r
1:2 |> rlang:::format_code()
#> [1] "`1` and `2`"

1:3 |> rlang:::format_code()
#> [1] "`1`, `2`, and `3`"
```

In my opinion, since the values ​​need to be exact integers and there are currently only two, I think enumeration makes more sense than using "to" to represent a range.

etiennebacher: Right, I just misunderstood this part.

man/dataframe__clone.Rd
etiennebacher: I agree it shouldn't be a big deal here, but it would be even better if `mini-roxygen` could be a 100% accurate drop-in replacement (for the supported tags). 

> The conditions under which escapes can be removed must depend on the meaning within the R code, which would necessitate interpreting the R code

But here you don't remove escapes, you add some, which I think makes a difference in complexity. I suppose this escape addition could be skipped in `@examples` sections for instance.

eitsupi: I think we may be using "100% accurate drop-in replacement" in slightly different senses here.

If you parse the two Rd forms with `tools::parse_Rd()`, the resulting Rd values are the same.
In particular, `\{` and `{` in this context produce the same R code in the parsed document.
So from the point of view of R, rendering, `R CMD check`, and tools consuming parsed Rd, this is already semantically equivalent.
The difference is in the serialization of that Rd value, not in the value itself.

I also think simply skipping the additional escaping in `@examples` would be considerably harder than it might look.
This PR itself has good examples of that: an examples section can contain ordinary R braces while also containing Rd constructs such as `\dontshow{...}`, `\dontrun{...}`, `\donttest{...}`, etc.

This means that when the writer sees a `}`, it cannot simply assume that emitting a bare `}` is safe.
Depending on the surrounding parser state, that brace may be part of the R code, or it may terminate an Rd argument.
The same problem spans strings, raw strings, comments, nested braces, embedded Rd tags, and even boundaries between separate nodes in the Rd AST.

So my previous wording about "interpreting the R code" was a little too strong: this does not require evaluating R code.
But reproducing roxygen2's minimal escaping safely would require tracking enough of R's/Rd's lexical and parser state to prove which braces can be emitted unescaped while still guaranteeing that the output parses back into exactly the same Rd AST.

The current `rd-writer` deliberately takes the conservative approach: it emits an unambiguous representation and then verifies that parsing the generated Rd produces exactly the original AST.

For mini-roxygen, the compatibility goal is therefore semantic equivalence rather than byte-for-byte reproduction of roxygen2's serializer.
I agree that minimizing these differences would make migration diffs nicer, but I am not sure that reproducing all of roxygen2's escaping decisions is worth the additional parser complexity unless we find a case where the current representation actually changes the behavior of a downstream tool.

eitsupi:
@etiennebacher Is there anything you'd like to check?

etiennebacher (approved):
Sorry, I had forgotten to approve

eitsupi:
Thanks!

PR #1841, opened 2026-08-25
chore: Fixes for Jarl 0.6.0

etiennebacher:
Either skip more rules that have too many false positives, or apply fixes.

### DIFF
--- a/R/dataframe-s3-base.R
+++ b/R/dataframe-s3-base.R
@@ -283,7 +283,9 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     seq_n_rows <- seq_len(n_rows)
     idx <- if (is_logical(i)) {
       # If logical, `i` must be of length 1 or number of rows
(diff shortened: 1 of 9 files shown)


R/dataframe-s3-base.R
```diff
@@ -283,7 +283,9 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     seq_n_rows <- seq_len(n_rows)
     idx <- if (is_logical(i)) {
       # If logical, `i` must be of length 1 or number of rows
-      if (!length_i %in% c(1L, n_rows)) {
+      if (length_i %in% c(1L, n_rows)) {
+        seq_n_rows[i]
+      } else {
```
etiennebacher, line 288: All changes on if conditions are to move from `if (!A) foo else bar` to `if (A) bar else foo`.

This is a readability rule so is necessarily opinionated, I don't mind reverting if you have a strong opinion on this.

eitsupi (approved):
Thanks!

PR #1847, opened 2026-08-30
chore: pre built binaries only support R 4.5.0 or later

eitsupi:
See yutannihilation/savvy#461

Binaries built with the latest version of R (currently 4.6) are no longer compatible with version 4.3.0 to 4.4.x.

### DIFF
--- a/.github/workflows/release-lib.yml
+++ b/.github/workflows/release-lib.yml
@@ -161,14 +161,14 @@ jobs:
           - windows-latest
           - ubuntu-latest
         r:
-          - "4.3"
+          - "4.5"
           - release
           - devel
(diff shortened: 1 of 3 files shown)


etiennebacher:
I don't understand, the `savvy` PR adds support back for R >= 4.3, so why do pre-built binaries now only work with R >= 4.5?

eitsupi:
Because the build process branches during construction, binaries that use `R_getVarEx` and those that don't are incompatible with each other.

https://github.com/pola-rs/r-polars/actions/runs/33301548779/job/99238135876#step:9:6252

```
undefined symbol: R_getVarEx
```

PR #1851, opened 2026-09-01
ci(release-lib): fix arm64 musl linux binary builds

### LINKED ISSUE #1760: bug(release-lib): __fprintf_chk (glibc fortification symbol) present in `aarch64-unknown-linux-musl` binary
The static library provided for the current 1.7.0 release candidate does not work correctly and as one would expect on alpine and also other `musl` environments.

It looks like this is a misconfiguration and consequence of an inconsistent setup on `r-polars` build side.

https://github.com/pola-rs/r-polars/blob/ef3c6d752f4a80e59f1fd2a841c67a4a853e3681/.github/workflows/release-lib.yml#L80

Since `GCC` is set as C compiler and `_FORTIFY_SOURCE=2`, the `jemalloc` memory allocator's C source is also compiled with GNU `GCC` (via cargo and `jemalloc-sys` -> `build.rs`) which is by default only intended for `glibc` targets. Here it is not only for the linker. 

 `__fprintf_chk` symbol emitted is not available for `musl` environments. 

I made a workaround with `LD_PRELOAD`, but this is very unclean and better fixed upstream here :)

```yaml
steps:
  'Check package':
    image: <myorgs-forgejo-container-registry-derived-from-alpine:3.22-image>
    pull: true
    environment:
      GITHUB_PAT:
        from_secret: GITHUB_PAT
      RCMDCHECK_NUM_COLORS: 300
      PKG_SYSREQS_PLATFORM: alpine
      PKG_SYSREQS: TRUE
      PKG_SYSREQS_VERBOSE: TRUE
      RENV_CONFIG_REPOS_OVERRIDE: https://<myorgs-cran-mirror>/arm64/alpine322/latest
      NOT_CRAN: 'true'
      LIBR_POLARS_BUILD: 'false'

    commands:
      - apk add --no-cache -q gcompat rust cargo
      - |
        mkdir -p ~/.R
        LD_PRELOAD=/lib/libgcompat.so.0 R -q -e '
          options(Ncpus = 4)
          options(repos = c(
            RMultiverse = "https://community.r-multiverse.org",
            CRAN = Sys.getenv("RENV_CONFIG_REPOS_OVERRIDE")
          ))
          renv::restore()
          rcmdcheck::rcmdcheck(
            args = c("--no-manual"),
            error_on = "warning"
          )
```

```sh
Error: Error installing package 'polars':
==================================

* installing *source* package ‘polars’ ...
** this is package ‘polars’ version ‘1.7.0’
- Installing polars ...                         FAILED
** using staged installation
--------------------- [ENVIRONMENT VARIABLES] ---------------------
TARGET = ''
PROFILE = 'release'
FEATURES = ''
NOT_CRAN = 'true'
LIBR_POLARS_BUILD = 'false'
-------------------------------------------------------------------
---------------------- [MOLD LINKER DISABLED] ----------------------
'mold' not found in PATH. Continuing without mold.
--------------------------------------------------------------------
---------------- [TRY TO DOWNLOAD PRE-BUILT BINARY] ----------------
The library was not found at </tmp/RtmpIiMgfm/renv-git-e6d71cf8e/tools/libr_polars.a>.
Trying to download pre-built binary from the Internet...
--------------------------------------------------------------------
Found pre-built binary at <[https://github.com/pola-rs/r-polars/releases/download/lib-v1.7.0-rc.3/libr_polars-1.7.0-rc.3-aarch64-unknown-linux-musl.tar.gz>.](https://github.com/pola-rs/r-polars/releases/download/lib-v1.7.0-rc.3/libr_polars-1.7.0-rc.3-aarch64-unknown-linux-musl.tar.gz%3E.)
Downloading...
Checking SHA256 for </tmp/RtmpBfcldA/fileb453799e84e.tar.gz>...
SHA256 matches for </tmp/RtmpBfcldA/fileb453799e84e.tar.gz>.
Extracted pre-built binary to </tmp/RtmpIiMgfm/renv-git-e6d71cf8e/tools> directory.
Done!
---------------------- [LIBRARY BINARY FOUND] ----------------------
The library was found at </tmp/RtmpIiMgfm/renv-git-e6d71cf8e/tools/libr_polars.a>. No need to build it.

Note: rustc version: rustc 1.87.0 (17067e9ac 2025-05-09) (Alpine Linux 1.87.0-r0)
--------------------------------------------------------------------
** libs
using C compiler: ‘gcc (Alpine 14.2.0) 14.2.0’
gcc -std=gnu23 -I"/opt/R/4.5.2/lib/R/include" -DNDEBUG   -I/usr/local/include -D__MUSL__    -fPIC  -g -O2  -D__MUSL__ -flax-vector-conversions -c init.c -o init.o
rm -f "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target//release/libr_polars.a"
if [ -f "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/../tools/libr_polars.a" ]; then \
	mkdir -p "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target//release" ; \
	mv "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/../tools/libr_polars.a" "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target//release/libr_polars.a" ; \
	exit 0; \
fi && \
export PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/root/.cargo/bin" && \
  if [ "" != "wasm32-unknown-emscripten" ]; then \
    cargo build --lib --profile="release" --manifest-path="/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/Cargo.toml" --target-dir="/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target" --target="" --features=""; \
  else \
    export CC="gcc -std=gnu23" && \
    export CFLAGS="-g -O2  -D__MUSL__ -flax-vector-conversions" && \
    export CARGO_PROFILE_DEV_PANIC="abort" && \
    export CARGO_PROFILE_RELEASE_PANIC="abort" && \
    export RUSTFLAGS=" -Zdefault-visibility=hidden" && \
    cargo build --lib --profile="release" --manifest-path="/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/Cargo.toml" --target-dir="/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target" --target="" --features="" -Zbuild-std=panic_abort,std; \
  fi
gcc -std=gnu23 -shared -L/opt/R/4.5.2/lib/R/lib -L/usr/local/lib -fPIC -o polars.so init.o -L/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target//release -lr_polars -L/opt/R/4.5.2/lib/R/lib -lR
if [ "true" != "true" ]; then \
	rm -Rf "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target//release/libr_polars.a" "" "" "/tmp/RtmpIiMgfm/renv-git-e6d71cf8e/src/rust/target//release/build"; \
fi
installing to /crow/src/<mydomain>/<myorg>/<mypackage>/renv/staging/1/00LOCK-renv-git-e6d71cf8e/00new/polars/libs
** R
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
*** copying figures
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
Error: package or namespace load failed for ‘polars’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/crow/src/<mydomain>/<myorg>/<mypackage>/renv/staging/1/00LOCK-renv-git-e6d71cf8e/00new/polars/libs/polars.so':
  Error relocating /crow/src/<mydomain>/<myorg>/<mypackage>/renv/staging/1/00LOCK-renv-git-e6d71cf8e/00new/polars/libs/polars.so: __fprintf_chk: symbol not found
Error: loading failed
Execution halted
ERROR: loading failed
* removing ‘/crow/src/<mydomain>/<myorg>/<mypackage>/renv/staging/1/polars’
install of package 'polars' failed [error code 1]
Traceback (most recent calls last):
13: renv::restore()
12: renv_restore_run_actions(project, diff, current, lockfile, rebuild)
11: renv_install_impl(records)
10: renv_install_staged(records)
 9: renv_install_default(records)
 8: handler(package, renv_install_package(record))
 7: renv_install_package(record)
 6: withCallingHandlers(renv_install_package_impl(record), error = function(e) writef("FAILED"))
 5: renv_install_package_impl(record)
 4: r_cmd_install(package, path)
 3: r_exec_error(package, output, "install", status)
 2: abort(all)
 1: stop(fallback)
```

philipp-baumann:
## Summary

Fixes https://github.com/pola-rs/r-polars/issues/1760

Builds the `aarch64-unknown-linux-musl` target's Rust code with `rustc` targeting musl, and cross-compiles its C components (jemalloc, ring, ...) with Zig (`zig cc`), which compiles C against musl with a working atomics implementation, so jemalloc builds without leaking glibc fortification symbols such as `__fprintf_chk`.

## What changed

- `.github/workflows/release-lib.yml`: Replaced the `CC=aarch64-linux-gnu-gcc` workaround with a `setup Zig` step that downloads the official Zig 0.16.0 `aarch64-linux` build with sha256 verification, and sets `CC_aarch64_unknown_linux_musl` / `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER` to the Zig wrapper for the `aarch64-unknown-linux-musl` target.
- `.github/scripts/zig-aarch64-musl-cc.sh`: Added a wrapper that rewrites the cc crate's Rust-style target triple (`--target=aarch64-unknown-linux-musl`) to Zig's own format (`--target=aarch64-linux-musl`) before executing `zig cc`, because Zig rejects the Rust triple with `UnknownOperatingSystem`.

## Behaviour change

The `aarch64-unknown-linux-musl` static library is now built with rustc targeting musl, and its C components (jemalloc, ring, ...) are compiled by Zig (`zig cc`) against musl libc instead of a glibc C compiler. As a result, glibc fortification symbols such as `__fprintf_chk` are no longer embedded, and the package should load correctly on Alpine/musl linux machines. `jemalloc` is retained as the prefereed memory allocator in polars.

```sh
build lib step
  └─ cargo build --target=aarch64-unknown-linux-musl
       └─ cc crate compiles ring/jemalloc/zstd C
            └─ invokes CC_aarch64_unknown_linux_musl  (= wrapper script)
                 └─ wrapper rewrites --target=aarch64-unknown-linux-musl → aarch64-linux-musl
                      └─ exec zig cc ...  (compiles C against musl)
```

- Before: C sources were compiled by glibc `aarch64-linux-gnu-gcc` → glibc symbols leaked in.
- Now: C sources are compiled by Zig (`zig cc`) against musl → no glibc symbols.

The added wrapper is a C compiler for the C-only crates; it injects nothing into the Rust build or the Rust-produced machine code.

## Notes

- The `aarch64-unknown-linux-gnu` matrix build passes; the `aarch64-unknown-linux-musl` target still needs a green run on the native ARM64 runner to confirm Zig cross-compiled the C components (jemalloc, ring, ...) correctly.
- GitHub Actions runs `uses:`-style actions as Node.js JavaScript actions, and its runtime refuses to execute them inside an Alpine container on an ARM64 runner:

```sh
Error: JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected Linux Arm64
```

That is a documented limitation of the Actions runtime, not something a config tweak fixes.

---
AI assistance notice: assisted with deepseek-v4-flash-0731 in pi agent setup ( https://codefloe.com/philipp-baumann/agentic-tooling-and-configs )

### DIFF
--- a/.github/scripts/zig-aarch64-musl-cc.sh
+++ b/.github/scripts/zig-aarch64-musl-cc.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+# zig-cc wrapper that bridges the cc crate's Rust-style target triple to Zig.
+#
+# The cc crate (used by ring, jemalloc-sys, etc.) passes the target triple as
+# `--target=aarch64-unknown-linux-musl`. Zig rejects the `-unknown-` vendor
+# component and requires its own format `aarch64-linux-musl`. This wrapper
+# rewrites the triple before exec'ing `zig cc`.
+set -eu
+
+newargs=""
+for arg in "$@"; do
+  # Skip a stray leading `cc` subcommand; cc-rs invokes the wrapper with the
+  # flags only, but some callers pass `cc` explicitly.
+  case "$arg" in
+    cc)
+      continue
+      ;;
+  esac
+  case "$arg" in
+    --target=aarch64-unknown-linux-musl)
+      arg="--target=aarch64-linux-musl"
+      ;;
+  esac
+  # Quote each arg so paths with spaces survive the final eval.
+  newargs="$newargs '$(printf '%s' "$arg" | sed "s/'/'\\\\''/g")'"
+done
+
+eval "exec zig cc $newargs"
--- a/.github/scripts/zig-aarch64-musl-cc.sh
+++ b/.github/scripts/zig-aarch64-musl-cc.sh
@@ -9,20 +9,20 @@ set -eu
 
 newargs=""
 for arg in "$@"; do
-  # Skip a stray leading `cc` subcommand; cc-rs invokes the wrapper with the
-  # flags only, but some callers pass `cc` explicitly.
-  case "$arg" in
-    cc)
-      continue
-      ;;
-  esac
-  case "$arg" in
-    --target=aarch64-unknown-linux-musl)
-      arg="--target=aarch64-linux-musl"
-      ;;
-  esac
-  # Quote each arg so paths with spaces survive the final eval.
-  newargs="$newargs '$(printf '%s' "$arg" | sed "s/'/'\\\\''/g")'"
+	# Skip a stray leading `cc` subcommand; cc-rs invokes the wrapper with the
+	# flags only, but some callers pass `cc` explicitly.
+	case "$arg" in
+	cc)
+		continue
+		;;
+	esac
+	case "$arg" in
+	--target=aarch64-unknown-linux-musl)
+		arg="--target=aarch64-linux-musl"
+		;;
+	esac
+	# Quote each arg so paths with spaces survive the final eval.
+	newargs="$newargs '$(printf '%s' "$arg" | sed "s/'/'\\\\''/g")'"
 done
 
 eval "exec zig cc $newargs"
--- a/.github/workflows/release-lib.yml
+++ b/.github/workflows/release-lib.yml
@@ -32,8 +32,6 @@ jobs:
           - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
           - os: ubuntu-22.04-arm
-            target: aarch64-unknown-linux-musl
-          - os: ubuntu-22.04-arm
             target: aarch64-unknown-linux-gnu
           - os: macos-latest
             target: x86_64-apple-darwin
@@ -71,13 +69,6 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y musl-tools
-
-      - name: pref for aarch64-unknown-linux-musl
-        if: matrix.target == 'aarch64-unknown-linux-musl'
-        # Ref <https://github.com/rust-lang/stacker/issues/80>
-        run: |
-          echo 'CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-gnu-gcc' >>"$GITHUB_ENV"
-          echo 'CC=aarch64-linux-gnu-gcc' >>"$GITHUB_ENV"
 
       - name: limit rust build jobs on aarch64 linux
         if: runner.os == 'Linux' && runner.arch == 'ARM64'
@@ -148,6 +139,75 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           name: libs-${{ matrix.target }}
+          path: ${{ env.ARTIFACT_NAME }}
+
+  # Build the aarch64 musl static library natively inside an Alpine (musl)
+  # ARM64 container on GitHub's native ARM64 runner. This is required because
+  # cross-compiling the musl target with a glibc C toolchain (the previous
+  # `CC=aarch64-linux-gnu-gcc` workaround) leaked glibc fortification symbols
+  # like `__fprintf_chk` into the library, which are missing on musl and broke
+  # loading on Alpine (#1760). A native musl toolchain cannot leak glibc
+  # symbols and avoids the jemalloc C-atomics detection failure that `musl-gcc`
+  # on the Ubuntu host runner triggers.
+  build-aarch64-musl:
+    runs-on: ubuntu-24.04-arm
+    container:
+      image: alpine:3.22
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: install system deps
+        run: |
+          # R and R-dev live in Alpine's community repository
+          apk add --no-cache --repository "https://dl-cdn.alpinelinux.org/alpine/v3.22/community" \
+            bash curl gcc jq make musl-dev pkg-config R R-dev tar
+
+      - name: install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
+          echo "$HOME/.cargo/bin" >>"$GITHUB_PATH"
+          TOOLCHAIN="$(sed -n 's/^channel = "\(.*\)"/\1/p' src/rust/rust-toolchain.toml)"
+          rustup toolchain install "${TOOLCHAIN}" --profile minimal
+          rustup target add aarch64-unknown-linux-musl --toolchain "${TOOLCHAIN}"
+          rustup default "${TOOLCHAIN}"
+
+      - name: install R pkgbuild
+        run: |
+          Rscript -e 'options(repos = c(CRAN = "https://cloud.r-project.org")); install.packages("pkgbuild")'
+
+      # Serialize rustc: the ARM runner runs out of memory while polars-plan and
+      # polars-expr compile at the same time.
+      - name: limit rust build jobs on aarch64 linux
+        run: |
+          echo 'CARGO_BUILD_JOBS=1' >>"$GITHUB_ENV"
+          echo 'CARGO_PROFILE_DIST_RELEASE_LTO=thin' >>"$GITHUB_ENV"
+
+      - name: build lib
+        env:
+          NOT_CRAN: "true"
+          TARGET: aarch64-unknown-linux-musl
+          LIBR_POLARS_BUILD: "true"
+          LIBR_POLARS_FEATURES: nightly
+          LIBR_POLARS_PROFILE: dist-release
+        run: |
+          # savvy needs R_INCLUDE_DIR at build time
+          export R_INCLUDE_DIR="$(Rscript -e 'cat(normalizePath(R.home("include")))')"
+          LIB_VERSION="$(cargo metadata --format-version=1 --no-deps --manifest-path src/rust/Cargo.toml | jq --raw-output '.packages[0].version')"
+          echo "LIB_VERSION=${LIB_VERSION}" >>"$GITHUB_ENV"
+          Rscript -e 'pkgbuild::compile_dll(debug = FALSE)'
+
+          ARTIFACT_NAME="${LIB_NAME}-${LIB_VERSION}-${TARGET}.tar.gz"
+          tar -czf "${ARTIFACT_NAME}" -C "src/rust/target/${TARGET}/${LIBR_POLARS_PROFILE}" "${LIB_NAME}.a"
+          echo "ARTIFACT_NAME=${ARTIFACT_NAME}" >>"$GITHUB_ENV"
+
+      - name: upload artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: libs-aarch64-unknown-linux-musl
           path: ${{ env.ARTIFACT_NAME }}
 
   test:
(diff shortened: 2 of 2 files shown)


philipp-baumann:
As the builds pass, this PR is ready for review from my side. If you agree with the approach, I think the fix could be part of the next release candidate -- as a patch release (e.g. `1.15.1-rc.1`) or the next minor (`1.16.0-rc.1`)?

At work we have an R package that relies on the `py-polars` runtime, but I would like to add `r-polars` (i.e., this release candidate with fixed musl buildchain) as a dependency, and run an extra check pipeline (crow CI) using `platform: linux/arm64` as `labels`. https://codefloe.com/Alliance-SwissPass/r-containerio/src/branch/main/.crow/rcmdcheck.yaml if the rc is successful without https://git.adelielinux.org/adelie/gcompat to make binary work. So this would be proof for working cross-compilation features on libc part of the triple. Because the glued interface of GitHub and their incompletely compiled node runtimes failing to be injected into actions within containers (i.e., arm64 musl glibc based Alpine containers) would otherwise require quite hacky ways.

So the rc test plan would be:
1. add the fixed r-polars RC as a dependency of R package
2. run the Crow rcmdcheck pipeline on r-alpine:4.5-3.24 container image on `linux/arm64`.
3. verify it works without the gcompat `LD_PRELOAD` workaround — proving the musl buildchain is now clean.

Below some annotations to illustrate where and how `zig cc` acts in the build chain.

```yaml
echo "CC_aarch64_unknown_linux_musl=$GITHUB_WORKSPACE/.github/scripts/zig-aarch64-musl-cc.sh" >>"$GITHUB_ENV"
echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=$GITHUB_WORKSPACE/.github/scripts/zig-aarch64-musl-cc.sh" >>"$GITHUB_ENV"
```

Both lines set environment variables that tell the Rust build which C compiler and linker to use for the `aarch64-unknown-linux-musl` target. Each points at the wrapper script `.github/scripts/zig-aarch64-musl-cc.sh`:

- `CC_aarch64_unknown_linux_musl` is the C compiler for the musl target. The `cc` crate (used by jemalloc, ring, zstd, lz4, brotli) invokes it to compile their C sources to `.o`. This is the var that actually matters for the fix.
- `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER` is the linker for the musl target that cargo/rustc would use for final linking. It is set defensively (the same wrapper), but for a `staticlib` (this crate) the `.a` is assembled by `ar`, so this linker var is not exercised in this build.

Both go through the same `zig-aarch64-musl-cc.sh` wrapper, which rewrites Zig's target triple to musl before running `zig cc`.

- Host (the runner, ubuntu-22.04-arm): has glibc, but `zig cc` which is added by the install step ships a complete musl libc bundled.
- Target: aarch64-unknown-linux-musl (.a)

In a nutshell, Zig produces C objects for the musl target, using musl's headers/libc, and not the host glibc. So even though the runner machine itself has glibc, none of glibc's fortification symbols (__fprintf_chk) end up in the built library.

```mermaid
flowchart TB
    RUST["Rust sources"] --> RUSTC["rustc"]
    RUSTC --> RO["Rust objects (.o)"]
    C["C deps (jemalloc, ring, zstd, lz4, brotli)"] -->|CC_aarch64_unknown_linux_musl| ZIG["zig cc (compiler + linker)"]
    ZIG --> OBJ["C objects (.o)"]

    RO --> AR["ar -- ar rcs libr_polars.a (archiver, combines .o)"]
    OBJ --> AR
    AR --> A["libr_polars.a (archive) [CARGO_TARGET_..._LINKER -> same wrapper]"]

    A -->|prep-lib.R download| C2["R CMD SHLIB (init.c + libr_polars.a)"]
    C2 --> SO["polars.so"]
    SO --> LD["R dyn.load"]

    classDef build fill:#dbeafe,stroke:#1d4ed8,stroke-width:2px,color:#1e3a8a;
    classDef tool fill:#ede9fe,stroke:#6d28d9,stroke-width:2px,color:#4c1d95;
    classDef install fill:#fef3c7,stroke:#b45309,stroke-width:2px,color:#78350f;
    classDef runtime fill:#dcfce7,stroke:#166534,stroke-width:2px,color:#14532d;
    class RUST,RO,OBJ,C,A build;
    class RUSTC,ZIG,AR tool;
    class C2 install;
    class SO,LD runtime;
```

eitsupi:
Thanks for this, looks great!

We should be able to download the artifacts here and try them.
I'll take a look later.
https://github.com/pola-rs/r-polars/actions/runs/33539566237

philipp-baumann:
> Thanks for this, looks great!
> 
> We should be able to download the artifacts here and try them. I'll take a look later. https://github.com/pola-rs/r-polars/actions/runs/33539566237

thanks for the artefifact download hint :) I will test it later today; looking forward.

philipp-baumann:
> > Thanks for this, looks great!
> > We should be able to download the artifacts here and try them. I'll take a look later. https://github.com/pola-rs/r-polars/actions/runs/33539566237
> 
> thanks for the artefifact download hint :) I will test it later today; looking forward.

I finally got the time to verify it works end-to-end on Alpine 3.24 :)

https://codefloe.com/philipp-baumann/polars-cicd-bench/src/branch/main/.crow/r-polars-prerelease.yaml
https://ci.codefloe.com/repos/1777/pipeline/12/3

What the pipeline does (r-polars-prerelease.yaml, linux/arm64 Alpine):
- downloads the prebuilt artifact: `libr_polars.a` (fixed artifact 9814836705 build from commit 9cb2c80)
- Installs R polars from the fork branch pinned to commit 9cb2c80 via `remotes::install_github()`
- Setting `LIBR_POLARS_BUILD=false` and `LIBR_POLARS_PATH` makes configure copy the downloaded `.a` into the package instead of building from source.
- Links the `.a` artefact into `polars.so` shared object, then loads polars and builds a simple DataFrame. The gcompat patching is not needed anymore and everything loads cleanly on musl libc, here Alpine 3.24 based container image.
- The install-side gcc on Alpine is musl-targeting gcc (gcc (Alpine 15.2.0) + -D__MUSL__). It only compiles init.c and links `polars.so`; the heavy C/Rust library comes from the prebuilt `.a` pinned to the last working commit and resulting actions artefact.

eitsupi (approved):
Thanks for checking! (Sorry, I wanted to pull out my Raspberry Pi and test it, but I couldn't find the time.)

Let's try merging it.

eitsupi:
I will aim to release version 1.16, including this, as soon as possible.

eitsupi:
#1875 regenerated the lib sums file including this.

PR #1874, opened 2026-09-08
docs(news): polish NEWS for 1.16.0

eitsupi:
Also execute auto formatting

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 This release is the migration bridge from R Polars 1.x to R Polars 2.0.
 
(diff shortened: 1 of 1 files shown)


etiennebacher:
Thanks for handling that @eitsupi. I pushed directly to this branch to make a few changes. Most are just formatting changes to wrap long lines, not much has changed in terms of content, except that I 1) made some paragraphs sound less like LLM and 2) added some sentences to say that `cat$get_categories()` and `$profile()` would be removed, and a few other tweaks.

etiennebacher (approved):

PR #1877, opened 2026-09-12
chore: prepare for 1.16.0

eitsupi:
Based on #1878, we will make final adjustments to the deprecation warnings and release 1.16.0.


eitsupi:
It was more difficult than I expected, so I've given up on releasing 1.16.0 from here. I'll merge it for now.

PR #1884, opened 2026-09-20
chore: bump polars to py-2.0.0-rc.2

eitsupi:


eitsupi:
The macOS CI failure seems unrelated.
A bug in the pak? Created r-lib/pak#915

R/expr-expr.R
```diff
@@ -4338,6 +4377,143 @@ expr__qcut <- function(
...
+#' df$select(bin = pl$col("foo")$bin_intervals(
+#'   c(-1, 1), labels = c("low", "mid", "high"), right_closed = TRUE
+#' ))
+expr__bin_intervals <- function(
+  intervals,
+  ...,
+  labels = NULL,
+  include_intervals = FALSE,
+  right_closed = FALSE
+) {
```
eitsupi, line 4412: Unlike Python, we can't distinguish between `1` and `c(1)` here, so I'm not sure if this API is ideal.

APIs like sample or hist are more likely to be appropriate.

```r
function(
  intervals = NULL,
  ...,
  n_bins = NULL,
  labels = NULL,
  include_intervals = FALSE,
  right_closed = FALSE
)
```

man/expr__bin_intervals.Rd
```diff
@@ -0,0 +1,48 @@
+% Generated by mini-roxygen (roxygen2 compatible): do not edit by hand
```
etiennebacher, line 1: Unrelated to this PR but one annoying thing with the switch to `mini-roxygen` is that Github no longer automatically collapses the Rd files when reviewing a PR. Do you know a way to automatically collapse them?

R/datatypes-classes.R
```diff
@@ -38,11 +38,18 @@
 #' | `Duration`                                     | Represents a time duration.                                                     |
 #' | `Array`                                        | Arrays with a known, fixed shape per series; akin to numpy arrays.              |
 #' | `List`                                         | Homogeneous 1D container with variable length.                                  |
+#' | `Map`                                          | Key/value entries; their types are exposed through `$key` and `$value`.        |
```
etiennebacher, line 41: I must say I'm confused about the use cases for this datatype but nice to have it here too.

R/expr-expr.R
```diff
@@ -4265,6 +4270,19 @@ expr__cut <- function(
...
     check_dots_empty0(...)
+    deprecate_warn(
+      c(
+        `!` = sprintf(
+          "The method %s is deprecated as of %s 2.0.0.",
+          format_fn("<expr>$cut"),
+          format_pkg("polars")
+        ),
+        i = sprintf(
+          "Use the experimental method %s instead.",
```
etiennebacher: Sounds a bit weird to deprecate a function and recommend something experimental instead. Should we just drop the "experimental" from this message?

R/expr-expr.R
```diff
@@ -4338,6 +4377,164 @@ expr__qcut <- function(
...
+  n_bins = NULL,
+  labels = NULL,
+  include_intervals = FALSE,
+  right_closed = FALSE
+) {
+  wrap({
+    check_dots_empty0(...)
+    if (is.null(intervals) == is.null(n_bins)) {
+      abort("Supply exactly one of `intervals` or `n_bins`.")
+    }
```
etiennebacher: `rlang::check_exclusive()`? (and same in the other bin functions)

R/expr-expr.R
etiennebacher: Not sure what you meant it

tests/testthat/test-datatypes-classes.R
```diff
@@ -47,6 +48,17 @@ patrick::with_parameters_test_that(
...
 )
 
+test_that("Map dtype constructor validates inputs", {
+  dtype <- pl$Map(pl$String, pl$Int32)
+  expect_s3_class(dtype, "polars_dtype_map")
+  expect_true(dtype$is_nested())
+  expect_equal(format(dtype$key), "String")
+  expect_equal(format(dtype$value), "Int32")
+
+  expect_snapshot(pl$Map(1, pl$Int64), error = TRUE)
```
etiennebacher, line 58: Can you add `expect_snapshot(pl$Map(pl$String, 1), error = TRUE)` as well?

tests/testthat/test-expr-expr.R
```diff
@@ -2978,7 +2978,119 @@ test_that("rle_id works", {
...
+  )
+
+  interval_bins <- df$select(
+    bin = pl$col("foo")$bin_intervals(n_bins = 2L, labels = NULL)
+  )
+  expect_identical(interval_bins$schema, list(bin = pl$UInt32))
+  expect_identical(
+    interval_bins$get_column("bin")$to_r_vector(int64 = "character"),
+    c("0", "0", "1", "1", "1")
+  )
```
etiennebacher: Why isn't this (and tests below) a single `expect_equal()` like above?

NEWS.md
etiennebacher, line 1: Just a note for later: I'd like to review the news file before 2.0 is released (same as for 1.16)

etiennebacher (commented):
Thanks, LGTM overall with some comments below

man/expr__bin_intervals.Rd
eitsupi: Defined here.
https://github.com/github-linguist/linguist/blob/538da05f034fa5c83bd81df128d144e208898da3/lib/linguist/generated.rb#L749-L754

I considered updating it, but I've held off for now. I think it's sufficient to just use `% Generated by` for the determination.

R/datatypes-classes.R
eitsupi: To be honest, I don't think there are any particular use cases for this.
The reason it was added is probably for compatibility.

For example, the ADBC ​​spec returns some metadata as Map type, so trying to use ADBC ​​drivers with Polars sometimes caused a panic.

R/expr-expr.R
eitsupi: Unfortunately, I believe it's the upstream polars that are being declared experimental...

etiennebacher: The docs contain "experimental" but the message doesn't: https://github.com/pola-rs/polars/blob/950d567d0220b934c7683e5e85c0484a5d0c8f25/py-polars/src/polars/expr/expr.py#L4785-L4869

Probably not that important though

etiennebacher (approved):
Thanks!

